Issue #11, opened 2018-12-21
Define behaviour of retrieve_job() from another backend


diego-plan9:
### What is the expected enhancement?

As noticed by @jaygambetta , currently `IBMQBackend.retrieve(job_id)` does _not_ enforce that the `job_id` belongs to a job that has _really_ been executed in that backend - it searches for that id among all the jobs that exists in the user account. This is prone to causing some confusion, and might make the user fetch jobs from other backends inadvertently.

### Proposed solution

We should detect that situation, and the function should act accordingly, with the ultimate goal of making it easy for the users - I think the decision is mostly about how strict we want to be. It might be worth noticing that `backend.jobs()` [does enforce a hard filter by backend](https://github.com/Qiskit/qiskit-terra/blob/master/qiskit/providers/ibmq/ibmqbackend.py#L137), so I'd lean towards keeping raising a `IBMQBackendError` if the job does not belong to a backend, along with a warning so users can correct the situation (ie. by using the right backend). Any suggestions?

delapuente:
Are you planning to warn with something like `The job XXXX does not belong to this backends but it belongs to "this other backend"`?

diego-plan9:
That would be a nice touch, yes :+1:

PR #23, opened 2019-01-21
Test that retrieved jobs come from their appropriate backend

Max-Reuter:
### Summary
Test that retrieved jobs come from their appropriate backend

### Details and comments
Filtered backends on the Python side instead of via query to the API. See issue #11.

### DIFF
--- a/qiskit/providers/ibmq/ibmqbackend.py
+++ b/qiskit/providers/ibmq/ibmqbackend.py
@@ -194,9 +194,16 @@ class IBMQBackend(BaseBackend):
         """
         try:
             job_info = self._api.get_job(job_id)
+            print(job_info['backend']['name'])
+            print(self.name())
             if job_info['backend']['name'] != self.name():
-                raise IBMQBackendError('Job "{}" belongs to backend "{}", not "{}".'
-                                       .format(job_id, job_info['backend']['name'], self.name()))
+                warnings.warn('Job "{}" belongs to another backend than the one queried. '
+                              'The query was made on backend "{}", '
+                              'but the job actually belongs to backend "{}".'
+                              .format(job_id, job_info['backend']['name'], self.name()))
+                raise IBMQBackendError('Failed to get job "{}": '
+                                       'job does not belong to backend "{}".'
+                                       .format(job_id, job_info['backend']['name']))
             if 'error' in job_info:
                 raise IBMQBackendError('Failed to get job "{}": {}'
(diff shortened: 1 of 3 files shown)


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -193,7 +193,10 @@ def retrieve_job(self, job_id):
             IBMQBackendError: if retrieval failed
         """
         try:
-            job_info = self._api.get_status_job(job_id)
+            job_info = self._api.get_job(job_id)
+            if job_info['backend']['name'] != self.name():
+                raise IBMQBackendError('Job "{}" belongs to backend "{}", not "{}".'
```
diego-plan9: Can you tune the behavior when the job does not belong to the backend, to follow #11 more closely? In particular:
* can you add a `warnings.warn()` between line 197 and 198, that informs the user that the job belongs to another backend (something along a slightly more user-facing version of  `'Job "{}" belongs to backend "{}", not "{}".` - the purpose is to give information to the user so they can correct the situation)
* can you tweak the `IBMQBackendError` message to keep it consistent with the rest of the exceptions raised by the function, as in `'Failed to get job "123": somethingaboutwrongbackend'`?

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -193,7 +193,10 @@ def retrieve_job(self, job_id):
             IBMQBackendError: if retrieval failed
         """
         try:
-            job_info = self._api.get_status_job(job_id)
+            job_info = self._api.get_job(job_id)
```
diego-plan9: This is a self-reminder that, in this PR, the endpoint called was changed as `get_status_job` does not support field inclusion yet, so the backend name is not provided. This is meant to be temporary - hopefully it will be implemented in the API soon, and then we can go back to using `get_status_job`, which is more efficient.

qiskit/providers/ibmq/ibmqbackend.py
Max-Reuter: Should there be a difference between the warning and the error message?

diego-plan9: Ideally, the message for the warning should aim for being informative and guide the user gently to solving the situation, which probably means might be longer - the one in in the exception can be a bit more compact, as it is mostly for developers which hopefully have more insight. Don't fret too much about them!

Issue #7, opened 2019-01-07
Move changes from qiskit-api-sdk branch


diego-plan9:
### Information

From 93, some items were not moved from the experimental qiskit-api-sdk branch (basically the unchecked boxes at https://github.com/Qiskit/qiskit-terra/issues/1175). We should port the following items to the IBMQ Provider (from the `qiskit-api-sdk` branch):
  * [x] accepting `include` in `get_job()` and `get_jobs()`, from [here](https://github.com/Qiskit/qiskit-api-py/commit/fa3a3cffa23cba7d9f22f7febfcdd5ec5cedf0f5)
  * [x] add `job.properties()` method, making use of the new connector method parameters
  * [ ] device name mangling, from [here](https://github.com/Qiskit/qiskit-api-py/commit/0f44fbcf3b5c3b8841f1c1607e8a7d8e1f1e6b45).

diego-plan9:
After #8 and #12, there is one test failing in travis for premium: 
https://travis-ci.org/Qiskit/qiskit-ibmq-provider/jobs/480389155#L781
`test_get_job_includes`
This will get solved in the next round of API updates.

diego-plan9:
Closed as probably stale or handled in other issues - the main remaining bit is the device name mangling, which seems will not be tackled any time soon.

PR #35, opened 2019-02-11
Add support for job.properties()

diego-plan9:
### Summary

Related to #7 

Add a new `IBMQJob.properties()` method, which internally calls a new endpoint that is now available in the API.


ajavadia:
can we add a changelog please? this is great.

PR #48, opened 2019-02-28
Fix job `error_message`

### LINKED ISSUE #10: error_message() does not work
running 

```
job = backend.retrieve_job(ran_job.job_id())
```

on a previous job that had an error. Followed by

```
job.error_message() 
```

returns no message

delapuente:
Fix #10 

When a job fails, requesting the result to the API returns the error details.

### DIFF
--- a/qiskit/providers/ibmq/ibmqjob.py
+++ b/qiskit/providers/ibmq/ibmqjob.py
@@ -169,9 +169,39 @@ class IBMQJob(BaseJob):
         self._future = None
         self._api_error_msg = None
 
+    def qobj(self):
+        """Return the Qobj submitted for this job.
+
+        Note that this method might involve querying the API for results if the
+        Job has been created in a previous Qiskit session.
+
+        Returns:
+            Qobj: the Qobj submitted for this job.
+        """
+        if not self._qobj_payload:
+            # Populate self._qobj_payload by retrieving the results.
+            self._wait_for_job()
+
+        return Qobj(**self._qobj_payload)
+
+    def properties(self):
+        """Return the backend properties for this job.
+
+        Returns:
+            BackendProperties: the backend properties used for this job.
+        """
+        self._wait_for_submission()
+
+        properties = self._api.job_properties(job_id=self.job_id())
+
+        return BackendProperties.from_dict(properties)
+
     # pylint: disable=arguments-differ
     def result(self, timeout=None, wait=5):
-        """Return the result from the job.
+        """Return the result from the job, consuming the job.
+
+        An attempt to retrieve the results of a failed job will raise an
+        exception.
 
         Args:
            timeout (float): number of seconds to wait for job
@@ -181,41 +211,21 @@ class IBMQJob(BaseJob):
             qiskit.Result: Result object
 
         Raises:
-            JobError: exception raised during job initialization
+            JobError: if attempted to recover a result on a failed job.
         """
-        job_response = self._wait_for_result(timeout=timeout, wait=wait)
-        return Result.from_dict(job_response['qObjectResult'])
-
-    def _get_error_details(self):
-        error_result = self._wait_for_result(expect_error=True)
-        results = error_result['qObjectResult']['results']
-        if len(results) == 1:
-            return error_result['qObjectResult']['results'][0]['status']
-
-        error_list = []
-        for index, result in enumerate(results):
-            if not result['success']:
-                error_list.append('Experiment {}: {}'.format(index, result['status']))
-
-        error_report = 'The following experiments failed:\n{}'.format('\n'.join(error_list))
-        return error_report
-
-    def _wait_for_result(self, timeout=None, wait=5, expect_error=False):
-        self._wait_for_submission(timeout)
-
-        try:
-            job_response = self._wait_for_job(timeout=timeout, wait=wait)
-            if not self._qobj_payload:
-                self._qobj_payload = job_response.get('qObject', {})
-        except ApiError as api_err:
-            raise JobError(str(api_err))
+        self._wait_for_completion(timeout=timeout, wait=wait)
 
         status = self.status()
-        if not expect_error and status is not JobStatus.DONE:
+        if status is not JobStatus.DONE:
             raise JobError('Invalid job state. The job should be DONE but '
                            'it is {}'.format(str(status)))
 
-        return job_response
+        try:
+            job_response = self._get_job()
+        except ApiError as error:
+            raise JobError(str(error))
+
+        return Result.from_dict(job_response['qObjectResult'])
 
     def cancel(self):
         """Attempt to cancel a job.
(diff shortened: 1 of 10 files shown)


qiskit/providers/ibmq/ibmqjob.py
```diff
@@ -271,8 +285,7 @@ def status(self):
         elif 'ERROR' in api_job['status']:
             # Error status are of the form "ERROR_*_JOB"
             self._status = JobStatus.ERROR
-            # TODO: This seems to be an inconsistency in the API package.
-            self._api_error_msg = api_job.get('error') or api_job.get('Error')
+            self._api_error_msg = self._get_error_details()
```
diego-plan9: This would end up implying a call to `api.get_job()`, and fetching the results, right? One of the constraints introduced previously was that that `job.status()` should explicitly _not_ consume results (ie. only call `api.get_status_job()`, as in some premium configurations, the results can only be read once, and are deleted after the first read).

For consistency and avoiding that scenario (even if errored jobs might behave slightly different), can you move the call to `job.error_message()` or a similar solution that deals with "read-once" results - perhaps populating `self._api_error_msg` only the first time it is called or a similar measure, etc.

qiskit/providers/ibmq/ibmqjob.py
```diff
@@ -186,7 +186,21 @@ def result(self, timeout=None, wait=5):
         job_response = self._wait_for_result(timeout=timeout, wait=wait)
         return Result.from_dict(job_response['qObjectResult'])
 
-    def _wait_for_result(self, timeout=None, wait=5):
+    def _get_error_details(self):
```
diego-plan9: Can you add a docstring - and while you are at it, add it also to `_wait_for_result`, taking advantage of the momentum :grin: ? They are the only two methods missing on the class, and it would be great to have all the class documented.

qiskit/providers/ibmq/ibmqjob.py
```diff
@@ -197,7 +211,7 @@ def _wait_for_result(self, timeout=None, wait=5):
             raise JobError(str(api_err))
 
         status = self.status()
-        if status is not JobStatus.DONE:
+        if not expect_error and status is not JobStatus.DONE:
```
atilag: What about naming `expect_error` as `expect_job_error` or `expect_general_error`? it's more verbose, but can't be misinterpreted with: "I'm expecting errors, even though they are coming from each of the experiments individually"

delapuente: `expect_error` is no longer needed. I think this belongs to an outdated commit.

qiskit/providers/ibmq/ibmqjob.py
```diff
@@ -186,7 +186,21 @@ def result(self, timeout=None, wait=5):
         job_response = self._wait_for_result(timeout=timeout, wait=wait)
         return Result.from_dict(job_response['qObjectResult'])
 
-    def _wait_for_result(self, timeout=None, wait=5):
+    def _get_error_details(self):
+        error_result = self._wait_for_result(expect_error=True)
+        results = error_result['qObjectResult']['results']
+        if len(results) == 1:
```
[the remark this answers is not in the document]
delapuente: I can change this but since this is the usual case in Terra, I preferred to simply return the error message without the report.

delapuente:
@diego-plan9, this is ready for another review.

qiskit/providers/ibmq/ibmqjob.py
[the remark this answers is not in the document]
diego-plan9: Actually, I'd go with @atilag's suggestion - seems more homogeneous (non-blocking)

qiskit/providers/ibmq/ibmqjob.py
```diff
@@ -281,7 +303,22 @@ def status(self):
         return self._status
 
     def error_message(self):
-        """Return the error message returned from the API server response."""
+        """Provide details about the reason of failure.
+
+        Calling this method if the job status is ERROR, consumes the job.
+
+        Returns:
+            str: An error report if the job errored or None elsewhere.
```
diego-plan9: Minor nit - `otherwise` instead of `elsewhere`?

qiskit/providers/ibmq/ibmqjob.py
```diff
@@ -398,32 +470,22 @@ def _wait_for_submission(self, timeout=60):
                 self._api_error_msg = str(submit_info['error'])
                 raise JobError(str(submit_info['error']))
 
-    def qobj(self):
-        """Return the Qobj submitted for this job.
-
-        Note that this method might involve querying the API for results if the
-        Job has been created in a previous Qiskit session.
+    def _build_error_report(self, results):
+        """Build the error report.
```
diego-plan9: Can you also document `results`?

qiskit/providers/ibmq/ibmqjob.py
```diff
@@ -281,7 +303,22 @@ def status(self):
...
+        Returns:
+            str: An error report if the job errored or None elsewhere.
+        """
+        self._wait_for_completion()
+        if self.status() is not JobStatus.ERROR:
+            return None
+
+        job_response = self._get_job()
+        results = job_response['qObjectResult']['results']
+        if not self._api_error_msg:
```
diego-plan9: Can you move this `if` block higher (before `_get_job()`) and adjust accordingly (as in, probably including the previous two lines in an `else` or similar)? Otherwise it seems it will attempt to consume the job twice, and the function might error out at that point ungracefully.

qiskit/providers/ibmq/ibmqjob.py
```diff
@@ -281,7 +303,22 @@ def status(self):
...
+            self._api_error_msg = self._build_error_report(results)
```
diego-plan9: Probably minor - but in theory there might be a case where the job errors out at a "general" level, without specific errors in the experiments (as per the `JobStatus` definition), and if that happens it seems the error report will likely fail or not provide meaningful information. Non-blocking as I think in practice it is an unlikely situation, but can you follow up on this case or keep an eye if this happens?

delapuente: I had it on my radar (and I'm still trying to force the situation) but I'll document it in a new issue. :+1:

delapuente:
@diego-plan9 ready again!

delapuente:
@diego-plan9 I've added the changelog but decided not to expose "consuming the job" concept in the documentation. Instead, I've modified and homogenized ``error_message()`` and ``result()`` to allow more than one calls by caching the result. The interaction with the job states, although complex, it is the expected so I don't think providing extra information would be of any help.

qiskit/providers/ibmq/ibmqjob.py
```diff
@@ -220,12 +221,11 @@ def result(self, timeout=None, wait=5):
...
 
-        try:
+        if not self._result:
             job_response = self._get_job()
-        except ApiError as error:
-            raise JobError(str(error))
+            self._result = Result.from_dict(job_response['qObjectResult'])
 
-        return Result.from_dict(job_response['qObjectResult'])
+        return self._result
```
diego-plan9: Can you revise the docstring? It has fallen out of sync with your latest comment.

PR #51, opened 2019-03-08
Return string for status rather than JobStatus

nonhermitian:
### Summary

The main reasons as to why this is desired are here: https://github.com/Qiskit/qiskit-terra/issues/1903#issuecomment-470855012

I will also add that when getting a Job from a backend and filtering by status, the status is a string, not `JobStatus` instance, which is also in line with the change presented here.

### Details and comments

I also do not see why `queue_position` and `error_message` should not be incorporated into the status message itself.  As an user I obviously want to know this information, and having to make two separate calls to get it doesn't make sense, but that can easily be added if so desired.

### DIFF
--- a/qiskit/providers/ibmq/ibmqjob.py
+++ b/qiskit/providers/ibmq/ibmqjob.py
@@ -59,7 +59,7 @@ class IBMQJob(BaseJob):
 
         try:
             job_status = job.status() # It won't block. It will query the backend API.
-            if job_status is JobStatus.RUNNING:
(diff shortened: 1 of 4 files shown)


delapuente:
Reasons for why the status is an enumeration are also explained in https://github.com/Qiskit/qiskit-terra/issues/1903#issuecomment-470611567 However, your points regarding filtering, queue position and error message are fair and could be tracked in different issues.

nonhermitian:
Thank you for again summarily closing something I was working on.

Issue #34, opened 2019-02-08
Revise authentication parameters for connector methods


diego-plan9:
### What is the expected behavior?

Currently, most of the methods of `IBMQConnector` that interact with and endpoint accept the credentials information (`hub`, `group`, `user_id`, etc) as parameters, and perform a similar set of operations at the beginning of the function:

```python
        if access_token:
            self.req.credential.set_token(access_token)
        if user_id:
            self.req.credential.set_user_id(user_id)
        if not self.check_credentials():
            raise CredentialsError('credentials invalid')
```

However, the credentials information is already present in `self.config`, which makes the parameters and some checks redundant (and actually, the parameters are unused by the rest of the codebase). It seems a good candidate for simplification, by just removing the parameters and fetching the needed information from `self.config` directly.

jwoehr:
It _does_ look like those parameters are only used within the package itself.

diego-plan9:
Yes - the issue is mostly about reorganizing things a bit to reduce complexity, taking advantage of that fact indeed!

jwoehr:
A little confused here!

Qiskit 8.0 has IBMQConnector.get_job() as:
`def get_job_url(config, hub=None, group=None, project=None):`

Current has IBMQConnector.get_job() as:
`def get_job(self, id_job, hub=None, group=None, project=None,
                exclude_fields=None, include_fields=None,
                access_token=None, user_id=None)`

So apparently fields were still being added to the code until recently.

Which of these fields is it desirable to eliminate and always fetch from the config? All of them?

- hub
- group
- project
- access_token
- user_id

??

diego-plan9:
> Which of these fields is it desirable to eliminate and always fetch from the config? All of them?

The ones that can be inferred directly from `self.config` and are only used in the `if` "block" mentioned on the issue description  - actually, the list you posted right after your sentence (`hub`, `group`, `project`, `access_token`, `user_id`) looks about right in the general case!

In the example you mentioned (`IBMQConnector.get_job()`), `id_job`, `exclude_fields` and `include_fields` are parameters that are not related to authentication and are used in order to provide the functionality required by the method - those are good examples of parameters that should _not_ be removed :+1:

jwoehr:
So for example this:
```
def get_job_url(config, hub=None, group=None, project=None):
    """Return the URL for a job."""
    hub = config.get('hub', hub)
    group = config.get('group', group)
    project = config.get('project', project)

    if hub and group and project:
[...]

PR #56, opened 2019-03-27
Many superfluous kwargs removed.

jwoehr:
### Summary

Address Issue #34 

### Details and comments

I have tested with my own published code, but not fully run the extant tests as there are issues with running them in my dev environment.

### DIFF
--- a/qiskit/providers/ibmq/api/ibmqconnector.py
+++ b/qiskit/providers/ibmq/api/ibmqconnector.py
@@ -50,7 +50,7 @@ def get_backend_defaults_url(config, backend_type):
     return '/Backends/{}/defaults'.format(backend_type)
 
 
-def get_backends_url(config, hub, group, project):
+def get_backends_url(config, hub=None, group=None, project=None):
     """Return the URL for a backend."""
     hub = config.get('hub', hub)
     group = config.get('group', group)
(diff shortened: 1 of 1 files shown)


jwoehr:
One function outside the class needs its args changed to kwargs. Have done that and pushed, now to re-pull.

jwoehr:
Try again with new change to non-class function

qiskit/providers/ibmq/api/ibmqconnector.py
```diff
@@ -50,7 +50,7 @@ def get_backend_defaults_url(config, backend_type):
     return '/Backends/{}/defaults'.format(backend_type)
 
 
-def get_backends_url(config, hub, group, project):
+def get_backends_url(config, hub=None, group=None, project=None):
```
diego-plan9: Actually, seeing that the only place where this function is called (at `IBMQConnector.available_backends()`) is done with only the `config` parameter, can you remove `hub`, `group`, and `project` from this function as well, adjusting the checks inside accordingly, as in:

```python
def get_backends_url(config):
    """Return the URL for a backend."""
    hub = config.get('hub', None)
    group = config.get('group', None)
    project = config.get('project', None)

    if hub and group and project:
    ...
```

And while we are at it, can you do the same for `get_backend_properties_url()` and `get_job_url()`? This would make all the four "URL" methods consistently depend only on `config` for authentication-related values.

jwoehr: Thanks for all the tips @diego-plan9  ... will take into account in future.
Committed changes and pushed.

diego-plan9: Nice, thanks!

Issue #52, opened 2019-03-11
Make use of qiskit.test.providers.ProviderTestCase


diego-plan9:
### What is the expected behavior?

Since a number of commits ago, terra includes [some classes for helping testing providers and backends](https://github.com/Qiskit/qiskit-terra/tree/master/qiskit/test/providers). Ideally this provider should make use of those for consistency - it involves changing the class the relevant test cases inherit from, and removing individual tests as necessary.

jwoehr:
So the new class would be `class IBMQProviderTestCase(ProviderTestCase)` ?

diego-plan9:
Exactly - the idea is to make use of inheriting from `BackendTestCase` and `ProviderTestCase`, and in the process revise the tests of each test case individually, removing those that are already covered (as in, they test "the same thing") by those base classes (as they include a minimal set of tests) to avoid duplication.

jwoehr:
I'm working on this now.

diego-plan9:
Yes, the decorators can and should be used.

jwoehr:
`requires_qe_access` is not a big problem. But qiskit/test/providers/backend.py seems oriented to a single backend, rather than a list of backends. `qiskit-ibmq-provider/test/ibmq/test_ibmq_backends.py` is designed around such a list. Perhaps I should factor out the testing of an individual backend from `/test_ibmq_backends.py` and not make that file itself inherit `IBMQBackendTestCase`.

jwoehr:
Perhaps  `qiskit-ibmq-provider/test/ibmqtest_ibmq_backends.py` should enumerate the QE backends and dynamically create a  `class IBMQBackend_backendname_TestCase(IBMQBackendTestCase)` for each enumerated instance and `run()` it ... `IBMQBackendTestCase` I have already defined as
```
from qiskit.test.providers import BackendTestCase
class IBMQBackendTestCase(BackendTestCase):
```
... etc.

jwoehr:
I understand the structure of the test cases better now and am proceeding with coding.

diego-plan9:
Great, looking forward to the results! Feel free to handle both in different PRs (ie. one for the provider test, and another one for the backend tests, which are probably the most complex as you mention) if it makes it more manageable - specially considering we will be having a number of changes in the upcoming couple of weeks and might involve a small burst of activity.

jwoehr:
[...]

Issue #53, opened 2019-03-12
Avoid deprecation warnings coming from terra


diego-plan9:
### What is the expected behavior?

The test suite currently emits several deprecation warnings, mostly due to the use of `qiskit.compile()`, which is to be deprecated in 0.9. We should move the tests to use the new transpilation workflow accordingly.

jwoehr:
Is this mostly a case of replacing phrases like:

test_ibmq_connector.py:        `qobj = compile(self.qc1, backend=backend, seed=self.seed, shots=1)`

with appropriate calls to `qiskit.transpiler.transpile()` using appropriate arguments?

diego-plan9:
Yes - it should be a matter of running the tests, capturing the deprecation warnings (most test runners will be able to provide a list of them, usually at the end of the output or during the tests), and tracing down which line of the test is causing the deprecation warnings.

I'd indeed expect that most of them would be due to `compile()` - for those, it would mean replacing those lines with calls to the appropiate non-deprecated function (usually the warnings themselves provide recommendation on how to adapt to the new behaviour, but might involve digging a bit into terra and being up to date with the recent `master` changes).

jwoehr:
I am not quite clear on how the test harness works to run the tests. What should I explore, please?

jwoehr:
(I am thinking about both this issue and #52 )

jwoehr:
Yes, separate PRs. Will explore Makefiles for more insight and tackle both #52 and #53 though possibly not this week yet.

jwoehr:
Working on this one.

Attributes have changed. In the case of kwargs like `shots=` I have noted them and dropped them from the argument list when switching from `compile()` to `transpile`.

There are other attributes I haven't quite figured out what to do about. Here are two examples of the errors:

```

**Comment by jwoehr** [comment]
Here's another attrib `as_dict`

```
======================================================================
ERROR: test_api_run_job_fail_backend (test.ibmq.test_ibmq_connector.TestIBMQConnector)
Test running a job against an invalid backend.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jax/work/QISKit/DEV/qiskit-terra/qiskit/test/decorators.py", line 178, in _wrapper
    return decorated_func(self, *args, **kwargs)
  File "/home/jax/work/QISKit/DEV/qiskit-ibmq-provider/test/ibmq/test_ibmq_connector.py", line 85, in test_api_run_job_fail_backend
[...]

PR #68, opened 2019-04-30
Use qiskit test providers 52

jwoehr:
### Summary

Address #52 

### Details and comments

This is a start. The rebases are not ideal for the intended purpose. I have taken one approach. You may wish to  provide guidance for more work on this.


jwoehr:
This is failing because I have created a new directory. Not sure how to make that work with a pull request.

diego-plan9:
Thanks @jwoehr !

> This is failing because I have created a new directory. Not sure how to make that work with a pull request.

Can you try placing the files currently at:
```
/qiskit/test/providers/ibmq/
```
inside `/test/ibmq` (both paths on the comment referring to the root dir of the repo)? The way it is currently (as of a5f2230) it will cause a clash between module names.

Also, don't worry too much about rebasing, etc - we can reconcile back with `master` when the PR is ready!

jwoehr:
Whee! It passes.

diego-plan9:
Thanks for the patience, @jwoehr! Just a comment to let you know the PR did not go unnoticed - the "rush period" has arrived and is requiring quite a lot of energy, but we will review this PR once things quiet down a bit (hopefully later this week).

jwoehr:
> Thanks for the patience, @jwoehr! Just a comment to let you know the PR did not go unnoticed -

No problem. You might want some changes on this, so I look forward to your review.

jwoehr:
Merged #52 with the accepted code from #53

jwoehr:
Odd, there see to be three checks now, and all have fully passed, but it thinks one is not complete.

diego-plan9:
Apologies for taking longer than usual to provide feedback, @jwoehr ! As you might have guessed by the number of issues and PRs created on the repo, we are in the midst of a quite high-activity period in different fronts, but finally managed the time to take a deeper look at your PR.

If I'm reading it right (thanks for the detailed comments in the new files!), most of the effort comes from trying to solve the "dynamically create and run a `BackendTestCase` for the discovered backends", which indeed is not well supported at the moment due to the nature of this particular provider (as in, you don't know what backends exists until the provider is instantiated, which is more complex that local providers).

I think the solution you achieved is quite creative, although it unfortunately has some inconvenient side effects. They are mostly derived from the `test_backends_IBMQBackendTestCase` implementation, combined with the way the Python test framework is structured:

```
        for backend in self.provider.backends():
            TextTestRunner(verbosity=2).run(
                TestLoader().loadTestsFromTestCase(self.gen_backend_test_class(backend))
            )
```

Usually creating a `TestRunner` is responsability of the user (as in, the user will launch the tests through a tool, which can be `python -m unittest` or other, such as the IDE unittest launcher, `pytest`, etc): a `SomethingTestRunner` is created, which then compiles the test suite, runs the individual tests, etc. These lines potentially make a new test runner to be created when running a test from the original test runner used by the user, which can have a number of oddities in the flow (for example, I ended up having all the tests for a backend as a _single_ pytest test with `unittest` output while running the full suite with `pytest`, and was unable to run the test individually with the Pycharm stock runner).

We will probably be able to take some ideas from your approach (such as customizing the loading, or seeing if we can move the dynamic loading under a more controlled layer), but it will probably require some exploratory effort - and I'm thinking a more sensible approach would be to split the PR in parts, so we can advance in each front individually.

What I'm proposing, re-reading #52, is:
* make a PR that only deals with making use of `ProviderTestCase`. Hopefully this will be quite tiny and quick to review and merge.
* follow up with another PR that evaluates with making use of `BackendTestCase`, and in turn deal with the dynamic nature of the backends. This one would actually not be high priority - more of a "wishlist" of sorts.

Would it make sense? Thanks for the effort put into the issue, which turned out to be a bit more tricky than expected!

diego-plan9:
On a minor note ...

> Odd, there see to be three checks now, and all have fully passed, but it thinks one is not complete.

Yes, we had a small hiccup with travis during the time your last merge commit was made - the next one will hopefully have the usual two checks instead of three.

jwoehr:
> Apologies for taking longer than usual to provide feedback, @jwoehr !

No problem, we're all keeping busy :)

> (thanks for the detailed comments in the new files!)

You're welcome.

> Usually creating a `TestRunner` is responsability of the user

Yes, I couldn't figure out how to simply chain the dynamically generated test entity to the calling test entity. Spent some hours reading the Python docs for `unittest` but did not succeed in untangling!



> What I'm proposing, re-reading #52, is:
> 
>     * make a PR that only deals with making use of `ProviderTestCase`. Hopefully this will be quite tiny and quick to review and merge.
> 
>     * follow up with another PR that evaluates with making use of `BackendTestCase`, and in turn deal with the dynamic nature of the backends. This one would actually not be high priority - more of a "wishlist" of sorts.

> Would it make sense? Thanks for the effort put into the issue, which turned out to be a bit more tricky than expected!

Yes this makes sense. Change the issue and I'll work on it.

jwoehr:
> On a minor note ...
> 
> > Odd, there see to be three checks now, and all have fully passed, but it thinks one is not complete.
> 
> Yes, we had a small hiccup with travis during the time your last merge commit was made - the next one will hopefully have the usual two checks instead of three.

Okay, so I was not hallucinating :)

jwoehr:
Closed in view of revised #52

PR #78, opened 2019-05-03
Add IBMQClient for api version 2

diego-plan9:
### Summary

Add a `IBMQClient` as the equivalent of `IBMQConnector` for the version 2 of the API. The classic code path (ie. using the same `IBMQConnector`) will still be used after this PR for all URLs except for the ones that belong to the authentication server.

### Details and comments

The `IBMQClient` aims to be a refactor of the previous connector, with the goal of being more modular and adjust quicker to API changes. This PR is meant to set the basis, but includes:
* understanding of the new login for API version 2 (auth server -> api server)
* use of `Request.Session` for retrying and passing session attributes.
* thin the `IBMQClient` itself, delegating the logic and parsing into smaller classes (with the aim of making them interchangeable, keeping `IBMQClient` public interface stable.
* interface compatibility with `IBMQConnector`.

PR #79, opened 2019-05-04
Add Qcircuits support

### LINKED ISSUE #63: Add support to new qc endpoint

Provide programmatic access to the new endpoint.

diego-plan9:
### Summary

Fixes #63

Add support for Qcircuits via a `QcircuitsManager` that is exposed as an attribute of `qiskit.IBMQ`. As with this round of pull requests related to version 2 (#78), the internal implementation is subjected to receive some tweaks short-term and might contain some workarounds while dealing with the transition between api versions.


qiskit/providers/ibmq/qcircuits/__init__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+
```
jaygambetta: can we have the folder circuits not circuits

qiskit/providers/ibmq/qcircuits/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Module for interacting with QCircuits."""
+
+from .manager import QcircuitsManager
```
jaygambetta: This should also be CircuitsManager

qiskit/providers/ibmq/qcircuits/__init__.py
diego-plan9: Sure - note this file and class is not really exposed to the user (the real binding of the user interface happens in https://github.com/Qiskit/qiskit-ibmq-provider/pull/79/files#diff-fed302e3beba659413d59e360ed8dca3R49), but I'm good changing it!

qiskit/providers/ibmq/api/ibmqconnector.py
```diff
@@ -385,6 +385,33 @@ def available_backends(self):
 
         return response
 
+    def qcircuit_run(self, name, **kwargs):
```
jaygambetta: I really want to drop the q

diego-plan9: Haha noted :+1:

Issue #87, opened 2019-05-07
Revise test_qobj_headers_in_result_sims


diego-plan9:
### What is the expected behavior?

After #86, the `test_qobj_headers_in_result_sims` was marked as expected failure, as passing `object_header` to `assemble()` no longer preserves the header pristine:

```
custom_qobj_header = {'x': 1, 'y': [1, 2, 3], 'z': {'a': 4}}
qobj_header = QobjHeader.from_dict(custom_qobj_header)
qobj = assemble(circuits, backend=backend, qobj_header=qobj_header)
```

```
> print(qobj.header.as_dict())
{'x': 1, 'backend_name': 'ibmq_qasm_simulator', 'backend_version': '0.1.547', 'y': [1, 2, 3], 'z': {'a': 4}}
```

We should revise if this is by design and adjust the test or terra accordingly (for the test, simply reverting the order and updating `qobj.header` directly after `assemble()` would work). `test_qobj_headers_in_result_devices` might need similar measures.

jwoehr:
I can adjust the tests accordingly, if desired. But is `assemble()` really doing what it should?
- Should the qobj dictionary be an ordered dict, and should `assemble()` likewise preserve and yield an ordered dict?
- Or should all consumers not presume any order to qobj's dict?

diego-plan9:
It seems indeed be a (good) question better handled at terra - the order of the dicts should not be taken into account (as the specs don't define an order for the header), but `assemble()` is appending two extra keys to the original dict (which seems to be by design, but there is a small debate regarding it).

For the provider, if you are up for tackling this issue, can you go for modifying directly `qobj.header` _after_ `assemble()` (might be just reverting the order of the lines)? This would actually resemble better the original semantics of those two tests: they are actually testing whether the device returns the headers in `qobj.header`, regardless of how the `Qobj` is constructed.

jwoehr:
Yes, I believe it should be that simple, changing order of operations. Looked at that when I first noticed the phenomenon in #53 but was not quite sure of the intent.

jwoehr:
Hmm, I can't figure out how to get my fork in sync ... everything says they're the same, but when I check out master I don't see the `@expected_fail` annotation.

diego-plan9:
Fixed via #94.

PR #94, opened 2019-05-09
Revise test qobj headers in result sims 87

jwoehr:


jwoehr:
Answers #87

test/ibmq/test_ibmq_backends.py
```diff
@@ -105,11 +103,10 @@ def test_qobj_headers_in_result_sims(self, qe_token, qe_url):
             with self.subTest(backend=backend):
                 circuits = transpile(self.qc1, backend=backend)
 
-                # Update the Qobj header.
-                qobj_header = QobjHeader.from_dict(custom_qobj_header)
                 # TODO: assemble appends extra keys to the header in terra 0.8.
```
diego-plan9: Can you also delete the line with the TODO?

jwoehr: Done.

diego-plan9:
Thanks @jwoehr ! It's looking good - can you also take the chance to revise `test_qobj_headers_in_result_devices`? It should probably end up containing the same lines of code (except for the `simulator=False`).

test/ibmq/test_ibmq_backends.py
diego-plan9: Great, thanks!

jwoehr:
editing and testing, just a moment ...

jwoehr:
(That should have bombed, I think it has been skipped as a "slow test")

diego-plan9:
> (That should have bombed, I think it has been skipped as a "slow test")

Yes - just in case you have not encountered it before, locally you can enable the running of slow tests via an environment variable `QISKIT_TESTS=run_slow`. However, note that they are indeed slow (the cause of slowness is that they use the real devices, which involves having to wait in the queue for minutes in most cases) - I'd recommend using it with caution (as in, only using it for launching a particular test, not the full suite by default).

jwoehr:
I'll try that now for good luck, since we changed a slow test.

jwoehr:
Well, that was interesting running the slow tests. The tests I have edited ran okay, but another one failed:
```
======================================================================
FAIL: test_run_device (test.ibmq.test_ibmq_job.TestIBMQJob)
Test running in a real device.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jax/work/QISKit/DEV/qiskit-terra/qiskit/test/decorators.py", line 72, in _wrapper
    return func(*args, **kwargs)
  File "/home/jax/work/QISKit/DEV/qiskit-terra/qiskit/test/decorators.py", line 178, in _wrapper
    return decorated_func(self, *args, **kwargs)
  File "/home/jax/work/QISKit/DEV/qiskit-ibmq-provider/test/ibmq/test_ibmq_job.py", line 96, in test_run_device
    self.assertDictAlmostEqual(counts_qx, counts_ex, shots * 0.1)
AssertionError: ('10': 208 != 0), ('11': 278 != 512.0), ('01': 135 != 0), ('00': 403 != 512.0) within 102.4 delta

----------------------------------------------------------------------
```

Issue #91, opened 2019-05-09
Add license check for new API


diego-plan9:
### What is the expected behavior?

Follow up to #78.

When using the new API through the authentication server, if the user did not accept the license, the error message is generic. For clarity, we should handle it as a special case and present the user with a clearer error message, similar to the classic API.

diego-plan9:
Closed via #111

Issue #106, opened 2019-05-14
Unify handling of exceptions


diego-plan9:
### What is the expected behavior?

In the `IBMQConnector` used for communicating with the classic API, the conventions for raising `Exceptions` grew organically - the raising of an exception that comes from an invalid HTTP response (400, etc) usually comes from [a deep `_response_good()` check]( https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/api/utils.py#L368) and under a generic `ApiError` which partially hiddens the root cause, merging several sources of exceptions together and relying on two message strings (`usr_msg`, `dev_msg`) for interpreting the failure. As a whole, it would be nice to be able to revise the way we are using exceptions in order to take full advantage.

A different direction is being taken during the `api_v2` contributions, but it should be revised when all the pieces are in place to make sure it is consistent and unified. Some notes and ideas can be found in the different PRs and reviews (#78, #100), and the approach is based on:
* try to take advantage of the exceptions raised by the libraries (`requests`, `websockets`) or Python directly
* instead of discarding the original exception, chain it to an ibmq-specific exception via `raise from`
* use Exception sub-classes and hierarchies to model different "types" of errors, instead of relying on specific messages
* make those message informative enough for end users, in case the exception ripples up
* ... others!

Ideally we should discuss about those guidelines for exceptions, and consolidate them (and make the pertinent changes to the code).

jyu00:
To give an example to point 3 above, `IBMQJob` raises generic `JobError` for almost all of the exceptions. It'd be nice to be able to differentiate say an exception raised during an API call, and one raised because the job failed.

PR #111, opened 2019-05-17
Raise descriptive error when login fails due to unaccepted license agreement

Max-Reuter:
### Summary
Raise a descriptive error when login fails due to an unaccepted license agreement, telling the user that they need to accept the license. See #91.


qiskit/providers/ibmq/api_v2/exceptions.py
```diff
@@ -47,3 +47,8 @@ class WebsocketAuthenticationError(WebsocketError):
 class WebsocketTimeoutError(WebsocketError):
     """Timeout during websocket communication."""
     pass
+
+
+class UnlicensedError(ApiError):
```
diego-plan9: Can you rename a bit, for clarity, to `AuthenticationLicenseError` or similar? Paving the way towards other `AuthenticationError`, and in similarity to the existing `WebsocketAuthenticationError`.

qiskit/providers/ibmq/api_v2/ibmqclient.py
```diff
@@ -64,7 +65,20 @@ def _request_access_token(self):
...
         Returns:
             str: access token.
+        Raises:
+            RequestsApiError: if the user hasn't accepted the license agreement.
         """
+        try:
+            self.client_auth.login(self.api_token)
+        except RequestsApiError as ex:
+            response = ex.original_exception.response
+            status_code = response.status_code
```
diego-plan9: Can you update the code block to something similar to:
```python
            if response.status_code == 401:
                try:
                    error_code = ...
                    message = ...
                    if error_code == ...:
                        raise AuthenticationLicenseError(ex.original_exception,
                                                         message)
                except (ValueError, KeyError):
                    # A comment saying that the response did not contain the json we were looking for blah
                    pass
```

in order to:
* take advantage that `response.status_code` is always present, so we can enter the `if` early
* prevent the block failing ungracefully via the extra `try` block: there could be cases where the response was not json, or did not contain the keys we expected. By being specific catching those [two](https://2.python-requests.org/en/master/user/quickstart/#json-response-content) [occurences](https://docs.python.org/3/library/exceptions.html#KeyError), we control what happens when that scenario is reached (in this case, we ignore them and let them bubble upwards).
* in this particular case, instead of chaining the exception, "rewrite" it as a `AuthenticationLicenseError` - mostly for compatibility until #106

qiskit/providers/ibmq/ibmqsingleprovider.py
```diff
@@ -111,7 +113,8 @@ def _authenticate(self, credentials):
             if 'License required' in str(ex):
                 # For the 401 License required exception from the API, be
                 # less verbose with the exceptions.
-                root_exception = None
+                raise UnlicensedError("Couldn't connect to IBMQ server: {0}"
```
diego-plan9: Instead of raising the new `UnlicensedError` exception here, in this case we want to keep this `IBMQSingleProvider` method only emitting `ConnectionError`s upwards. However, what we can do is make use of the new class for identifying the situation in a cleaner way than before, via an extra `except` block:

```python
        # Check if the URL belongs to auth services of the new API.
        try:
            ...existing code
        except  AuthenticationLicenseError as ex:
            # now we can be sure that if we get here, is because of license problems!
            raise ConnectionError(.....) from None
        except Exception as ex:
            ....
```

the `from None` bit is for preventing too verbose output reaching the user - again, partly pending an overhaul during #106 .

qiskit/providers/ibmq/ibmqsingleprovider.py
```diff
@@ -75,6 +76,7 @@ def _authenticate(self, credentials):
             IBMQConnector: instance of the IBMQConnector.
         Raises:
             ConnectionError: if the authentication resulted in error.
+            UnlicensedError: if the user hasn't accepted the license agreement.
```
diego-plan9: Can you remove this line, as part of the renaming cleanup?

Issue #97, opened 2019-05-10
Enable proxy support for new API


diego-plan9:
### What is the expected behavior?

As part of the transition, support for the proxy configuration keys needs to be adjusted for the new API v2 - the basics are in place for the `Session`, but it needs to be enabled fully. Ideally, we should also include some tests for the feature, probably with the help of some lightweight proxy server in the CI.

On the wishlist, testing support for NTLM proxies would be high on the list - although it might not be technically feasible.

diego-plan9:
Fixed by #119

yunfan:
do you support socks5 proxy?

PR #119, opened 2019-05-23
Enable proxy support

Max-Reuter:
### Summary
Enable proxy support for the new API. See #97.


diego-plan9:
It seems `pproxy` dropped support for Python 3.5 after version 1.2.2 - which has been a while back! Can you either:
* update the code to work with 1.2.2 (it seems the command line arguments and module structure has changed since them)
* reevaluate and try to find another Python proxy server instead (https://github.com/abhinavsingh/proxy.py might do the trick)

The reason why the initial issue suggested a _Python_ proxy server was for making it easier to perform assertions or tweaks (mock some methods, etc) in order to be able to test that the requests went through the proxy in a programmatic way: instead of calling an executable blindly via `subprocess` , we should be able to have finer control.

For example, with proxy.py, you can do:
```
server = proxy.HTTP(port=PORT)
server.run()
```

and similarly with `pproxy` as well.

test/ibmq/test_proxies.py
```diff
@@ -0,0 +1,56 @@
...
+    @requires_qe_access
+    @requires_new_api_auth
+    def test_proxies(self, qe_token, qe_url):
+        """Test IBMQClient proxy connection."""
+        input_proxies = {
+            'https': '{}:{}'.format(ADDRESS, PORT)
+        }
+
+        client = IBMQClient(qe_token, qe_url, input_proxies)
+        self.assertEqual(client.proxies, input_proxies)
```
diego-plan9: This `assert` does not provide enough guarantees that the requests have gone through the proxy. As mentioned on the main comment, if we run the server programmatically we should be able to perform some assertions against the server directly.

We can leave those more complex assert as a follow-up, but as a minimal version, can you make this assert dig a bit deeper by:
* making a request "manually" during this tests (by manually, I mean a request against this `IBMQClient`'s `Session`)
* check that the request contains the right `proxy_manager`

Similar to:
```
...
x = IBMQClient.client_api.session.get('/version seems a safe choice')
self.assertIn('theproxyurl', x.connection.proxy_manager.keys())
```

test/ibmq/test_proxies.py
```diff
@@ -0,0 +1,56 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+
+        # Close the mock server.
+        cls.proxy_process.kill()
+
+    @requires_qe_access
+    @requires_new_api_auth
+    def test_proxies(self, qe_token, qe_url):
```
diego-plan9: Can you also add a test for "we are trying to use a proxy that does not exist (with the wrong port, for example), and we fail as expected"?

delapuente:
Finding a non-blocking programmatic Python proxy is being trickier than expected. With this version, we are testing that requests reach the proxy. Testing if the requests abandon the proxy to go to the network is implicitly tested by the fact that the API is authenticating. However, we could also manipulate the session to check for something different thant the auth endpoint. @diego-plan9 what do you think?

test/ibmq/test_proxies.py
```diff
@@ -0,0 +1,84 @@
...
+        # Close the mock server.
+        if cls.proxy_process.returncode is None:
+            cls.proxy_process.terminate()
+
+    @requires_qe_access
+    @requires_new_api_auth
+    def test_proxies(self, qe_token, qe_url):
+        """Should reach the proxy."""
+        request = 'http auth.quantum-computing.ibm.com:443'
+        _ = IBMQClient(qe_token, qe_url, {'https': 'http://{}:{}'.format(ADDRESS, PORT)})
```
diego-plan9: Can you make this call more similar to the rest of the tests (declaring `input_proxies` and passing it as a named parameter), for readability?

qiskit/providers/ibmq/api_v2/ibmqclient.py
```diff
@@ -25,17 +25,19 @@
...
         """IBMQClient constructor.
 
         Args:
             api_token (str): IBM Q api token.
             auth_url (str): URL for the authentication service.
+            proxies (dict): proxies used in the connection.
         """
         self.api_token = api_token
         self.auth_url = auth_url
+        self.proxies = proxies
```
diego-plan9: Can we not store the attribute in the object, and instead pass it to `_init_service_clients()`? It would reflect better that is only relevant for the lower (session) layer.

test/ibmq/test_proxies.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # Close the mock server.
+        if cls.proxy_process.returncode is None:
+            cls.proxy_process.terminate()
+
+    @requires_qe_access
+    @requires_new_api_auth
+    def test_proxies(self, qe_token, qe_url):
+        """Should reach the proxy."""
+        request = 'http auth.quantum-computing.ibm.com:443'
```
diego-plan9: Can we find a slightly more descriptive name for this variable (`request` is a bit overloaded), and more importantly make its content based on `qe_url`?

Issue #113, opened 2019-05-20
KeyError in `IBMQJob::error_message()` while running a circuit that throws


atilag:
### Informations

- **Qiskit version**: 0.8
- **Python version**: All
- **Operating system**: All

### What is the current behavior?
This circuit:
```python
from qiskit import QuantumRegister, ClassicalRegister
from qiskit import QuantumCircuit, Aer, execute
from qiskit.providers.ibmq import IBMQ

qe_token = "<HIDDEN>"
qe_url =  "The.API.url"

q = QuantumRegister(50)
c = ClassicalRegister(2)
qc = QuantumCircuit(q, c)

qc.h(q[0])
qc.cx(q[0], q[1])
for k in range(2,50):
    qc.t(q[k])

IBMQ.enable_account(qe_token, qe_url)
backend = IBMQ.get_backend('ibmq_qasm_simulator')
job_sim = execute(qc, backend)
try:
  job_sim.result()
except Exception:
  pass

print(job_sim.error_message())
```
Won't execute because the number of Qubits is too high, but the IBMQ Provider seems to not handle the error properly, because if I try to get the error message like:
```
job_sim.error_message()
```
Python throws an exception complaining about the inexistence of this key: `qObjectResult`.

### Steps to reproduce the problem
Execute the provided example program

### What is the expected behavior?
 `job_sim.error_message()` should show the error message.

diego-plan9:
Thanks @atilag (nice bug report)! It seems the result of #50, which hopefully is on its way of being fixed.

diego-plan9:
Closed via #126

Issue #50, opened 2019-03-07
Review `error_message` in case of a total failure.


delapuente:
### What is the expected behavior?
PR #48 deals with an error coming from a failed experiment but it could be cases in which the job fails globally. The current implementation does not cover this case. This issue is intended to track this possible corner case.

ryan-woo:
Hi!

Backend is currently not schema compliant. Currently, you'll get a "KeyError" if an error occurs on the backend. We're fixing it.

`error_message()` (namely `IBMQConnector._get_job()`) should still check that the job it received is indeed schema compliant, so this is still an issue.

diego-plan9:
Closed by #126 - @ryan-woo , hopefully it includes a safeguard against jobs not being fully conformant to the schema: please reopen if you find the problem still persists for the cases you mention.

PR #126, opened 2019-05-26
Address `KeyError` in `IBMQJob::error_message()`.

Max-Reuter:
### Summary
Address `KeyError` in `IBMQJob::error_message()`. See #50 and #113.

### DIFF
--- a/qiskit/providers/ibmq/job/ibmqjob.py
+++ b/qiskit/providers/ibmq/job/ibmqjob.py
@@ -349,8 +349,10 @@ class IBMQJob(BaseJob):
                 self._api_error_msg = build_error_report(results)
             elif 'qasms' in job_response:
                 qasm_statuses = [qasm['status'] for qasm in job_response['qasms']]
-                self._api_error_msg = 'Job resulted in the following QASM state(s): ' \
+                self._api_error_msg = 'Job resulted in the following QASM status(es): ' \
                                       '%s.' % ', '.join(qasm_statuses)
+            else:
+                self._api_error_msg = 'An unknown error occurred.'
 
         return self._api_error_msg
 
--- a/qiskit/providers/ibmq/job/ibmqjob.py
+++ b/qiskit/providers/ibmq/job/ibmqjob.py
@@ -349,10 +349,8 @@ class IBMQJob(BaseJob):
                 self._api_error_msg = build_error_report(results)
             elif 'qasms' in job_response:
                 qasm_statuses = [qasm['status'] for qasm in job_response['qasms']]
-                self._api_error_msg = 'Job resulted in the following QASM status(es):' \
-                                      '{}'.format(', '.join(qasm_statuses))
-            else:
-                self._api_error_msg = 'An unknown error occurred.'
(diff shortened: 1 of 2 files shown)


qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -344,8 +344,15 @@ def error_message(self):
...
             job_response = self._get_job()
-            results = job_response['qObjectResult']['results']
-            self._api_error_msg = build_error_report(results)
+            if 'qObjectResult' in job_response:
+                results = job_response['qObjectResult']['results']
+                self._api_error_msg = build_error_report(results)
+            elif 'qasms' in job_response:
+                qasm_statuses = [qasm['status'] for qasm in job_response['qasms']]
+                self._api_error_msg = 'Job resulted in the following QASM status(es): ' \
+                                      '%s.' % ', '.join(qasm_statuses)
```
delapuente: Could you use [`.format()`](https://docs.python.org/3/library/string.html#string.Formatter.format) instead of `%` string formatting?

delapuente:
By the way @Max-Reuter, the fix seems solid enough, is it still `WIP`? If not, please, remove it from the short description. Also, it would be great if you could add a regression test.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -344,8 +344,13 @@ def error_message(self):
...
             job_response = self._get_job()
-            results = job_response['qObjectResult']['results']
-            self._api_error_msg = build_error_report(results)
+            if 'qObjectResult' in job_response:
+                results = job_response['qObjectResult']['results']
+                self._api_error_msg = build_error_report(results)
+            elif 'qasms' in job_response:
+                qasm_statuses = [qasm['status'] for qasm in job_response['qasms']]
+                self._api_error_msg = 'Job resulted in the following QASM status(es): ' \
+                                      '{}.'.format(', '.join(qasm_statuses))
```
diego-plan9: Can you reintroduce the third `else: ` block? It might be hard to reach, but it would guard against odd responses that do not contain `qObjectResult` or `qasms` in the content. For extra safety:

```python
else:
                self._api_error_msg = job_response.get('status', 'An unknown error occurred.')
```

(untested line!) The intent of the line is to make use of the `status` field if present (as the API might return useful information that we can present to the user other than the default test, and the `else` makes sure that we write something into `self._api_error_msg` one way or the other.

test/ibmq/test_ibmq_job.py
```diff
@@ -385,41 +383,29 @@ def test_double_submit_fails(self, qe_token, qe_url):
...
-    def test_qobj_enabled_result(self):
-        """Jobs can be retrieved."""
-        qobj = assemble(
-            transpile(self._qc, backend=self._backend), backend=self._backend)
-        job = self._backend.run(qobj)
         try:
-            job.result()
-        except JobError as err:
-            self.fail(err)
+            job_sim.error_message()
```
diego-plan9: Instead of "trying and failing", can you use an assert:

```
something = job_sim.error_message()
self.assertIn('some string that is in the error message', something)
```

PR #128, opened 2019-05-29
Move "requires_qe_access" into the package

diego-plan9:
### Summary

Due to the changes proposed in https://github.com/Qiskit/qiskit-terra/pull/2448, the `requires_qe_access` decorator might end up being deprecated or with its functionaly removed. In this PR, the decorator and some auxiliary function is moved to provider-land.

### DIFF
--- a/test/decorators.py
+++ b/test/decorators.py
@@ -146,4 +146,4 @@ def _get_credentials():
         # Use the first available credentials.
         return list(discovered_credentials.values())[0]
 
-    raise Exception('Could not locate valid credentials.')
+    raise Exception('Could not locate valid credentials.') from None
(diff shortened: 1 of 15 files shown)


test/decorators.py
```diff
@@ -73,3 +78,72 @@ def _wrapper(self, *args, **kwargs):
...
+        # will always be in place, as is used by the Travis setup.
+        return Credentials(os.getenv('IBMQ_TOKEN'), os.getenv('IBMQ_URL'))
+
+    # Attempt to read the standard credentials.
+    discovered_credentials = discover_credentials()
+
+    if discovered_credentials:
+        # Decide which credentials to use for testing.
+        if len(discovered_credentials) > 1:
+            try:
```
delapuente: Non blocking, consider using `contextlib.suppress` or a presence test.

diego-plan9:
Thanks @delapuente ! I'll aim for a cautious "if the ported code was working, try not to touch it too much" approach and we can handle the comments as follow-ups.

Issue #105, opened 2019-05-14
Make the websocket mocked server during the tests more dynamic


diego-plan9:
### What is the expected behavior?

In #100, a mock server for testing websockets was introduced, that currently uses a `host` and `port` hardcoded to specific values. As mentioned during the review, it would be a good idea to make it a bit more flexible - for example, picking up a random available port, or moving the variables to class-level for being able to use it in more advanced ways.

See https://github.com/Qiskit/qiskit-ibmq-provider/pull/100#discussion_r283760340, https://github.com/Qiskit/qiskit-ibmq-provider/pull/100#discussion_r283759427

diego-plan9:
Closed via #131

PR #131, opened 2019-05-30
Make the websocket mocked server during the tests more dynamic

Max-Reuter:
### Summary
Define `IP` and `port`s used by the `Websocket` tests at the class level. See #105.

### DIFF
--- a/test/ibmq/websocket/test_websocket.py
+++ b/test/ibmq/websocket/test_websocket.py
@@ -27,7 +27,8 @@ from qiskit.test import QiskitTestCase
 
 from .websocket_server import (
     TOKEN_JOB_COMPLETED, TOKEN_JOB_TRANSITION, TOKEN_WRONG_FORMAT,
-    TOKEN_TIMEOUT, websocket_handler)
+    TOKEN_TIMEOUT, TEST_IP_ADDRESS, TEST_WSS_PORT, TEST_WS_PORT,
+    websocket_handler)
 
 
 class TestWebsocketClient(QiskitTestCase):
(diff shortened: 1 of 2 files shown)


test/ibmq/websocket/test_websocket.py
```diff
@@ -29,13 +29,17 @@
     TOKEN_JOB_COMPLETED, TOKEN_JOB_TRANSITION, TOKEN_WRONG_FORMAT,
     TOKEN_TIMEOUT, websocket_handler)
 
+TEST_IP_ADDRESS = '127.0.0.1'
+TEST_WSS_PORT = '9876'
```
diego-plan9: Can you rename this variable? The difference is not that it is `wss` vs `ws`, it is that it signals an invalid port that we use for the `test_invalid_url` to fail.

test/ibmq/websocket/test_websocket.py
```diff
@@ -44,13 +48,12 @@ def test_invalid_url(self):
...
 class TestWebsocketClientMock(QiskitTestCase):
     """Tests for the the websocket client against a mock server."""
-
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
 
         # Launch the mock server.
-        start_server = websockets.serve(websocket_handler, '127.0.0.1', 8765)
+        start_server = websockets.serve(websocket_handler, TEST_IP_ADDRESS, int(TEST_WS_PORT))
```
diego-plan9: Can you make `TEST_WS_PORT` an integer directly when defined?

PR #132, opened 2019-05-31
Use nest_asyncio for allowing controlling the event loop in Jupyter

### LINKED ISSUE #120: Revise WebSockets interaction with Jupiter Notebooks

Our current implementation of WebSockets is based on `asyncio`, which causes some issues with Jupyter Notebooks since the control of the event loop is directly managed by Jupyter (see https://github.com/jupyter/notebook/issues/3397).

We should revise the implementation deciding between:
* not attempting to start the event loop manually (probably the preferred solution), and instead try to add the task if it is already running
* use another package that allows nesting event loops (might have some implications in the complexity and side effects)
* as a last resort, verify that Exceptions raised from the WebSockets module are handled gracefully, triggering the http-only pooling as a fallback

diego-plan9:
### Summary

Closes #120

... via the second approach: 

> use another package that allows nesting event loops (might have some implications in the complexity and side effects)

As hinted on the [related Jupyter issue](https://github.com/jupyter/notebook/issues/3397#issuecomment-419474214), `nest_asyncio` can be a viable alternative for being able to manage the event loop directly (allowing nested loops via patching). In theory, recent Jupyter versions would not require this patch, but for preserving compatibility it seems a good compromise. We should however explore cleaner alternatives and revise the websockets usage in upcoming PRs after the release, but hopefully this will avoid fixing the bug for the time being - I'll expand after merging on the issues.

### DIFF
--- a/qiskit/providers/ibmq/api_v2/websocket.py
+++ b/qiskit/providers/ibmq/api_v2/websocket.py
@@ -20,6 +20,7 @@ import logging
 import time
 from concurrent import futures
 
+import nest_asyncio
 from websockets import connect, ConnectionClosed
 
 from qiskit.providers.ibmq.apiconstants import ApiJobStatus, API_JOB_FINAL_STATES
@@ -29,6 +30,9 @@ from .exceptions import (WebsocketError, WebsocketTimeoutError,
 
 
 logger = logging.getLogger(__name__)
+
+# Patch asyncio to allow nested use of `loop.run_until_complete()`.
+nest_asyncio.apply()
 
 
 class WebsocketMessage:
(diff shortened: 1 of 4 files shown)


qiskit/providers/ibmq/api_v2/websocket.py
```diff
@@ -30,6 +31,9 @@
 
 logger = logging.getLogger(__name__)
 
+# Patch asyncio to allow nested use of `loop.run_until_complete()`.
```
delapuente: Before merging, can we add a comment at the point we use `run_until_complete` saying that without this dependency, it would not work under a Jupyter environment?

PR #137, opened 2019-06-03
Increase Travis timeout

delapuente:
Set to 120 minutes. I'll adjust if this is too much after seeing one complete suite finishing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ stage_generic: &stage_generic
     - pip install -U -r requirements-dev.txt
   script:
     # Run the tests.
-    - make test
+    - travis_wait 120 make test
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -38,7 +38,7 @@ stage_generic: &stage_generic
     - pip install -U -r requirements-dev.txt
   script:
     # Run the tests.
-    - make test
+    - travis_wait 120 make test
```
diego-plan9: Can you move it only to the slow tests job? Mostly for guarding against longer times on the rest of the jobs, that are invoked more often: in that case it is probably more handy to have the job timeout and then re-launch manually if needed rather than having those spots in the queue being held up for 2 hours.

Issue #116, opened 2019-05-22
job.properties fails validation if job not successful 


nonhermitian:
### Informations

Moved from: https://github.com/Qiskit/qiskit-terra/issues/2408

- **Qiskit version**: master
- **Python version**:
- **Operating system**:

### What is the current behavior?
```python
q = QuantumRegister(2)
c = ClassicalRegister(2)
qc = QuantumCircuit(q, c)

qc.h(q[0])
qc.cx(q[0], q[1])
qc.measure(q, c)

backend = IBMQ.get_backend('ibmqx4')
job = execute(qc, backend)
job.properties()
```
gives 
```python

ModelValidationError: {'gates': {0: {'qubits': ['Missing data for required field.'], 'parameters': {0: {'unit': ['Missing data for required field.'], 'date': ['Missing data for required field.'], 'name': ['Missing data for required field.'], 'value': ['Missing data for required field.']}}, 'gate': ['Missing data for required field.']}}, 'general': {0: {'unit': ['Missing data for required field.'], 'date': ['Missing data for required field.'], 'name': ['Missing data for required field.'], 'value': ['Missing data for required field.']}}, 'backend_name': ['Missing data for required field.'], 'backend_version': ['Missing data for required field.'], 'last_update_date': ['Missing data for required field.'], 'qubits': ['Missing data for required field.']}
```

if the job is not yet finished.

### What is the expected behavior?
I would expect the properties to be available as nothing in the properties object actually requires the job having been successful.  Indeed, it is nothing but the backend properties.

diego-plan9:
> I would expect the properties to be available as nothing in the properties object actually requires the job having been successful. Indeed, it is nothing but the backend properties.

From the API side, it seems the properties are being fetched at the time the job is completed - mostly as an attempt to ensure that they are as up to date as possible (taking as a reference point the "time the job was actually executed").

There are also other corner cases that might prevent the properties being available at all times - for example, jobs that error out early before reaching the devices. We might be able to push the availability of `properties()` a bit earlier in some cases, but that is a question probably better suited to the API team - can you go through the cases and the desired behaviour with them?

On the provider side, a reasonable compromise might be to clarify in the documentation that the `properties()` of a job are not available in all cases (ideally with a bit more of info about which cases), and revise the function to handle empty `properties()` gracefully - which indeed is not happening at the moment.

nonhermitian:
[...]

PR #139, opened 2019-06-05
Fix for empty job properties

jyu00:
### Summary
This is a fix for #116. Job properties may not be available if the job hasn't completed. Rather than throwing an exception, `job.properties` will return `None` with this fix.


qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -190,11 +190,17 @@ def properties(self):
...
-            BackendProperties: the backend properties used for this job.
+            BackendProperties: the backend properties used for this job, or None if
+                properties are not available.
         """
         self._wait_for_submission()
 
         properties = self._api.job_properties(job_id=self.job_id())
+
+        # Backend properties of a job might not be available if the job hasn't
+        # completed. This is to ensure the properties returned are up to date.
```
diego-plan9: Can you find a way to include this or a similar clarification in the actual docstring? In this particular issue, it feels important for the users to be aware that a job's properties will not be available in all cases (as summarized in the issue), and the docstrings are arguably the best way to notify them (as opposed to inline comments in the code, with are a great practice - but more oriented towards developers and future maintainers).

test/ibmq/test_ibmq_job.py
```diff
@@ -407,6 +407,16 @@ def test_error_message_qasm(self, qe_token, qe_url):
         message = job_sim.error_message()
         self.assertIn('Job resulted in the following QASM status(es): ', message)
 
+    @requires_qe_access
+    def test_running_job_properties(self, qe_token, qe_url):
+        """Test fetching properties of a running job."""
+        IBMQ.enable_account(qe_token, qe_url)
+        backend = least_busy(IBMQ.backends(simulator=False))
```
diego-plan9: By convention, when a test performs an execution on a real device, we mark the test as `@slow_test` - mostly as the test is likely to spend a substantial amount of time in the queue.

Can you add the decorator accordingly? It would make the test be skipped by default (both locally and on the CIs), but you can still trigger it manually by using the `QISKIT_TESTS=run_slow` environment variable (it is recommended to use with caution or be prepared to wait, as running all the slow tests might take a while :grin: )

Issue #64, opened 2019-04-12
Add object storage support


diego-plan9:
### What is the expected behavior?

Add support for using object storage when submitting a job.

diego-plan9:
The basis has been implemented at #110, but for closing the issue still needs:

> At the moment the functionality is not directly reachable (other than during the test) - as a preliminary measure, a new IBMQStorageJob class is introduced that contains the changes needed for making use of object storage. A follow up is needed to complete the approach, as there are likely cases where object storage cannot be used (proxified environments, etc) and we will need to resort to "try using object storage and fallback to regular job" instead, which requires a bit more tweaking of the jobs (slightly related to #104 ).

diego-plan9:
After #144, "fallback" behaviour was added - for the new API, an attempt will be made to submit the job via object storage, and if not possible, it will revert to submitting the job via regular `POST`.

Labelled as `API action` as the support for object storage in production is not fully rolled out, in order to coordinate and test changes accordingly.

diego-plan9:
After #215, object storage is enabled by default. Support for the devices depends on the API and will be enabled gradually over the backends - the provider relies on an error during job submission in order to decide whether submitting a job via object storage, or falling back to the behavior.

PR #144, opened 2019-06-06
Add object storage fallback option

diego-plan9:
### Summary

Follow-up to #110, related to #64.

This PR moves the decision of using object storage (and small tweaks to websocket) to `IBMQJob`, instead of using a new class. The main motivation is being able to gracefully fall back to _not_ using object storage if it is not supported - at the moment it needs to be done dynamically.


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -62,8 +63,15 @@ def run(self, qobj):
         Returns:
             IBMQJob: an instance derived from BaseJob
         """
-        job = IBMQJob(self, None, self._api, qobj=qobj)
+        kwargs = {}
```
delapuente: Non-blocking. I found passing parameters in a dict hard to read, so I'm proposing this to you, although it is a matter of personal taste:
```python
is_new_api = isinstance(self._api, IBMQClient)
job = IBMQJob(
    self, None, self._api, qobj=qobj, use_object_storage=is_new_api, use_websockets=is_new_api)
```

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -226,8 +243,16 @@ def retrieve_job(self, job_id):
                                        .format(job_id, self.name()))
 
             # Check for pre-qobj jobs.
+            kwargs = {}
             try:
-                ApiJobKind(job_info.get('kind', None))
+                job_kind = ApiJobKind(job_info.get('kind', None))
+
+                if isinstance(self._api, IBMQClient):
```
delapuente: Non-blocking. Again, I would change this to:
```python
is_new_api = isinstance(self._api, IBMQClient)
requires_qobj_storage = job_kind is ApiJobKind.QOBJECT_STORAGE
```

And then pass both options explicitely at [line 269](https://github.com/Qiskit/qiskit-ibmq-provider/pull/144/files#diff-e2b490e8ae3a5cdb85265b098bea5665R269).

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -181,15 +189,24 @@ def jobs(self, limit=50, skip=0, status=None, db_filter=None):
                                                   filter=api_filter)
         job_list = []
         for job_info in job_info_list:
+            kwargs = {}
             try:
-                ApiJobKind(job_info.get('kind', None))
+                job_kind = ApiJobKind(job_info.get('kind', None))
+
+                if isinstance(self._api, IBMQClient):
```
delapuente: Same here.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -181,15 +189,24 @@ def jobs(self, limit=50, skip=0, status=None, db_filter=None):
                                                   filter=api_filter)
         job_list = []
         for job_info in job_info_list:
+            kwargs = {}
             try:
-                ApiJobKind(job_info.get('kind', None))
+                job_kind = ApiJobKind(job_info.get('kind', None))
```
delapuente: I would put **only** the line `job_kind = ...` inside the `try` to highlight the fact that is at that point that it can fail. If you want to keep all the code inside the `try`, I'd add a comment clarifying the line producing the `ValueError`.

diego-plan9: Fixed at a80ce4c

qiskit/providers/ibmq/ibmqbackend.py
diego-plan9: Yes, both are sane options - in this case, I went with the kwargs option because I think it signals better "using object storage and/or websockets are a special case: if they don't apply, rely on the default `Job` behavior", which reflects better the present situation.

delapuente: Actually, `rely on the default behaviour` is different of what I was proposing (which is set to `False` explicitely) so, fair enough.

Issue #138, opened 2019-06-04
Provide more informative exception messages


diego-plan9:
### What is the expected behavior?

As a pre-task while waiting for being able to tackle #106, one thing we can do is improve the way the exceptions are presented to the user in api v2.

Currently, they are [cast to str](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/api_v2/session.py#L140) directly for backwards compatibility - but in the process, some information is left out (for example, if the API provides specific details in the body of the response, even if it is a 403 or 404). By performing that casting a bit more intelligently (ideally taking advantage of the fact that the error message format from the API is quite structured), we should be able to provide more informative error messages during the transition period.

diego-plan9:
Fixed via #150

PR #150, opened 2019-06-07
More informative exception msg

jyu00:
### Summary
This is a fix for #138 by adding the message from the api response to the `RequestsApiError`.

### Details and comments
The response sent back by the api looks like something like 
`error': {'statusCode': 400, 'name': 'INVALID_DATE', 'message': 'Invalid date format.', 'code': 'INVALID_DATE'}`

I wanted to preserve the error message generated by `response.raise_for_status()`, so I just append the api error message to the end of that. `name` and `code` were not added to avoid making the exception message too complicated. The exception message should now look like
`400 Client Error: Bad Request for url: ... Invalid date format.`


qiskit/providers/ibmq/api_v2/session.py
```diff
@@ -138,6 +138,13 @@ def request(self, method, url, bare=False, **kwargs):
             # Wrap the requests exceptions into a IBM Q custom one, for
             # compatibility.
             message = str(ex)
+            if ex.response is not None:
+                try:
+                    message += ". {}".format(ex.response.json()['error']['message'])
```
diego-plan9: Can you try to include also `code` in the message? Even if `message` is sufficient for the user, it would be nice to also have the code to uniquely identify the error for debugging/support purposes.

test/ibmq/test_ibmq_client_v2.py
```diff
@@ -169,6 +169,21 @@ def test_get_job_includes(self, qe_token, qe_url):
         job_excluded = api.get_job(job_id, exclude_fields=['shots'])
         self.assertNotIn('shots', job_excluded)
 
+    @requires_qe_access
+    @requires_new_api_auth
+    def test_exception_message(self, qe_token, qe_url):
+        """Check exception has proper message."""
+        api = self._get_client(qe_token, qe_url)
+
+        with self.assertRaises(RequestsApiError):
```
diego-plan9: Can you revamp this block a bit? In general we try to make use of the context: You can actually make use of the "context":

```python
with self.assertRaises(X) as context_manager:
   do_something_that_raises_X

self.assertIn(context_manager.exception...., ...)
```

which helps a bit with readability, as we can avoid some nesting, indenting and re-raising.

PR #151, opened 2019-06-10
Fix CHANGELOG

diego-plan9:
### Summary

Revise the `CHANGELOG.md` for increased readability and closer adherence to  https://keepachangelog.com/en/1.0.0/ :
* use reference links instead of inline
* use simple pound-style headers instead of underline
* remove extra spacing in entries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ The format is based on [Keep a Changelog].
 ## [0.2.1] - 2019-05-06
 
 ### Fixed
-.
+
 - Improved compatibility with older setuptools versions. (\#85)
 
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,120 +2,112 @@
...
--   Improved compatibility with older setuptools versions. (\#85)
 
-[0.2.0](https://github.com/Qiskit/qiskit-ibmq-provider/compare/0.1.1...0.2.0) - 2019-05-06
-==========================================================================================
+## [0.2.1] - 2019-05-06
 
-Added
------
+### Fixed
+.
```
jyu00: I don't think this dot is meant to be here

diego-plan9: Correct, thanks for spotting it!

PR #152, opened 2019-06-10
Fix README.md links, other fixes

diego-plan9:
### Summary

Fix the links to the recently moved documents, and adjust some other minor things in the process in the `README.md`.


jyu00:
Since you're updating the README, do you want to change the `[IBM Q account]` link? The current link points to the old qx website, which doesn't allow creating new accounts anymore.

diego-plan9:
It will be changed indeed, but as part of a bigger `README` update when we make a release that officially  sets the new API as the default - I'll open an issue to keep track, thanks for double-checking!

Issue #133, opened 2019-05-31
Revise `test_get_job_includes` and endpoint


diego-plan9:
### What is the expected behavior?

The `test_get_job_includes` has been [failing](https://travis-ci.com/Qiskit/qiskit-ibmq-provider/jobs/204168475#L797) in a number of recent travis runs for the premium accounts:

```
  File "/home/travis/build/Qiskit/qiskit-ibmq-provider/qiskit/providers/ibmq/api/ibmqconnector.py", line 194, in get_job
    if ('result' in qasm) and ('data' in qasm['result']):
TypeError: argument of type 'NoneType' is not iterable
```

The endpoint has been slightly flaky for a while - we should investigate and coordinate with the API team for a proper solution, and ensure that:

* [ ] the endpoint is working for all environments + both old api and new api
* [ ] the test use the same and correct fields (for connector we use `deleted`, for client we use `shots`)
* [ ] the test also exercises `include_fields`
* [ ] the function handles unexpected cases gracefully (excluding/including non-existent fields, etc)

jyu00:
`get_job` with a filter now fails with 500 internal error. It still works if no filter is specified. The API team opened an issue on this.

jyu00:
`get_job` with `include_fields=['status']` is now failing with a generic error for v1 api. It appears to be the only field failing, and `exclude_fields['status']` works.

diego-plan9:
For tracking purposes, it seems we still [need some tweaks](https://travis-ci.com/Qiskit/qiskit-ibmq-provider/jobs/207482143) for the old API after #154.

jyu00:
v2 api is now working. v1 api will not be fixed as they're being deprecated.

diego-plan9:
Closed by #173 - `master` became green again! :tada:

diego-plan9:
Reopening after #186 and #188 and moved to the proposal project - it seems we are getting some assertion failures when using `v2` and the endpoints with h/g/p (ie. the default ones to be used after the changes):

```
>       self.assertNotIn('backend', job_excluded)
E       AssertionError: 'backend' unexpectedly found in {...}
```

@jyu00 , can you take another look and check with the API team?

diego-plan9:
Moved to `0.3.1` milestone - after #244, the tests are temporarily skipped.

jyu00:
[...]

PR #154, opened 2019-06-11
Fix get job with filter

jyu00:
### Summary
This is a partial fix for #133 . It modifies the field we use to determine the job type. v2 api part of the fix will wait for api changes.


test/decorators.py
```diff
@@ -73,7 +73,7 @@ def _wrapper(self, *args, **kwargs):
         # TODO: provide a way to check it in a more robust way.
         if 'quantum-computing.ibm.com/api' in qe_url:
             raise SkipTest(
-                'Skipping test that requires new API auth credentials')
+                'Skipping test that requires classic API auth credentials')
```
diego-plan9: Good catch! :+1:

Issue #148, opened 2019-06-07
Add verify support to api v2


diego-plan9:
### What is the expected behavior?

In a similar spirit to #97, our configuration supports a `verify` argument which [is used by `requests`](https://2.python-requests.org/en/master/user/advanced/#ssl-cert-verification) to determine whether we should bypass SSL validation.

We should add support to it in the `IBMQClient`, hopefull in a very similar way as `proxy` support is implemented - ending up in the `session`.

diego-plan9:
Fixed via #158

PR #158, opened 2019-06-12
Add `verify` arg support to `IBMQClient`

Max-Reuter:
### Summary
Add `verify` arg support to `IBMQClient` to allow the bypassing of SSL certificates errors. Addresses #148.

### DIFF
--- a/qiskit/providers/ibmq/api_v2/ibmqversionfinder.py
+++ b/qiskit/providers/ibmq/api_v2/ibmqversionfinder.py
@@ -29,7 +29,6 @@ class IBMQVersionFinder:
             verify (bool): if False, ignores SSL certificates errors.
             proxies (dict): proxies used in the connection.
         """
-        self.url = url
         self.client_version_finder = VersionFinder(
             RetrySession(url, verify=verify, proxies=proxies)
         )
--- a/qiskit/providers/ibmq/api_v2/ibmqversionfinder.py
+++ b/qiskit/providers/ibmq/api_v2/ibmqversionfinder.py
@@ -21,15 +21,18 @@ from .rest.version_finder import VersionFinder
 class IBMQVersionFinder:
     """Client for finding the API version being used."""
 
-    def __init__(self, url, proxies=None):
+    def __init__(self, url, verify=True, proxies=None):
         """IBMQVersionFinder constructor.
 
         Args:
             url (str): URL for the service.
+            verify (bool): if False, ignores SSL certificates errors.
             proxies (dict): proxies used in the connection.
         """
         self.url = url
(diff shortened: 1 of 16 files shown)


qiskit/providers/ibmq/api_v2/ibmqversionfinder.py
```diff
@@ -21,14 +21,18 @@
...
+    def __init__(self, url, verify=True, proxies=None):
         """IBMQVersionFinder constructor.
 
         Args:
             url (str): URL for the service.
+            verify (bool): if False, ignores SSL certificates errors.
             proxies (dict): proxies used in the connection.
         """
-        self.client_version_finder = VersionFinder(RetrySession(url, proxies=proxies))
+        self.url = url
```
diego-plan9: This line seems to be a result of a wrong merge - can you check?

qiskit/providers/ibmq/ibmqsingleprovider.py
```diff
@@ -90,12 +91,14 @@ def _authenticate(self, credentials):
...
                 return IBMQClient(api_token=credentials.token,
                                   auth_url=credentials.url,
-                                  proxies=proxies)
+                                  verify=verify, proxies=proxies)
             else:
                 # Prepare the config_dict for IBMQConnector.
                 config_dict = {
                     'url': credentials.url,
                 }
+                if credentials.websocket_url:
```
diego-plan9: This line seems to be a result of a wrong merge - can you check?

qiskit/providers/ibmq/ibmqsingleprovider.py
```diff
@@ -66,11 +66,12 @@ def backends(self, name=None, filters=None, **kwargs):
 
         return filter_backends(backends, filters=filters, **kwargs)
 
-    def _authenticate(self, credentials):
+    def _authenticate(self, credentials, verify=True):
```
diego-plan9: Instead of receiving `verify` as a parameter, can you use `credentials.verify`? The overall idea is that the `credentials` objects act as "containers" that hold all the information relevant for connecting to the API, being passed around in the upper layers, and being inspected and used precisely by this function (so the lower layers deal with the individual elements).

qiskit/providers/ibmq/ibmqsingleprovider.py
Max-Reuter: Should this block be removed?

diego-plan9: Only the websocket-related two lines - as they are not in `master`, and they seemed to be an unintended side effect of a previous merge on your branch.

Issue #156, opened 2019-06-12
Revise subprocess warning during tests


diego-plan9:
### What is the expected behavior?

After #119 and #140 , there is an intermittent warning during the tests, that seems related to the use of sub-processes (probably the process is not fully terminated before the completion of the test suite).

diego-plan9:
Fixed via #181

PR #159, opened 2019-06-12
Terminate mock proxy processes more thoroughly

Max-Reuter:
### Summary
Terminate the mock proxy server processes more thoroughly, avoiding intermittent warnings. Addresses #156.

### Details and comments
Though this solves the intermittent warning in `test_proxies`, there is a single warning that happens at the end of the full tests:
```
.../qiskit-ibmq-provider/lib/python3.6/site-packages/marshmallow/fields.py:194: ResourceWarning: unclosed <socket.socket fd=7, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('127.0.0.1', 52617), raddr=('127.0.0.1', 8080)>
  for validator in self.validators:


Ran 6 tests in 95.337s
```

Ran individually, the tests don't raise this warning. I tried adding some waiting time between tests to see if the socket would close, but it appears to *maybe* be something that's sustained throughout the test managing process itself (with hints from my `Activity Monitor`). It looks like manually closing the socket would be messy and hacky, and the socket is always released after the test manager ends anyway (because no warnings/errors are raised when running the tests again on the same addresses/ports). Would be happy to look into this further in another commit.

There's now a `TIMEOUT` variable that determines when to stop waiting for the processes to terminate after each test, incase they don't terminate soon enough. In that case, the current warnings would raise but the tests would behave as they do now. However, this means that the test warnings are technically not deterministic as the warnings could raise if the OS/something else is being slow and the timeout is surpassed.


test/ibmq/test_proxies.py
```diff
@@ -25,30 +26,37 @@
...
 
     def setUp(self):
-        # launch a mock server.
         super().setUp()
 
-        self.proxy_process = subprocess.Popen([
-            'pproxy', '-v', '-i', 'http://{}:{}'.format(ADDRESS, PORT)
-        ], stdout=subprocess.PIPE)
+        # launch a mock server.
+        command = ['pproxy', '-v', '-i', 'http://{}:{}'.format(ADDRESS, str(PORT))]
```
delapuente: You don't need the `str` part. The `format` method takes care of doing it.

test/ibmq/test_proxies.py
```diff
@@ -25,30 +26,37 @@
...
+        self.proxy_process = subprocess.Popen(command, stdout=subprocess.PIPE, close_fds=True)
```
delapuente: Not sure if this is more than necessary. Is this mandatory, @Max-Reuter ?

Max-Reuter:
In 08d4341, I merged the two tests for valid proxy functionality. Testing valid proxy functionality for `IBMQVersionFinder` and `IBMQClient` independently was the cause of the warnings (the unclosed sockets)., despite closing up the proxy processes as nicely as possible.

I merged the two tests and named the new test `test_valid_proxies`. It would be nice to have `IBMQVersionFinder` and `IBMQClient` tested separately, but I've done a lot of looking into facilitating that without raising warnings with no such luck. If anyone has an idea, please share!

Issue #153, opened 2019-06-10
Allow configurable header in requests


diego-plan9:
### What is the expected behavior?

As per API request, we need to be able to customize the header that we send with the requests based on an environment variable.

diego-plan9:
Fixed via #165

PR #165, opened 2019-06-13
Allow custom qx-client-application header

jyu00:
### Summary
Change the request header `X-Qx-Client-Application` from `qiskit-api-py` to either 
`ibmqprovider/version-<version>` or
`ibmqprovider/version-<version>/<additional custom header>`

 where `<additional custom header>` is specified via the environment variable `QE_CUSTOM_CLIENT_APP_HEADER`. This addresses #153 .

Also moved `custom_envs` and `no_envs` context managers to a new `contextmanagers.py` file, so they can be used by multiple test modules.

### DIFF
--- a/qiskit/providers/ibmq/api_v2/session.py
+++ b/qiskit/providers/ibmq/api_v2/session.py
@@ -14,11 +14,13 @@
 
 """Session customized for IBM Q access."""
 
+import os
 from requests import Session, RequestException
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
 
 from .exceptions import RequestsApiError
+from ..version import __version__ as ibmq_provider_version
 
 STATUS_FORCELIST = (
     500,  # Internal Server Error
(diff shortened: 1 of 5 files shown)


qiskit/providers/ibmq/api_v2/session.py
```diff
@@ -14,19 +14,22 @@
...
+from ..version import __version__ as ibmq_provider_version
 
 STATUS_FORCELIST = (
     500,  # Internal Server Error
     502,  # Bad Gateway
     503,  # Service Unavailable
     504,  # Gateway Timeout
 )
-CLIENT_APPLICATION = 'qiskit-api-py'
+CLIENT_APPLICATION = 'ibmqprovider/version-' + ibmq_provider_version
```
diego-plan9: Can you change it to `ibmqprovider/1.2.3.4`? The rationale would be that this way "parsers" can just split by `/` to get the desired fields.

test/contextmanagers.py
```diff
@@ -0,0 +1,54 @@
...
+
+"""Context managers for using with IBMQProvider unit tests."""
+
+import os
+from contextlib import contextmanager
+
+
+@contextmanager
+def custom_envs(new_environ):
+    """Context manager that modifies environment variables."""
```
diego-plan9: Even if not present in the original file, can you add a docstring for the `new_environ` parameter, while we are at it?

jyu00:
Fixed

Issue #160, opened 2019-06-13
Double login and exception handling in IBMQClient


diego-plan9:
### What is the expected behavior?

When using the new API, the login endpoint ends up being called twice, probably related to the changes in #111. While fixing it, we should also double-check `_request_access_token` overall, as the exceptions other than license errors seem to be ignored rather than re-raised.

diego-plan9:
Fixed by #166

PR #166, opened 2019-06-13
Login only once, address exceptions

Max-Reuter:
### Summary
Addresses #160 by logging in only once, or throwing errors if unsuccessful. No longer passes when exceptions are raised.

### DIFF
--- a/qiskit/providers/ibmq/api_v2/ibmqclient.py
+++ b/qiskit/providers/ibmq/api_v2/ibmqclient.py
@@ -80,11 +80,15 @@ class IBMQClient:
         except RequestsApiError as ex:
             response = ex.original_exception.response
             if response is not None and response.status_code == 401:
-                error_code = response.json()['error']['name']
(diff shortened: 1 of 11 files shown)


qiskit/providers/ibmq/api_v2/ibmqclient.py
```diff
@@ -69,25 +69,27 @@ def _request_access_token(self):
...
                         message = response.json()['error']['message']
                         raise AuthenticationLicenseError(message)
-                except (ValueError, KeyError):
+                except KeyError as invalid_key:
                     # the response did not contain the expected json.
-                    pass
-
-        response = self.client_auth.login(self.api_token)
-        return response['id']
+                    raise KeyError('Invalid key access attempt in the JSON response '
```
diego-plan9: This is actually the exception that we should _not_ be raising - it only happens during the specific block that attempts to check for the license error.

PR #171, opened 2019-06-17
IBMQJob publishes its creation along with its instance.

nonhermitian:
### Summary
`IBMQJob` uses the Pub/Sub interface in Terra to announce its creation, and return its instance to any subscriber.  This allows for automated job status checking functionality.

### Details and comments
This will be used to automatically monitor all jobs created in a given Python session.


nonhermitian:
It seems the AppVeyor build is setup incorrectly giving an error:
```
The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command.
```

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -413,6 +414,7 @@ def submit(self):
         if self._future is not None or self._job_id is not None:
             raise JobError("We have already submitted the job!")
         self._future = self._executor.submit(self._submit_callback)
+        Publisher().publish("terra.job.started", self)
```
diego-plan9: Can you revise the name of the event? It seems we (slightly) favour the present tense (`.start`), and depending on the scope of the changes, it might be interesting to include something that explicitly mentions the provider in the name.

diego-plan9: Also, can you add an `Events:` section to the docstring?

nonhermitian: Ok, something like `ibmq.job.start` work for you?

diego-plan9: If it works for you, yeah, sounds good!

nonhermitian:
This is now good to go.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -404,6 +405,10 @@ def job_id(self):
     def submit(self):
         """Submit job to IBM-Q.
 
+        Events:
+            Publishes job creation using `ibmq.job.start`
```
diego-plan9: Can you change this to be more in line with the other places where we document emitted events?

```
Events:
     name.of.the.event: Explanation for the event
```

nonhermitian: Ok, had no idea we actually had onther `Events` docs

diego-plan9: Not many indeed, but hopefully we can make use of the pub/sub mechanism in other places eventually - thanks for adjusting!

PR #173, opened 2019-06-18
Add v2 api tests for get job with filters

jyu00:
### Summary
Adds tests for `get_job` with filters for v2 api and disables the same tests for v1 api (since they are being deprecated and thus won't be fixed). This addresses #133.

### DIFF
--- a/test/ibmq/test_ibmq_client_v2.py
+++ b/test/ibmq/test_ibmq_client_v2.py
@@ -15,7 +15,6 @@
 """Tests for the IBMQClient for API v2."""
 
 import re
-from unittest import skip
 
 from qiskit.circuit import ClassicalRegister, QuantumCircuit, QuantumRegister
 from qiskit.compiler import assemble, transpile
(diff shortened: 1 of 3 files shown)


test/ibmq/test_ibmq_client_v2.py
```diff
@@ -149,26 +148,56 @@ def test_api_version(self, qe_token, qe_url):
...
-        api = backend._api
-        job = api.submit_job(qobj.to_dict(), backend_name)
+        # Get the job, including a field.
+        self.assertIn('backend', job)
+        self.assertIn('shots', job)
+        job_included = api.get_job(job_id, include_fields=['backend'])
+
+        # Ensure the result has only the included field
+        self.assertIn('backend', job_included)
+        self.assertNotIn('shots', job_included)
```
diego-plan9: Can you make these assertions more strict, for clarity - something like "the only key present in the dict is `backend`)?

test/ibmq/test_ibmq_client_v2.py
```diff
@@ -149,26 +148,56 @@ def test_api_version(self, qe_token, qe_url):
...
+        job_excluded = api.get_job(job_id, exclude_fields=['backend'])
+        # Ensure the result only excludes the specified field
+        self.assertNotIn('backend', job_excluded)
         self.assertIn('shots', job)
-        job_excluded = api.get_job(job_id, exclude_fields=['shots'])
-        self.assertNotIn('shots', job_excluded)
+
+    @requires_qe_access
+    @requires_new_api_auth
+    def test_get_job_with_invalid_filter(self, qe_token, qe_url):
```
diego-plan9: Can you revise this test's name and docstring? "Invalid" might be a bit ambiguous, as the filter should end up being well-formed but containing non-existent fields.

test/ibmq/test_ibmq_client_v2.py
```diff
@@ -149,26 +148,56 @@ def test_api_version(self, qe_token, qe_url):
...
+        """Check get_job with an invalid filter."""
+        IBMQ.enable_account(qe_token, qe_url)
+
+        api, job = self._submit_job_to_backend('ibmq_qasm_simulator')
+        job_id = job['id']
+
+        # Get the job, excluding an non-existent field.
+        self.assertNotIn('dummy_exclude', job)
+        self.assertNotIn('dummy_include', job)
+        filtered_job = api.get_job(job_id, exclude_fields=['dummy_exclude'],
```
diego-plan9: Can you split the handling in two - one for "non-existent excluded field" and other for "non-existent included field" somehow (via different blocks of call+asserts, different tests, etc)? They seem to be different enough to merit testing each of the two separately.

test/ibmq/test_ibmq_connector.py
```diff
@@ -145,6 +146,7 @@ def test_qx_api_version(self, qe_token, qe_url):
         version = api.api_version()
         self.assertIn('new_api', version)
 
+    @skip('function is being deprecated')
```
diego-plan9: Can you remove them directly instead of skipping (and maybe also `test_get_job_excludes` and the auxiliary function)? We'll consider the feature fully deprecated for the api v1 :tada:

Issue #161, opened 2019-06-13
Add NTML proxy support to api_v2


diego-plan9:
### What is the expected behavior?

In the same spirit of previous issues, we should add _NTLM_ proxy support for the api v2.

Max-Reuter:
See #97 for reference.

diego-plan9:
Fixed via #174

PR #174, opened 2019-06-18
Enable NTLM proxy support for new api

Max-Reuter:
### Summary
Enable NTLM proxy support for `api_v2`. See #161.


qiskit/providers/ibmq/ibmqsingleprovider.py
```diff
@@ -73,30 +74,34 @@ def _authenticate(self, credentials):
...
         Raises:
             ConnectionError: if the authentication resulted in error.
         """
         # Use an IBMQVersionFinder for finding out the API version.
         proxies = credentials.proxies.get('urls')
+        ntlm_credentials = credentials.proxies.get('ntlm_credentials')
+        auth = None if not ntlm_credentials else HttpNtlmAuth(
+            ntlm_credentials['username'],
+            ntlm_credentials['password']
+        )
```
diego-plan9: Can you revise this block? `credentials.proxies.get('ntlm_credentials')` will always return `False` - you can get a glimpse of how the original configuration is specified [here](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/api/utils.py#L171), and trace accordingly (probably using the old api as a reference would help).

We should also take into consideration some subtleties to avoiding `_authenticate()` failing ungracefully (that _both_ username and password are present, etc).

Issue #99, opened 2019-05-11
Add support for regenerating access_token to api_v2


diego-plan9:
### What is the expected behavior?

In the classic API, there was some support for transparently regenerating an access token if the old one expired, based on intercepting `401` errors, calling the endpoint for login again, and retrieving the call using the new access token.

It would be good to reincorporate the behaviour into `api_v2` - the basis are in place:
* the login for the initial setting up is `IBMQClient._init_api_client()`
* the `Session`s support modify the access token directly as an attribute.
* the Exception policy has been unified (all client methods are thin calls to a request, which hopefully means the handling can be performed at that level).

The priority of this addition should be consulted with the API team - if the access token are not likely to expire in a short amount of time, it might not be a frequent enough case to merit re-introducing.

Max-Reuter:
Spoke with the API team - the priority is high. The access token expires every 2 weeks. However, as above, when a `401` error is hit in `api_v2`, the provider doesn't generate another access token. Working on this now.

PR #175, opened 2019-06-19
Regenerate user access tokens upon 401 error

Max-Reuter:
### Summary
Regenerate user's access token if a 401 error is received. If it fails again regardless, raise an error. See #99.


diego-plan9:
After talking with the API team, closing in the hopes of a fresh start for this particular issue - ideally, we should be able to trigger the regenerating at a deeper level (ie. in the session), and not require explicitly adding the lines to all the calls of a `Client` , for maintainability purposes, as part of the goal of the issue is to make that renewal transparent. Combined with the recent changes for the authentication proposal, re-starting it sounds like a good choice.

PR #181, opened 2019-06-21
Address subprocess warning during tests

Max-Reuter:
### Summary
Add a destructor to `RetrySession` that closes the session to avoid subprocess warnings. Add more subprocess cleanup to `test_proxies.py`. Addresses #156.


diego-plan9:
Thanks @Max-Reuter , seems like a nifty fix - for reference, I'm relying on the review at #159 for the warnings related to the subprocess.

Issue #93, opened 2019-05-09
Increase coverage of IBMQClient tests


diego-plan9:
### What is the expected behavior?

During #78 , the `test_ibmq_client_v2.py` was added in order to provide _minimal_ testing for the new api, heavily inspired by the classic `test_ibmq_connector.py`.

We should ensure that the tests for `IBMQClient` are more comprehensive, as it is a key component for the provider and will become more relevant after the API 2 moves out of beta. In particular:
* [ ] ensure that all public methods are covered by a test (potentially more for complex methods)
* [ ] ensure that there are no skipped tests (might need action on the API side)
* [ ] ensure the asserts are meaningful and relatively strong

Lower priority:
* [ ] revise `_get_client` to see if we can avoid some double instantiation `_ = get_client()`  (see https://github.com/Qiskit/qiskit-ibmq-provider/pull/78#discussion_r280908472)
* [ ] check if it makes sense to avoid compiling a circuit (loading the qobj directly from qasm or similar), or move it to an auxiliary function, for clarity and not coupling the test to the compilation

In the process, some oddities might be found - It might be cleaner to rewrite the tests from scratch instead of basing them on the previous ones.

PR #188, opened 2019-06-24
Increase coverage of `IBMQClient` tests

Max-Reuter:
### Summary
[WIP] Addresses #93.

### DIFF
--- a/test/ibmq/test_ibmq_client_v2.py
+++ b/test/ibmq/test_ibmq_client_v2.py
@@ -74,7 +74,7 @@ class TestAccountClient(QiskitTestCase):
 
         # Run the job through the IBMQClient directly.
         api = backend._api
-        job = api.job_submit(qobj.to_dict(), backend_name)
+        job = api.job_submit(backend_name, qobj.to_dict())
 
         self.assertIn('status', job)
         self.assertIsNotNone(job['status'])
@@ -216,7 +216,7 @@ class TestAccountClientJobs(QiskitTestCase):
         self.assertIn('backend', self.job)
         self.assertIn('shots', self.job)
         job_included = self.client.job_get(self.job_id,
-                                           include_fields=['backend', 'shots'])
+                                           included_fields=['backend', 'shots'])
 
         # Ensure the result has only the included fields
         self.assertEqual({'backend', 'shots'}, set(job_included.keys()))
@@ -226,7 +226,7 @@ class TestAccountClientJobs(QiskitTestCase):
         # Get the job, excluding a field.
         self.assertIn('shots', self.job)
         self.assertIn('backend', self.job)
-        job_excluded = self.client.job_get(self.job_id, exclude_fields=['backend'])
+        job_excluded = self.client.job_get(self.job_id, excluded_fields=['backend'])
 
         # Ensure the result only excludes the specified field
         self.assertNotIn('backend', job_excluded)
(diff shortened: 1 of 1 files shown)


test/ibmq/test_ibmq_client_v2.py
```diff
@@ -298,3 +231,163 @@ def test_url_unreachable(self, qe_token, qe_url):
...
+
+    def test_job_final_status_websocket(self):
+        """Test getting a job's final status via websocket."""
+        result = self.client.job_final_status_websocket(self.job_id)
+        status_present = 'status' in result
+        self.assertTrue(status_present)
+
+    def test_job_properties(self):
+        """Test getting job properties."""
+        # TODO - is {} an acceptable response here?
```
diego-plan9: Yes - `properties` might not be available until the job has finished. Actually, can you get the job to finish inside this test, and after that do the relevant assertions (you can use a field that is required for the `BackendProperties` - for example, `backend_name`, probably).

test/ibmq/test_ibmq_client_v2.py
```diff
@@ -298,3 +231,163 @@ def test_url_unreachable(self, qe_token, qe_url):
...
+        result = self.client.job_properties(self.job_id)
+        self.assertIsNotNone(result)
+
+    # TODO: RequestsApiError: 404 Client Error: Not Found for url: [...]
+    def test_job_cancel(self):
```
diego-plan9: Can you remove this test? It is also one of those that require slightly trick logic (in this case, you need the job to be in a "cancellable" state) and better of included during #189

test/ibmq/test_ibmq_client_v2.py
```diff
@@ -298,3 +231,163 @@ def test_url_unreachable(self, qe_token, qe_url):
...
+        # TODO
+        pass
+
+    def test_job_get_responsive(self):
+        """Test getting a job responds with something other than None."""
+        result = self.client.job_get(self.job_id)
+        status_present = ['status' in [dict_ for dict_ in result.get('qasms')]]
+        self.assertTrue(status_present)
+
+    def test_job_get_filtered_fields(self):
```
diego-plan9: Can you check if this tests is not already covered by the moved `test_get_job_includes_xxx/test_get_job_excludes_xxx` tests a few lines above? Note that `client.get_job()` is actually an alias for `client.job_get()` for compatibility purposes (actually, we should rename the mentioned tests to `test_job_get_includes_....` for consistency).

test/ibmq/test_ibmq_client_v2.py
```diff
@@ -298,3 +231,163 @@ def test_url_unreachable(self, qe_token, qe_url):
...
+    @skip
+    # TODO: RequestsApiError: 404 Client Error: Not Found for url: [...] None, Error code: NotFound.
+    def test_job_result_object_storage(self):
+        """Test getting a job's result using object storage."""
+        # result = self.client.job_result_object_storage(self.job_id)
+
+        # TODO
+        pass
+
+    def test_job_get_responsive(self):
```
diego-plan9: Can you rename this one to `test_job_get`, and also relax the assert?

test/ibmq/test_ibmq_client_v2.py
Max-Reuter: Those tests test `get_job_xxx`, not `job_get_xxx`. I can update them to do the latter, or if we need both tests I'll keep the similar ones currently in the PR.

PR #202, opened 2019-06-25
Fix cancel for v2

### LINKED ISSUE #189: job.cancel() raises JobError with v2 API
### Information

- **Qiskit IBMQ Provider version**: master
- **API URL or version**: version 2

- **Python version**:
- **Operating system**:

`job.cancel()` raises

```
qiskit.providers.exceptions.JobError: 'Error cancelling job: 404 Client Error: Not Found for url: https://api-qcon.quantum-computing.ibm.com/api/Jobs/Nonecancel?access_token=...
```

when canceling a job right after submission using v2 API.

```
 job = backend.run(qobj)
 job.cancel()
```

job is cancelled

### Suggested solutions
Two things need to be fixed:
- URL should be `/cancel`
- `job.cancel()` should wait for job submission (i.e. a valid job id) before doing the cancel.

jyu00:
### Summary
This fixes #189 . `job.cancel()` will wait for job submission to ensure a job id is available to do the cancel. 

### Details and comments
The new test case is in `test_ibmq_client_v2` but will likely be moved/refactored per #93 .
An existing test case `test_cancel_while_initializing_fails` was deleted because it looks for failure to cancel when a job is in initializing state. However, I think `job.cancel()` should wait for a job id to be available then do the cancel; otherwise the user would need to know to poll for the job status to become non initializing before calling the cancel.

### DIFF
--- a/test/ibmq/test_ibmq_client_v2.py
+++ b/test/ibmq/test_ibmq_client_v2.py
@@ -18,6 +18,7 @@ import re
 
 from qiskit.circuit import ClassicalRegister, QuantumCircuit, QuantumRegister
 from qiskit.compiler import assemble, transpile
+from qiskit.providers import JobStatus
 from qiskit.providers.ibmq import IBMQ
 from qiskit.providers.ibmq.api_v2 import IBMQClient
 from qiskit.providers.ibmq.api_v2.exceptions import ApiError, RequestsApiError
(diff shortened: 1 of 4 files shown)


test/ibmq/test_ibmq_client_v2.py
```diff
@@ -28,7 +29,7 @@
 
 
 class TestIBMQClient(QiskitTestCase):
-    """Tests for IBMQConnector."""
+    """Tests for IBMQClient."""
```
diego-plan9: :+1:

diego-plan9:
Note that in the latest merge of #186, I dropped the `TestIBMQClient.test_job_cancel()` test - as it was performing the test not at the `client` level, but at the `Job` level, which should be somewhat covered at `TestIBMQJob.test_cancel()`. Double-checking that the existing tests covers your changes and issuing a fix if not would be advisable!

PR #203, opened 2019-06-25
Set up CI with Azure Pipelines to replace appveyor

mtreinish:
This commit adds a yaml config to start using azure pipelines instead of using appveyor for our windows testing. For our purposes it's mostly the same except that azure is faster and provides much more capacity.


mtreinish:
So I'm a bit confused by that actually? Why would it try to run the online tests by default? The jobs do not configure any env variables or pass anything extra into the environment, so why don't they skip?

mtreinish:
I can actually reproduce this failure locally, with a fresh checkout and no credentials in my environment all the tests fail.

mtreinish:
Ok, I fixed the tests to properly skip when there are no credentials found instead of hard fail like before. That fixed the job so all the online tests are skipped (and also fixes anyone trying to run tests without any credentials configured)

mtreinish:
Quite frankly that's a broken convention, you do a fresh clone, run `make test` and get 77 test failures. How do you expect new contributors to know that they need to either setup credentials or set `QISKIT_TESTS=skip_online` or `QISKIT_TESTS=mock_online`. Or that those 77 failures aren't caused by their local changes. Tests should always work off of a clean environment so there is a common baseline to make changes against. But, sure that's a separate discussion we can continue elsewhere.

I'll revert that and set the env var. What I'm not seeing is how appveyor was configured to skip online tests before. It's not in it's yaml file and not in the web ui from what I could see from a quick poke around the settings. Where was the skip being set over there?

diego-plan9:
> I'll revert that and set the env var.

Thanks!

Can you double-check the comment [in the first review](https://github.com/Qiskit/qiskit-ibmq-provider/pull/203#pullrequestreview-254440285), though? At a glance and being no expert on the azure configuration, wouldn't the current configuration would not fulfill the bolded part and apply the skipping also to master?

> This seems to be trying to run the online tests during PR runs - can you adjust the configuration so the behaviour stays consistent with the rest of the CIs (**no online tests on PRs, only when merging to master**)?

mtreinish:
I do not have the credentials that you used to set up appveyor. Until those are added to the azure pipelines webui as a secret we will not be able to do that. You can add the credentials (I sent you an invite to be an admin on the ibmq provider project in azure pipelines yesterday as I pushed this PR) and update the logic after you do.

Issue #101, opened 2019-05-13
Cleanup websocket TODOs and details


diego-plan9:
### What is the expected behavior?

After #100, there are some items that need confirming with the API team, and decide on the best course of action:
* [ ] check whether the connection is _always_ closed when a job reaches a final state. In #100, we explicitly close the connection, but might be unneeded. https://github.com/Qiskit/qiskit-ibmq-provider/pull/100#discussion_r283289233
* [ ] check whether a timeout during websockets (as in, the last response received from websocket did not contain a final state) can be reliable mapped to "the job is still in a final state". https://github.com/Qiskit/qiskit-ibmq-provider/pull/100#discussion_r283289148
* [ ] ~check the protocol for authenticating: as of #100, we rely on getting a specific response after trying to authenticate, and on the closing of the connection for determining that the token was invalid. It might be useful if we get an explicit reply when the token is invalid. https://github.com/Qiskit/qiskit-ibmq-provider/pull/100#discussion_r283752557~
* [ ] check that unabilty to connect to the socket server (for example, due to using proxies) falls back in all cases to non-websocket status pooling

diego-plan9:
Closed via #206

PR #206, opened 2019-06-26
Cleanup websocket TODOs

jyu00:
### Summary
This addresses #101 . Main changes
- don't raise `ConnectionClosed` if job is in a final state (api will close connection)
- adjust timeout when retry with http
- create new event loop if one is not found

### DIFF
--- a/test/ibmq/websocket/test_websocket_integration.py
+++ b/test/ibmq/websocket/test_websocket_integration.py
@@ -13,12 +13,17 @@
 # that they have been altered from the originals.
 
 """Test for the Websocket client integration."""
+from unittest import mock
+from threading import Thread
+from queue import Queue
 
 from qiskit import ClassicalRegister, QuantumCircuit, QuantumRegister
 from qiskit.compiler import assemble, transpile
+from qiskit.providers import JobTimeoutError
 from qiskit.providers.ibmq import IBMQ, least_busy
-from qiskit.providers.ibmq.api_v2.exceptions import (
-    WebsocketError, WebsocketTimeoutError, WebsocketIBMQProtocolError)
+from qiskit.providers.ibmq.api_v2.websocket import WebsocketClient, WebsocketMessage
+from qiskit.providers.ibmq.job.ibmqjob import IBMQJob
+from qiskit.providers.jobstatus import JobStatus
 from qiskit.test import QiskitTestCase, slow_test
 
 from ...decorators import requires_new_api_auth, requires_qe_access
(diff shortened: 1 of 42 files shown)


test/ibmq/websocket/test_websocket_integration.py
```diff
@@ -26,45 +33,129 @@
...
+        self.qc1.measure(qr[0], cr[0])
 
-        job = backend.run(qobj)
-        # Manually disable the non-websocket pooling.
-        job._wait_for_final_status = None
-        result = job.result()
+        # Create and run a default Qobj using the simulator.
+        self.circuit = transpile(self.qc1, backend=self.sim_backend)
+        self.qobj = assemble(self.circuit, backend=self.sim_backend, shots=1)
+        self.job = self.sim_backend.run(self.qobj)
```
diego-plan9: Can you take this call out of `setUp` and have the tests that need a job against the simulator run it explicitly (in a function or with this line directly, etc)? Mostly in order to avoid running extra jobs during the tests, and in general as we tend to avoid cramming up `setUp` for readability and performance (don't worry about the rest of changes to this `setUp` for the moment).

test/ibmq/websocket/test_websocket_integration.py
```diff
@@ -26,45 +33,129 @@
...
+        backend = least_busy(self.provider.backends(simulator=False))
 
-        qc = transpile(self._qc1, backend=backend)
+        qc = transpile(self.qc1, backend=backend)
         qobj = assemble(qc, backend=backend)
 
         job = backend.run(qobj)
-        # Manually disable the non-websocket pooling.
+        # Manually disable the non-websocket polling.
         job._wait_for_final_status = None
```
diego-plan9: Can you double-check that, for the tests where you intend _not_ to rely on the fallback behaviour (ie. where you really need that the calls to websocket suceed without resorting to pooling), a line similar to this one is added? This was a bit hackish, but the end goal is being able to leave the fallback out of the equation  (better alternatives might be available).

test/ibmq/websocket/test_websocket_integration.py
```diff
@@ -26,45 +33,129 @@
...
+        """Test checking status of a job in a final state via websockets."""
+        # Cancel the job to put it in a final (cancelled) state.
+        self.job.cancel()
+        self.job._wait_for_completion()
+        self.assertIs(self.job._status, JobStatus.CANCELLED)
+
+    def test_websockets_retry_bad_url(self):
+        """Test http retry after websocket error due to an invalid URL."""
+        # Use fake websocket address.
+        self.job._api.client_ws.websocket_url = 'wss://wss.wayne-enterprises.com'
```
diego-plan9: I like the theme, but in this case can you please use `localhost` or similar, to avoid extra network calls?

Issue #194, opened 2019-06-25
Add Provider compatibility methods to IBMQFactory


diego-plan9:
### What is the expected behavior?

For backwards-compatibility, the `IBMQFactory` should still have the methods that are available in the current `IBMQProvider` (in order to perform #192 gracefully).

These methods should be added to `IBMQFactory`, retaining the signature from `IBMQProvider`. In general, we can take advantage of the fact that we are _not_ allowing mixing v1 accounts and v2 accounts at the same time, and noop/delegate into `self._v1_provider`, plus emitting different `DeprecationWarning`.

* Related to the `Provider` superclass:
  * [ ] `.get_backend()`
  * [ ] `.backends()`

  For these two, If a v2 account is active in these two methods, we can return an empty list and raise and raise a `QiskitBackendNotFoundError` with a custom message.

There are other account-handling related methods that will be tackled in a separate issue.

diego-plan9:
Closed via #208, with a slight change in specs: the two `get_backend()` and `backends()` method now make an attempt to return meaningful results.

Issue #196, opened 2019-06-25
Facilitate transitioning from v1 to v2 stored credentials


diego-plan9:
### What is the expected behavior?

With the introduction of the new proposal, we should add mechanisms for guiding the user from using API v1 to using v2 as seamlessly as possible.

One point of particular relevance is the credentials stored in disk: it should be safe to assume that, at this point in time, most users have their credentials stored in the `.qiskitrc` and rely on `IBMQ.load_accounts()` for their usage. While we can encourage users to use `load_account()` (note the lack of `s` at the end) for dealing with v2, it would be also a good idea to make the transition a bit more transparent.

An idea would to add a `IBMQ.update_account(dry_run=True)` that:
* reads the credentials currently stored
* for no url or the default QE url, replaces it with the new _auth_ url
* for qconsole urls, replaces with the new auth url and adds instructions on how to use those accounts
* the `dry_run` flag is to make the user explicitly confirm the overwriting
* overall, the function should be oriented towards the user and focus on the instructions and clarity

PR #208, opened 2019-06-26
Factory with provider methods

jyu00:
### Summary
This addresses #194 . Added the public methods of `IBMQProvider` to `IBMQFactory` for backward compatibility and a small set of tests (may be refactored/merged with `test_registration` later).

### Details and comments
A couple of things I'm on the fence about:
- instead of returning an empty list when `backends()` is called with a v2 credential, I just aggregate all the provider backends to be more backward compatible. Hopefully most users don't have that many accounts to make this a long call.
- both `IBMQProvider` and `IBMQFactory` have a `save_account()` method. The easier way is to have a new name for the `IBMQFactory` one and deprecate the old one, but that is difficult since all the method names follow a nice pattern. I ended up retaining the signature of `IBMQProvider.save_account()` but deprecating the old parameters.

### DIFF
--- a/qiskit/providers/ibmq/ibmqfactory.py
+++ b/qiskit/providers/ibmq/ibmqfactory.py
@@ -24,7 +24,7 @@ from .accountprovider import AccountProvider
 from .api_v2.clients import AuthClient, VersionClient
 from .credentials import Credentials
 from .exceptions import IBMQAccountError, IBMQApiUrlError, IBMQProviderError
-from .ibmqprovider import IBMQProvider, QE_URL
+from .ibmqprovider import IBMQProvider
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


diego-plan9:
Thanks @jyu00 ! As discussed an in order to further split the remaining pieces and avoid conflicts/blocking, I made some modifications to the PR for moving the unknowns to future PRs and issues and keep it in the spirit of #194 (which was "restricted" to the backend methods mainly for being able to split the work, actually). 

In particular:
* for the deprecated account-management v1 methods (except "save"), make them raise if we are using a v2 account directly, with no attempt of compatibility. This can be revised later and reintroduced, but the rationale would be that, if users are transitioning to an v2 account, we can assume they will transition to the "way accounts are handled in v2" as well, saving us from corner cases and further enforcing moving their accounts (while still preserving access to v1 if needed).
* for the backend related methods, it was also a bit of a pickle. I'm leaning towards indeed treating those a bit special (in the sense of preserving compatibility in a stronger sense) with the arguments of it being of a different scope (in general, the account management ones are "fire and forget" in most cases, but for the backends they tend to be used more generally). So let's keep your changes for the time being, and again we can revisit if needed.
* for `save_account`, it is indeed the most tricky - hopefully we will have a clearer view after #196, so I reverted in the meantime.

Issue #193, opened 2019-06-25
Add IBMQFactory.disable_account and others


diego-plan9:
### What is the expected behavior?

As a follow-up to #186, the method `IBMQFactory.disable_account()` needs to be implemented: the function should clear the v2 credentials and providers that are in use (if any).

diego-plan9:
After some discussion, expanded the scope of the issue to also implement:

* `save_account()`
* `stored_account()`

diego-plan9:
Fixed via #212

PR #212, opened 2019-06-28
disable, save, and stored account

jyu00:
### Summary
This addresses #193. 

### Details and comments
Since the signature for `save_account` remains the same as IBMQProvider, I think we should take v1 token and issue a warning, similar to what we do with `enable_account`.

### DIFF
--- a/qiskit/providers/ibmq/ibmqfactory.py
+++ b/qiskit/providers/ibmq/ibmqfactory.py
@@ -23,7 +23,8 @@ from qiskit.providers.exceptions import QiskitBackendNotFoundError
 from .accountprovider import AccountProvider
 from .api_v2.clients import AuthClient, VersionClient
 from .credentials import Credentials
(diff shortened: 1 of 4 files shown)


qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -73,6 +74,9 @@ def enable_account(self, token, url=QX_AUTH_URL, **kwargs):
 
         # For API 1, delegate onto the IBMQProvider.
         if not version_info['new_api']:
+            warnings.warn('IBM Experience 1 account is deprecated. '
+                          'IBMQ.update_account() for updating your stored credentials.',
```
diego-plan9: In this case, I think we don't want to mention `update_account()`, as `enable_account` is meant to be invoked directly (ie. without involving _stored_ credentials). Let's leave this out for the moment, and revise later.

PR #213, opened 2019-06-28
Add IBMQFactory.load_account()

diego-plan9:
### Summary

Add the last method to `IBMQFactory`. The method is relatively similar to `enable_account` but is kept separate for the moment, and attempts to follow the rationale for other methods (only allow the method to be used with `v2` accounts).

### Details and comments

In the process, added some clarifications to both `load_accounts` and `load_account` - in these two methods we perform the version check by querying the API instead of using URLs directly, to allow the tests and development against other valid urls (staging) to work as expected (it also explains the inclusion of more context managers in the test).

### DIFF
--- a/qiskit/providers/ibmq/ibmqfactory.py
+++ b/qiskit/providers/ibmq/ibmqfactory.py
@@ -107,20 +107,47 @@ class IBMQFactory:
                                    'Please use IBMQ.disable_accounts() to '
                                    'disable the account.')
 
-        if self._credentials is not None:
-            self._credentials = None
-            self._providers = OrderedDict()
-
-        else:
+        if not self._credentials:
             raise IBMQAccountError('No account is in use for this session.')
+
+        self._credentials = None
+        self._providers = OrderedDict()
 
     def load_account(self):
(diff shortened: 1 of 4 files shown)


qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -107,20 +113,68 @@ def disable_account(self):
...
             AccountProvider: the provider for the default open access project.
+
+        Raises:
+            IBMQAccountError: if an IBM Q Experience 1 account is already in
+                use, or no IBM Q Experience 2 accounts can be found.
         """
-        raise NotImplementedError
+        if self._credentials:
+            # For convention, emit a warning instead of raising.
+            warnings.warn('Credentials are already in use.')
```
jyu00: shouldn't you return after this instead of continuing to do all the work?

diego-plan9: Good catch! Yes, it was an attempt to preserve previous behavior - I tried to find a better place for it along with a message closer to your suggestion in b6071c2

PR #215, opened 2019-06-28
Reenable object storage

### LINKED ISSUE #64: Add object storage support

diego-plan9:
### Summary

Fixes #64 

As the API now seems to provide a way to detect if a backend does not support object storage (at submit time), reenable support in the provider.

This includes tuning the level of the warning when it is not supported to `INFO` (as in the short term, most of the backends will not support object storage, which can turn out to be quite noisy for users), and adjust a test that depends on the backend supporting it.


jyu00:
The code looks good. However, why is the object storage the default if most backends don’t support it yet? Is it just to preemptively add the support so we don't have to do another release when it do become available?

diego-plan9:
Yes - seems the enabling will be done gradually, and this way it can be used as soon as it is enabled for each device without tying it to our versions.

PR #217, opened 2019-06-28
Use the default client for CircuitsManager

### LINKED ISSUE #195: Use standard AccountClient for CircuitsManager

Currently, the `CircuitsManager` is instantiated by using a client that points to the root urls instead of the h/g/p URLs. In conversations with the API team, we agreed to move the circuits endpoints into the h/g/p convention as well, as they were the last endpoints not under that tree.

Once the endpoints go live, we should update the instantiation of the `CircuitsManager`, being able to remove the `circuit_client` created specifically for it in `AccountProvider.__init__`.

diego-plan9:
### Summary

Fix #195 

Update the `AccountProvider.circuits` attribute in order to use
a `CircuitManager` against a normal `AccountClient` instead of using
a custom `AccountClient` without h/g/p.


diego-plan9:
Marking as `on hold` until the endpoint is confirmed to be ready by the API team.

diego-plan9:
Merging as the changes will be in production soon.

PR #224, opened 2019-07-01
cache backend properties and defaults

ajavadia:
cache backend properties and defaults.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ The format is based on [Keep a Changelog].
   jobs. (\#110)
 - Added support for custom qx-client-application request header via the environment
   variable `QE_CUSTOM_CLIENT_APP_HEADER` when using the new API. (\#165)
+- A refresh flag to `backend.properties()` and `backend.defaults()`. When False,
+  cached data will be returned.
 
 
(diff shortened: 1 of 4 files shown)


diego-plan9:
Why not make the method return the existing properties if they have already been fetched, and populate them in the first call, ideally allowing the user to re-fetch them from the API if desired? Ie. something along the lines of:

```python
def properties(self, refresh=False):
    if refresh or not self._properties:
        ...do the api call and populate _properties
    return self._properties
```

Hopefully this will keep the responsibilities a bit more manageable (ie. keep the "creation" of properties inside the backend itself), and at the same time keep allowing users to refresh them if needed (sort of leaving the "cache invalidation" problem to the user :grin: ) . It probably also helps us scaling, as we could use the pattern for other attributes eventually (the one that comes to mind is `pulse_defaults`, which seems to be still being hit when assembling even non-qasm qobjs).

ajavadia:
I agree. I can change it to be like that.
I mostly mimicked what was done for `backend.configuration()`. I think `configuration`, `properties` and `defaults` are all known at Backend instance creation time, and would not change (unlike something like `status`). So I don't think there is much use for ever setting `refresh=True`. But I like keeping the constructor arguments small.

Issue #184, opened 2019-06-24
Add minimal pulse test


diego-plan9:
### What is the expected behavior?

Following up #108, we should try to include a minimal test that attempts to execute a pulse qobj on a device (ideally the minimal one).

diego-plan9:
Closed via #227

PR #227, opened 2019-07-02
Add pulse test

jyu00:
### Summary
This addresses #184 . 

### Details and comments
Does not verify the results because 
1) verification requires a backend with `meas_level=2`, severely limiting where the test can run and 
2) the job has a small number of shots to make it finish faster, but that makes it hard to have a meaningful tolerance level


jyu00:
Need to revise to dynamically find suitable backend when api endpoints become available

jyu00:
Both api and backend are fixed. This is now ready to go.

Issue #197, opened 2019-06-25
Verify remaining h/g/p endpoints for proposal


diego-plan9:
### What is the expected behavior?

During #186, there are a few errors when running the tests that seem to be related to missing endpoints in the API. The API team is working on them, and this issue is mostly for keeping track of the status:

* [ ] backend properties
* [ ] backend status
* [ ] job properties

diego-plan9:
Closed as they are already in production.

Issue #96, opened 2019-05-10
Add new API accounts to travis


diego-plan9:
### What is the expected behavior?

Currently, travis performs checks against the old API when pushing to `master`. We should also enable testing of the new API urls, in preparation for getting out of beta.

Might have some overlap with #69.

diego-plan9:
After #229, we have one job in travis that uses the new API. The plan would be to eventually reverse the situation closer to the release date - have most of the tests against the new API, keeping one or two against the old api during the transition window.

Issue #218, opened 2019-06-29
Revise job cancelation in api v2


diego-plan9:
### What is the expected behavior?

It seems `test_websockets_job_final_state()` can fail ungracefully during the cancel of the job:

```
(at IBMQJob.cancel())
....
        except ApiError as error:
            self._cancelled = False
>           raise JobError('Error cancelling job: %s' % error.usr_msg)
E           qiskit.providers.exceptions.JobError: 'Error cancelling job: 400 Client Error: Bad Request for url: https://XXX/Jobs/YYY/cancel?access_token=.... Job not cancelled., Error code: JOB_NOT_CANCELLED.'
```

It might simply a small issue with the design of that particular test (the job might enter a non-cancellable state by the time we try to cancel it), but we should double-check.

jyu00:
I spoke with the api team and they said `JOB_NOT_CANCELLED` should only happen if there's an issue with the queue. @diego-plan9 do you have more information on the job that encountered this error?

jyu00:
There's a possibility of hitting `JOB_NOT_RUNNING` though. I'll fix that part.

diego-plan9:
Moved to `0.3.1` milestone - after #244, the tests are temporarily skipped.

PR #229, opened 2019-07-02
Add new api job to travis

diego-plan9:
### Summary

Related to #96

Add a single job to travis that runs against the new API. Not marking the issue as fixed, as this is a temporary change to have at least one job against the new API in the CIs - when things get a bit more stable, all jobs will run against the new API.

Marking as WIP pending the final fix for #197, which is upcoming.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -158,12 +158,13 @@ jobs:
 
     # "new api" stage
     ###########################################################################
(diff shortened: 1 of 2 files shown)


diego-plan9:
Since the endpoints are finally in production, merging in order to test! In the process, added an `allow_failures` as there are some tests that still depend on #218, #197

PR #236, opened 2019-07-07
Typo updater

jwoehr:
### Summary
Typo in warning message in update_credentials

### Details and comments
Apparently typo cut-and-pasted to README.md as well


jwoehr:
Apologies for the sloppy construction of the branch which includes some noise. The only real commit is the last.

diego-plan9:
Thanks for catching this, @jwoehr, as it is indeed very important that the instructions are accurate in this particular piece! I took the chance to also add a space that was missing in the original lines, and we are good to go :+1:

PR #241, opened 2019-07-09
Add timeout parameter to job_id

### LINKED ISSUE #238: Add timeout argument to job.job_id()
Allow users to configure how long they would like to wait to get job id.

Now, the `job_id` method in the `IBMQJob` class called the `_wait_for_submission` without any argument, which implies the timeout is 60 seconds; however, sometimes users might not be able to get the job id within 60 seconds, e.g., no available credits or queue is closed. (https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/job/ibmqjob.py#L403)

In this case, users aware of the issue of why they can not submit the jobs but they are willing to wait since they can submit the job and work on other tasks rather than checking whether or not they can submit the jobs regularly.

Thus, to address that, I think we could add a timeout argument to `job_id` method and with default value 60.

diego-plan9:
### Summary

Fixes #238


qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -400,13 +400,19 @@ def creation_date(self):
...
         return self._creation_date
 
-    def job_id(self):
-        """Return backend determined id.
+    def job_id(self, timeout=60):
+        """Return the job ID assigned by the API.
 
-        If the Id is not set because the job is already initializing, this call
-        will block until we have an Id.
+        If the job ID is not set because the job is already initializing, this
```
jyu00: I think it's clearer to say the job is _still_ initializing, rather than _already_ initializing.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -400,13 +400,19 @@ def creation_date(self):
...
+        call will block until a job ID is available of the timeout is reached.
+
+        Args:
+            timeout (float): number of seconds to wait for the job id.
```
jyu00: super nitpicking... can you change `id` to `ID` so they're all consistent in this method?

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -400,13 +400,19 @@ def creation_date(self):
...
 
-    def job_id(self):
-        """Return backend determined id.
+    def job_id(self, timeout=60):
+        """Return the job ID assigned by the API.
 
-        If the Id is not set because the job is already initializing, this call
-        will block until we have an Id.
+        If the job ID is not set because the job is already initializing, this
+        call will block until a job ID is available of the timeout is reached.
```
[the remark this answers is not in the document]
diego-plan9: Yes, creative English strikes again!

PR #245, opened 2019-07-11
Additional client tests

jyu00:
### Summary
This is for #93 . Added additional tests for AccountClient to cover all public methods (except circuits related ones, which are covered in `test_circuits`.

### DIFF
--- a/test/ibmq/test_account_client_v2.py
+++ b/test/ibmq/test_account_client_v2.py
@@ -12,7 +12,7 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
-"""Tests for the IBMQClient for API v2."""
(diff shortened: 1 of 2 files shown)


test/ibmq/test_account_client_v2.py
```diff
@@ -178,6 +178,35 @@ def test_custom_client_app_header(self):
...
+
+        api = backend._api
+        job = backend.run(qobj)
+        job_id = job.job_id()
+
+        try:
+            api.job_cancel(job_id)
+        except RequestsApiError as ex:
+            # TODO: rewrite using assert
+            if all(err not in str(ex) for err in ['JOB_NOT_RUNNING', 'JOB_NOT_CANCELLED']):
```
diego-plan9: Added a `TODO` here in the hopes of being able to use asserts eventually when we reenable the test :grimacing:

PR #263, opened 2019-07-17
Use pagination in backend.jobs()

### LINKED ISSUE #261: Use pagination actively in backend.jobs() limit

Currently, `backend.jobs()` passes the `limit` argument to the API call directly, placing the responsibility of managing pagination in the user (via making different calls to `.jobs()` varying the `limit` and `skip` parameters). We should be able to place some "intelligence" in the method so the extra calls are done transparently, so the `limit` passed to the Python method reflects better the number of jobs the user will end up receiving.

Some finer points to consider:
* check similarities between API v1 and API v2 when dealing with `limit`
* consider allowing/disallowing a way of specifying "no limit" 
* revise handling of invalid jobs (for example, pre-qobj jobs are currently dropped by the provider, which affects the number of jobs returned finally)

diego-plan9:
### Summary

Fixes #261 

Make `backend.jobs()` deal with pagination directly when the `limit` argument is above the hard limit on the number of jobs returned by the API, resulting in several calls to the API until the number of jobs equals the desired amount. 

Additionally, revise the case of initializing jobs with errored status, to avoid doing an extra call to the API in those cases.


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -207,10 +209,32 @@ def jobs(self, limit=50, skip=0, status=None, db_filter=None):
...
+            if not job_page:
+                # Stop if there are no more jobs returned by the API.
+                break
+
+            if limit:
+                if len(job_responses) >= limit:
+                    # Stop if we have reached the limit.
+                    break
+                current_page_limit = limit - len(job_responses)
+            else:
```
jyu00: Does `limit=0` (or `None`) have special meanings (e.g. unlimited)? If it just means don't retrieve any jobs (which would make this method call useless), then shouldn't we just check up front and return without making any api calls? If it has special meanings it should be documented.

Issue #46, opened 2019-02-28
Test we are specializing the backend class for simulators


delapuente:
When using a simulator, a special class allowing to call `run` with different parameters is used. We should test we effectively instantiate this class when choosing a simulator.

luisg5:
The IBMQSimulator class contains a run method which handles the different parameters: backend_options and noise_model. 

To test whether the IBMQSimulator class is effectively instantiated, would it be necessary to call the IBMQSimulator run method with the additional parameters. Then, validate that the QObj returned by the IBMQSimulator run method has updated configs?

diego-plan9:
Welcome, @luisg5 !

In this issue, we are only interested in making sure that "the classes that create `IBMQBackend`s and `IBMQSimulator` perform their job as expected" - that is, without really worrying too much about what happens after instantiation (if the objects are of the "right" class, their methods such as `run()` and others will be covered by other tests already in place).

> To test whether the IBMQSimulator class is effectively instantiated ...

... you can just make use of assertions that, once a list of backends is obtained via the regular means, each object that represent a backend is of "the right class" - [`assertIsInstance()`](https://docs.python.org/3.5/library/unittest.html#unittest.TestCase.assertIsInstance) might be helpful in solving the issue.

diego-plan9:
Fixed via #270

PR #270, opened 2019-07-18
Test that backend simulators are instantiated correctly (#46)

luisg5:
### Summary
Fixes (#46). Added a test to check whether backends that are simulators get correctly instantiated as an IBMQSimulator class.


test/ibmq/test_ibmq_provider.py
```diff
@@ -64,6 +65,14 @@ def test_remote_backends_exist_simulator(self):
         remotes = self.provider.backends(simulator=True)
         self.assertTrue(remotes)
 
+    def test_remote_backends_instantiate_simulators(self):
+        """Test if remote backends that are simulators
+        get instantiated as an IBMQSimulator class."""
+        remotes = self.provider.backends(simulator=True)
+
+        for backend in remotes:
+            self.assertIsInstance(backend, IBMQSimulator)
```
diego-plan9: Instead of making the assertions in a loop directly, can you make use of [`subTest()`](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests)? The main practical difference is :

> Without using a subtest, execution would stop after the first failure, and the error would be less easy to diagnose because the value of i wouldn’t be displayed:

In general, it is a way of "splitting" a test into several subtests that are run "_independently_" - there should not be much difference in this particular test, but it would help in an hypothetical case where one particular backend fails to be of the required class: with subtests, we could still be able to assert over the rest of the backends, isolating that single failure.

test/ibmq/test_ibmq_provider.py
```diff
@@ -64,6 +65,14 @@ def test_remote_backends_exist_simulator(self):
         remotes = self.provider.backends(simulator=True)
         self.assertTrue(remotes)
 
+    def test_remote_backends_instantiate_simulators(self):
+        """Test if remote backends that are simulators
+        get instantiated as an IBMQSimulator class."""
```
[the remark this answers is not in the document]
luisg5: Thanks for the links @diego-plan9. I'll make sure to keep the comments short for that reason.

test/ibmq/test_ibmq_provider.py
luisg5: That's very interesting. I wasn't aware that future tests would fail in this case. This will be very helpful for future testing. Thank you.

PR #277, opened 2019-07-23
Allow datetime in backend properties

### LINKED ISSUE #272: Historical calibration data
For a `IBMQBackend` instance, `backend.properties()` will give the latest recorded properties. As far as I can tell, there is currently no way to access older properties. This would be useful to have. No preference on the particular API. Could be a new method, or maybe a `limit` kwarg to `properties()`, similar to `jobs()`. Or maybe a `start_date`/`end_date` filter to return calibration data over a particular period of time.

luisg5:
### Summary

This fixes #272


test/ibmq/test_filter_backends.py
```diff
@@ -56,15 +56,27 @@ def test_filter_least_busy(self, provider):
...
         backends = provider.backends(simulator=False)
 
         my_filter = {'last_update_date': {'lt': '2019-01-01T00:00:00.000'}}
         filter_date = datetime.fromisoformat(my_filter['last_update_date']['lt'])
 
         for backend in backends:
             with self.subTest(backend=backend):
-                properties = backend.properties()
-                self.assertTrue(properties.last_update_date < filter_date)
+                properties = backend.properties(db_filter=my_filter)
```
luisg5: This is the line that fails. When the iteration reaches backend='ibmqx2', the attempt to get the backend properties throws the following Exception:

SubTest error: Traceback (most recent call last): File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/case.py", line 59, in testPartExecutor yield File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/case.py", line 533, in subTest yield File "/Users/luisgarcia/IBMQ/qiskit-ibmq-provider/test/ibmq/test_filter_backends.py", line 80, in test_get_properties_filter_date properties = backend.properties(db_filter=my_filter) File "/Users/luisgarcia/IBMQ/qiskit-ibmq-provider/qiskit/providers/ibmq/ibmqbackend.py", line 96, in properties self._properties = BackendProperties.from_dict(api_properties) File "/Users/luisgarcia/IBMQ/venvs/venv-qiskit-ibmq-provider/lib/python3.7/site-packages/qiskit/validation/base.py", line 352, in from_dict ex.messages, ex.field_names, ex.fields, ex.data, **ex.kwargs) from None qiskit.validation.exceptions.ModelValidationError: {'qubits': ['Missing data for required field.'], 'general': {0: {'date': ['Missing data for required field.'], 'unit': ['Missing data for required field.'], 'name': ['Missing data for required field.'], 'value': ['Missing data for required field.']}}, 'last_update_date': ['Missing data for required field.'], 'backend_name': ['Missing data for required field.'], 'gates': {0: {'qubits': ['Missing data for required field.'], 'parameters': {0: {'date': ['Missing data for required field.'], 'unit': ['Missing data for required field.'], 'name': ['Missing data for required field.'], 'value': ['Missing data for required field.']}}, 'gate': ['Missing data for required field.']}}, 'backend_version': ['Missing data for required field.']}

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -78,19 +78,21 @@ def run(self, qobj):
 
         return job
 
-    def properties(self, refresh=False):
+    #TODO: Update docstring for db_filter
+    def properties(self, refresh=False, db_filter=None):
```
diego-plan9: For this particular function, can you try to be a bit more restrictive with the new parameter, and instead of allowing a filter lean on a more simple type (the one that first comes to mind is ... a datetime or similar)?

The reason is that this one is mainly an user-facing function (as opposed to the rest of functions changed, which are more "internal") - and actually the one that will define the "user API" part of the issue. In general it is about balancing the different forces, and for user-facing functions a good thumb rule is to tend to simplicity: in this case, and considering the requirement was to give the user the "ability to retrieve an older calibration" and the filter for that use case will always be the same with the only variable part being the timestamp, it makes sense to avoid having the user worry about the specific format of the filter (at least, until requirements change :grin:  ).

test/ibmq/test_filter_backends.py
diego-plan9: This is likely due to no calibrations being available for the combination of timestamp + device used in the test - which actually is something that might happen "often enough" that we can intercept it during `backend.properties()`, in order to avoid the cryptic message reaching the user.

luisg5: Yeah, that seems to be the issue. After using a more recent date, there are no tests that fail. So, I will look into intercepting these cases.

qiskit/providers/ibmq/ibmqbackend.py
luisg5: Yeah, that makes more sense! I'm changing the parameter into a datetime, because it might be the most common and simple type.

Initially, I tried to replicate the way job filtering was done. But, I agree. This would make the filtering much more user friendly since it removes the complexity of the filtering format. Thanks for the tips @diego-plan9! 👍. I'll keep that in mind going forward as well.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -97,7 +98,11 @@ def properties(self, refresh=False, datetime_filter=None):
 
         if refresh or self._properties is None:
             api_properties = self._api.backend_properties(self.name(), extra_filter=api_filter)
-            self._properties = BackendProperties.from_dict(api_properties)
+            # Todo: from_dict fails because it could map an empty dictionary to the BackendProperties Schema
```
luisg5: When there are no configurations for a backend, api_properties is set to an empty dictionary. When trying to map the empty dictionary to a BackendProperties schema, a ModelValidationError is thrown, with the following message:

`{'qubits': ['Missing data for required field.'], 'general': {0: {'value': ['Missing data for required field.'], 'date': ['Missing data for required field.'], 'unit': ['Missing data for required field.'], 'name': ['Missing data for required field.']}}, 'gates': {0: {'qubits': ['Missing data for required field.'], 'parameters': {0: {'value': ['Missing data for required field.'], 'date': ['Missing data for required field.'], 'unit': ['Missing data for required field.'], 'name': ['Missing data for required field.']}}, 'gate': ['Missing data for required field.']}}, 'backend_version': ['Missing data for required field.'], 'last_update_date': ['Missing data for required field.'], 'backend_name': ['Missing data for required field.']}
`

As of right now, I'm returning the exception message above. Is there another message that is preferable to return, or is this message okay to return?

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -97,7 +98,11 @@ def properties(self, refresh=False, datetime_filter=None):
...
+            try:
+                self._properties = BackendProperties.from_dict(api_properties)
+            except ModelValidationError as ex:
+                self._properties = ex
```
luisg5: When there are no configurations for a backend, api_properties is set to an empty dictionary. When trying to map the empty dictionary to a BackendProperties schema, a ModelValidationError is thrown with the following message:

{'qubits': ['Missing data for required field.'], 'general': {0: {'value': ['Missing data for required field.'], 'date': ['Missing data for required field.'], 'unit': ['Missing data for required field.'], 'name': ['Missing data for required field.']}}, 'gates': {0: {'qubits': ['Missing data for required field.'], 'parameters': {0: {'value': ['Missing data for required field.'], 'date': ['Missing data for required field.'], 'unit': ['Missing data for required field.'], 'name': ['Missing data for required field.']}}, 'gate': ['Missing data for required field.']}}, 'backend_version': ['Missing data for required field.'], 'last_update_date': ['Missing data for required field.'], 'backend_name': ['Missing data for required field.']}

As of right now, I'm returning the exception message above. Is there another message that is preferable to return, or is this message okay to return?

luisg5:
I noticed there are some lint errors that did not pass. I will make sure to fix those.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -78,20 +79,28 @@ def run(self, qobj):
...
         return job
 
-    def properties(self, refresh=False):
+    def properties(self, refresh=False, datetime_filter=None):
         """Return the online backend properties.
 
         Args:
             refresh (bool): if True, the return is via a QX API call.
                 Otherwise, a cached version is returned.
+            datetime_filter (datetime.datetime): datetime to filter properties.
```
jyu00: The docstring should be clearer about what the user should expect. For example, by specifying a filter, this function will return an instance of the backend properties whose timestamp is closest to, but less than the specified datatime filter.

CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ The format is based on [Keep a Changelog].
 - The `backend.jobs()` function now has a default `limit` of 10 results, and
   in the IBM Q Experience v2 will automatically perform several API calls in
   order to retrieve the specified number of jobs, if higher. (\#263)
+- The `backend.properties()` function now accepts an optional `datetime_filter`
```
jyu00: Your changes should go under the `UNRELEASED` section

qiskit/providers/ibmq/api_v2/rest/backend.py
```diff
@@ -36,10 +37,19 @@ def __init__(self, session, backend_name):
         self.backend_name = backend_name
         super().__init__(session, '/devices/{}'.format(backend_name))
 
-    def properties(self):
+    def properties(self, filter=None):
         """Return backend properties."""
+        # pylint: disable=redefined-builtin
```
jyu00: Is there any reason to disable lint here instead of renaming your variable? It's general better to not use the same name as a builtin.

qiskit/providers/ibmq/api_v2/rest/backend.py
```diff
@@ -36,10 +37,19 @@ def __init__(self, session, backend_name):
         self.backend_name = backend_name
         super().__init__(session, '/devices/{}'.format(backend_name))
 
-    def properties(self):
+    def properties(self, filter=None):
```
jyu00: You should update the docstring to add the new parameter (this applies to some other functions you modified as well).

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -78,20 +78,32 @@ def run(self, qobj):
...
+    def properties(self, refresh=False, datetime_filter=None):
         """Return the online backend properties.
 
         Args:
             refresh (bool): if True, the return is via a QX API call.
                 Otherwise, a cached version is returned.
+            datetime_filter (datetime.datetime): By specifying a filter,
+                this function returns an instance of the BackendProperties whose
+                timestamp is closest to, but less than the specified datetime
+                filter.
```
jyu00: 1) For consistency, `By` should be lower case `by`
2) `older than` might be better than `less than` since it's a datetime comparison.

qiskit/providers/ibmq/api_v2/rest/backend.py
luisg5: I was using the job filtering functionality as reference. The last function that made the api call used `filter` as the parameter name, and it disabled the builtin warning as well. I know it's best not to use builtin names, it was more to be consistent with other areas of the code. Would it be best to change it to another name perhaps?

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -78,20 +78,32 @@ def run(self, qobj):
 
         return job
 
-    def properties(self, refresh=False):
-        """Return the online backend properties.
+    def properties(self, refresh=False, datetime_filter=None):
```
diego-plan9: After giving it some thought, can you rename the parameter to just `datetime`, for user-clarity?

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -78,20 +78,32 @@ def run(self, qobj):
...
+        api_filter = {}
+        if datetime_filter:
+            api_filter['last_update_date'] = {'lt': datetime_filter.isoformat()}
+
         if refresh or self._properties is None:
-            api_properties = self._api.backend_properties(self.name())
-            self._properties = BackendProperties.from_dict(api_properties)
+            api_properties = self._api.backend_properties(self.name(), api_filter=api_filter)
+            if not api_properties:
+                self._properties = None
```
diego-plan9: As informally discussed, can you make sure that if the argument is specified, no modification of `self._properties` or any functionality related to "caching" is involved?

The caching was introduced as asking for the properties of a backend happens a number of times during a typical Qiskit program (without the user explicitly asking for them in a number of cases), and it became a bit expensive to make an API call each time. However, in this case let's assume that asking for the properties of a backend _at a specified timestamp_ is a more specific operation, and fully under the control of the user, and he will be ok assuming the "cost" of the call each time (or that it is their responsibility to do some intelligent caching if needed).

test/ibmq/test_ibmq_provider.py
```diff
@@ -91,6 +92,19 @@ def test_remote_backend_properties(self):
...
+    def test_remote_backend_properties_filter_date(self):
+        """Test backend properties filtered by date."""
+        backends = self.provider.backends(simulator=False)
+
+        datetime_filter = datetime.fromisoformat('2019-02-01T00:00:00.000')
+        for backend in backends:
+            with self.subTest(backend=backend):
+                properties = backend.properties(datetime_filter=datetime_filter)
+                if isinstance(properties, BackendProperties):
+                    self.assertLess(properties.last_update_date, datetime_filter)
```
diego-plan9: Very small nit - perhaps `assertLessEqual()`?

Issue #281, opened 2019-07-29
Add notification of downloading results when using object storage


diego-plan9:
### What is the expected behavior?

As a new request by the API team, when a result is downloaded via object storage, we should notify that the download of the results has been successful.

jyu00:
Closed via #282

PR #282, opened 2019-07-29
Introduce sending a notification after downloading results

yeralin:
### Summary
As requested by the API team, now we are sending a notification in the form of POST request to 
`/jobResultDownloaded` URL after we download results using object storage.

### Details and comments
Implements #281


qiskit/providers/ibmq/api_v2/clients/account.py
```diff
@@ -154,7 +154,12 @@ def job_download_qobj_object_storage(self, job_id):
         download_url = job_api.download_url()['url']
 
         # Download the result from object storage.
-        return job_api.get_object_storage(download_url)
+        object_storage = job_api.get_object_storage(download_url)
```
diego-plan9: Minor, but can you find a name for this variable that better reflects its content? The function `job_api.get_object_storage` actually means "get (the job) via object storage" (and returns a dict that we can use for reconstructing a job) - off the top of my head, either going with a generic name (`response`) or a more specific one (`job_something`) sounds good.

yeralin: I can think of job_response/response_job, job_result, resulting_job 😄

yeralin: Renamed to `qobj_payload`

diego-plan9:
We can finally test, so removing `on hold` 👍

qiskit/providers/ibmq/api_v2/clients/account.py
diego-plan9: After the move to the right method, can you revise the variable name again?

yeralin: Idk Github is being weird again. If you go to `Files Changed`, you'll see:
```
# Download the result from object storage.
qobj_payload = job_api.get_object_storage(download_url)

# Notify the API via the callback
_ = job_api.callback_download()

return qobj_payload
```

jyu00:
The endpoint appears to be available and working. However, as discussed with @diego-plan9, it'd be nice if we `try/except` errors thrown by `callback_download()` so we can still return the already downloaded result to the user, especially since the user may not be able to re-retrieve the result. Maybe issue a warning instead when an exception is caught? And the test case can look for the warning and fail (the test).

yeralin:
I could secure it with the following:
```
import requests, logging

...
try:
    _ = job_api.callback_download()
except (requests.HTTPError, ValueError):
    logger.warning("Download completion callback failed")
```

Does the warning message sound ok?

jyu00:
How about "An error occurred while sending download completion acknowledgement. The results are still available for use." ? Since the message may be seen by the user we don't want them to be alarmed.

yeralin:
Hmmm, the part with `The results are still available for use` might be redundant since it is already a warning message i.e. nothing **horrible** has happened. What do you think?

jyu00:
That's fine. Between the fact it's a warning and not an error, and that the message says "download completion", it should be sufficient for the user to ignore it.

yeralin:
Hmm I'm thinking, should I also log the actual cause?
```
except (HTTPError, ValueError) as error:
    logger.warning("An error occurred while sending download completion acknowledgement: " + error)
```

jyu00:
Go for it. 

One other thing. `session` converts all `RequestException` to `RequestsApiError`, so that's the one you'll need to catch.

Issue #69, opened 2019-04-30
Revise skipped tests


diego-plan9:
### What is the expected behavior?

Currently we have [~19 skipped tests](https://travis-ci.com/Qiskit/qiskit-ibmq-provider/jobs/193654694#L480), for different reasons. We should re-asses the situation and decide if some of them can be enabled - as a preliminary review:
* a number of them come from `test_ibmq_qobj.py`, and are likely a remainder of the transition to qobj, as the file intent was originally intended to verify that backends were moved to supporting qobj (and they should be since some time ago). The individual tests are probably superseeded by other tests - and if not, they are candidates to moving them to other files.
* a number of them are `slow_tests` - as @ajavadia mentioned, we should run them in the CIs now that the test suite is relatively small, probably via cron or similar.
* some are due to features not available in the API yet - they are likely to stay skipped.

Max-Reuter:
Here’s a compilation of the current 19 skipped tests:

`test/ibmq/test_ibmq_qobj.py` - these are all skipped with the note `'No credentials or testing device available for testing Qobj capabilities.’` From what I can tell, these tests are unique and should be retained:
1.	Test if backend support Qobj.
2.	Test Atlantis deterministic ry operation.
3.	Test conditional operation.
4.	Test two circuits, two registers, out-of-order readout.
5.	Test one circuit, two registers, out-of-order readout.
6.	Test one circuit, one register, in-order readout.
7.	Test one circuit, one register, in-order readout.
8.	Test if backend is operational.
9.	Test one circuit, one register, out-of-order readout.
10.	Test one circuit, one register, in-order readout.

Most of the following tests are slow tests and are candidates for CI:
`test/ibmq/test_ibmq_backends.py`:

11.	Test that the qobj headers are passed onto the results for devices.
    - Apparently unique and necessary, skipped with the message ` skipped 'Skipping slow tests'`, a possible candidate for CI.

`test_ibmq_job.py`:
12.	Test job cancelation.
    - Apparently unique and necessary, skipped with the message ` skipped 'Skipping slow tests'`, a possible candidate for CI.
13.	The test `Test backend pulse defaults.` is skipped with the message `'Skipping until support in production API'`. Is there support in the production API for this now?

There are these tests from the same file that may have some overlap or could be combined with other tests. They are skipped with the message ` skipped 'Skipping slow tests'`, and are possible candidates for CI:

14.	Test that retrieved jobs come from their appropriate backend.
[...]

PR #287, opened 2019-07-30
Re-enable test.ibmq.test_proxies.TestProxies tests

yeralin:
## What is the expected behavior?
Related to #69
Currently, `test.ibmq.test_proxies.TestProxies` tests are skipped using `@skipIf` decorator
```
@skipIf(sys.version_info >= (3, 7), 'pproxy version not supported in 3.7')
```
The tests can be re-enabled if we update `pproxy` package from version `1.2.2` to `2.1.8` in `requirements-dev.txt` file.
Then, `test.ibmq.test_proxies.TestProxies` can be run under `>= 3.7` python versions.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,4 +2,4 @@ pycodestyle
 pylint>=2.3,<2.4
 pylintfileheader>=0.0.2
 vcrpy
-pproxy==1.2.2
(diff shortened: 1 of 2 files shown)


yeralin:
Seems like `2.1.8` version is not available to python=3.5

PR #290, opened 2019-07-31
Re-enable test.ibmq.test_proxies.TestProxies tests

### LINKED ISSUE #69: Revise skipped tests

### LINKED ISSUE #284: Re-enable TestProxies tests
## What is the expected behavior?
Related to #69
Currently, `test.ibmq.test_proxies.TestProxies` tests are skipped using `@skipIf` decorator
```
@skipIf(sys.version_info >= (3, 7), 'pproxy version not supported in 3.7')
```
The tests can be re-enabled if we update `pproxy` package from version `1.2.2` to `2.1.8` in `requirements-dev.txt` file.
Then, `test.ibmq.test_proxies.TestProxies` can be run under `>= 3.7` python versions.

yeralin:
## What is the expected behavior?
Fixes #284 and related to #69

Currently, `test.ibmq.test_proxies.TestProxies` tests are skipped using `@skipIf` decorator
```
@skipIf(sys.version_info >= (3, 7), 'pproxy version not supported in 3.7')
```

The tests can be re-enabled if have conditional dependencies defined in `requirements-dev.txt` i.e. install `pproxy==1.2.2` for python versions beq 3.5, and `pproxy==2.1.8` for above. We'd also have to use a different listen flag.


.idea/dictionaries/Daniyar_Yeralin.xml
```diff
@@ -0,0 +1,7 @@
+<component name="ProjectDictionaryState">
```
diego-plan9: These `.idea` files seem to have slipped through - can you remove them from the PR? We tend not to include IDE-specific files in the project `.gitignore`, but for convenience you might find setting `.git/info/exclude` locally or a similar approach handy.

requirements-dev.txt
```diff
@@ -3,4 +3,5 @@ pycodestyle
 pylint>=2.3,<2.4
 pylintfileheader>=0.0.2
 vcrpy
-pproxy==1.2.2
+pproxy==1.2.2; python_version > '2.6' and python_version <= '3.5'
```
diego-plan9: Actually, I preferred the previous iteration - if somebody is attempting to use Python 2, they will have bigger problems trying to run the package than the version of a development dependency :) Can you remove the bit about `2.6`, for simplicity?

test/ibmq/test_proxies.py
```diff
@@ -15,20 +15,18 @@
 """Tests for the AuthClient and VersionClient proxy support."""
 
 
-from unittest import skipIf
-import urllib
 import subprocess
-import sys
+import urllib
+from sys import version_info as python_version
```
diego-plan9: Minor (and style-related), but in general we avoid using an alias when importing unless strongly justified (ie. in case of conflicts, etc). The rationale is to avoid one extra layer of indirection - on a global scale, when re-reading the code it is arguably faster to mentally parse just `version_info` or `sys.version_info` directly, specially if the reader has played with that part of the standard library.

test/ibmq/test_proxies.py
```diff
@@ -37,15 +35,15 @@
 INVALID_ADDRESS_PROXIES = {'https': '{}:{}'.format('invalid', PORT)}
 
 
-@skipIf(sys.version_info >= (3, 7), 'pproxy version not supported in 3.7')
+@skipIf(python_version < (2, 7), 'pproxy is supported only on or above python 2.7')
```
diego-plan9: Slightly related to the comment on the .txt file, and to the last comment in the issue - the idea was to keep skipping if an invalid combination was found, ie. something similar to:
* Python>3.5 and pproxy==1.2.2
* Python==3.5 and pproxy > 1.2.2

Can you update the `skipIf` accordingly?

test/ibmq/test_proxies.py
```diff
@@ -37,15 +35,15 @@
...
 class TestProxies(IBMQTestCase):
     """Tests for proxy capabilities."""
 
     def setUp(self):
         super().setUp()
-
+        listen_flag = '-i' if python_version <= (3, 5) else '-l'
```
diego-plan9: For maintainability, can you make this `if` based on the pproxy version instead of the python version?

PR #291, opened 2019-07-31
Made datetimes timezone unaware

### LINKED ISSUE #288: Revise test_remote_backend_properties_filter_date

A couple of issues arose from the test included #277:
* the test is using `.fromisoformat()`, which is [only available in 3.7+](https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat) - can we find an alternative way that uses only Pyhon 3.5 constructs?
* the comparison between `properties.last_update_date` and the filter might fail due to time zone issues:

```
    def test_remote_backend_properties_filter_date(self):
        """Test backend properties filtered by date."""
        backends = self.provider.backends(simulator=False)
    
        datetime_filter = datetime.fromisoformat('2019-02-01T00:00:00.000')
        for backend in backends:
                properties = backend.properties(datetime=datetime_filter)
                if isinstance(properties, BackendProperties):
>                   self.assertLessEqual(properties.last_update_date, datetime_filter)
E                   TypeError: can't compare offset-naive and offset-aware datetimes
```

luisg5:
### Summary

Fixes #288 

### Details and comments

Made the datetime objects timezone unaware. This allows them to be compared without issues.

### DIFF
--- a/qiskit/providers/ibmq/ibmqbackend.py
+++ b/qiskit/providers/ibmq/ibmqbackend.py
@@ -98,6 +98,7 @@ class IBMQBackend(BaseBackend):
                 warnings.warn('Retrieving the properties of a '
                               'backend in a specific datetime is '
                               'only available when using IBM Q v2')
+                return None
 
             # Do not use cache for specific datetime properties.
             api_properties = self._api.backend_properties(self.name(), datetime=datetime)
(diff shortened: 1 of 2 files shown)


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -94,6 +94,11 @@ def properties(self, refresh=False, datetime=None):
         """
         # pylint: disable=arguments-differ
         if datetime:
+            if not isinstance(self._api, BaseClient):
+                warnings.warn('Retrieving the properties of a '
+                              'backend in a specific datetime is '
+                              'only available when using IBM Q v2')
```
diego-plan9: Can you `return None` at this point?

Issue #265, opened 2019-07-18
Add type hints to the package


diego-plan9:
### What is the expected behavior?

For maintainability and keeping up with good practices, it would be nice to add [type hints](https://docs.python.org/3.5/library/typing.html) to the package, in the same spirit as the pulse terra module.

The idea would be to introduce them gradually, aiming to minimize conflicts and for allowing contributors to work on different pieces - a naive approach would be on a file/module basis. There are a number of files where it probably makes sense to wait until the deprecation of `v1` is complete - namely `ibmq/api`, the files that for the time being directly expect v1/v2 objects, and possibly others.

This issue does not have a strong milestone assigned, and hopefully will be collaborative - just ping when intending to take on"typing"  a chunk of the project!

PR #293, opened 2019-08-02
Added type hints to api_v2/client modules

luisg5:
### Summary

Adresses #265 

### Details and comments

Added type hints to auth.py, version.py, and account.py, within the api_v2 directory.

### DIFF
--- a/qiskit/providers/ibmq/api_v2/clients/account.py
+++ b/qiskit/providers/ibmq/api_v2/clients/account.py
@@ -16,7 +16,7 @@
 
 import asyncio
 
-from typing import List, Dict, Union, Any
+from typing import List, Dict, Any
 # Disabled unused-import because datetime is used only for type hints.
 from datetime import datetime  # pylint: disable=unused-import
 
@@ -27,8 +27,6 @@ from .base import BaseClient
 from .websocket import WebsocketClient
 
 # Type aliases for type hints.
-QobjDict = Dict[str, Union[str, List[Dict[str, Any]], Dict[str, Any]]]
-JobStatus = Dict[str, str]
 
 
 class AccountClient(BaseClient):
@@ -66,7 +64,7 @@ class AccountClient(BaseClient):
         """
         return self.client_api.backends()
 
-    def backend_status(self, backend_name: str) -> Dict[str, Union[str, int, bool]]:
+    def backend_status(self, backend_name: str) -> Dict[str, Any]:
         """Return the status of a backend.
 
         Args:
@@ -81,7 +79,7 @@ class AccountClient(BaseClient):
             self,
             backend_name: str,
             datetime: datetime = None
-    ) -> Dict[str, Union[str, List[Dict]]]:
+    ) -> Dict[str, Any]:
         """Return the properties of a backend.
 
         Args:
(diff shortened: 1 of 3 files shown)


qiskit/providers/ibmq/api_v2/clients/auth.py
```diff
@@ -13,18 +13,23 @@
...
 """Client for accessing authentication features of IBM Q Experience."""
+from typing import Dict, List
 
 from ..exceptions import AuthenticationLicenseError, RequestsApiError
 from ..rest import Api, Auth
 from ..session import RetrySession
 
 from .base import BaseClient
 
+# Type aliases for type hints.
```
luisg5: Some annotations were used multiples times, so I defined type aliases.

qiskit/providers/ibmq/api_v2/clients/account.py
```diff
@@ -16,20 +16,34 @@
...
+
 
 class AccountClient(BaseClient):
     """Client for accessing an individual IBM Q Experience account.
 
     This client provides access to an individual IBM Q hub/group/project.
     """
 
-    def __init__(self, access_token, project_url, websockets_url, **request_kwargs):
+    def __init__(
```
luisg5: As a result of type hints, the length of some functions were beyond the line length limit. I decided handle it this way, but it's up in the air if there are better alternative suggestions.

jyu00:
I think for this module there isn't really much a need to specify the type _inside_ a dictionary, especially since many of them are just responses from the api, which could change in the future. For example, 
```
def backend_status(self, backend_name: str) -> Dict[str, Union[str, int, bool]]:```

Knowing that the values of the returned dictionary can be one of `str, int, bool`, IMO, doesn't really help the user much.

File: qiskit/providers/ibmq/api_v2/clients/auth.py (hunk 1 above)
Comment (diego-plan9) [reply qiskit/providers/ibmq/api_v2/clients/auth.py]: Can you actually remove the aliases for the time being? The mean reason is starting the effort "simple and explicit", and we can revise using aliases once we are a bit more advanced in the typing effort if needed.

File: qiskit/providers/ibmq/api_v2/clients/account.py [hunk 3]
```diff
@@ -16,20 +16,32 @@
...
+# Disabled unused-import because datetime is used only for type hints.
+from datetime import datetime  # pylint: disable=unused-import
+
 from ..rest import Api
 from ..session import RetrySession
 
 from .base import BaseClient
 from .websocket import WebsocketClient
 
+# Type aliases for type hints.
```
diego-plan9: This is probably a leftover from a previous iteration?

qiskit/providers/ibmq/api_v2/clients/auth.py
```diff
@@ -112,7 +117,7 @@ def user_hubs(self):
         """
         response = self.client_api.hubs()
 
-        hubs = []
+        hubs = []  # type: Hubs
```
diego-plan9: Let's not introduce type hints _inside_ the functions during this round - as with other comments, we can reevaluate once we have a wider picture (don't worry as well it is produces warnings, etc).

qiskit/providers/ibmq/api_v2/clients/account.py
luisg5: Yeah, this was leftover. Thanks for the catch! I'll make sure to remove it 👍🏽.

qiskit/providers/ibmq/api_v2/clients/auth.py
luisg5: It was the warnings that made me decide to add the type hints for the variables. But, that makes sense. I'll focus on the bigger picture as we go on for now. I removed them in the most recent PR.

PR #299, opened 2019-08-06
Added type hints to ibmqbackend.py

luisg5:
### Summary

Adresses #265 

### Details and comments

Marked as WIP, because I was looking to getting some feedback on the way I handled a circular dependency.

### DIFF
--- a/qiskit/providers/ibmq/ibmqbackend.py
+++ b/qiskit/providers/ibmq/ibmqbackend.py
@@ -17,18 +17,30 @@
 import logging
 import warnings
 
+from typing import Dict, List, Union, Optional, Any, TYPE_CHECKING
+from datetime import datetime  # pylint: disable=unused-import
 from marshmallow import ValidationError
 
+from qiskit.qobj import Qobj
 from qiskit.providers import BaseBackend, JobStatus
 from qiskit.providers.models import (BackendStatus, BackendProperties,
-                                     PulseDefaults)
+                                     PulseDefaults, BackendConfiguration)
 
-from .api import ApiError
+from .api import ApiError, IBMQConnector
 from .api_v2.clients import BaseClient
 from .apiconstants import ApiJobStatus, ApiJobKind
+from .credentials import Credentials
 from .exceptions import IBMQBackendError, IBMQBackendValueError
 from .job import IBMQJob
 from .utils import update_qobj_config
+
+# IBMQProvider is only used for type checking,
+# but importing it results in the circular dependency:
+#     ibmqbackend.py -> ibmqprovider.py -> ibmqsingleprovider.py
+# This conditional ensures the module is only imported during
+# type checking and not runtime, effectively handling the circular dependency.
+if TYPE_CHECKING:
+    from .ibmqprovider import IBMQProvider  # pylint: disable=cyclic-import
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -51,7 +43,7 @@ class IBMQBackend(BaseBackend):
     def __init__(
             self,
             configuration: BackendConfiguration,
-            provider: 'IBMQProvider',
+            provider,
```
luisg5: Decided to leave this variable without a type hint, because importing IBMQProvider, to use as a type, resulted in a cyclic import error. In the future, we plan to address errors like this. But, for now we are deciding to leave variables untyped, if type hinting them results in an error.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -388,7 +380,7 @@ def run(
             self,
             qobj: Qobj,
             backend_options: Optional[Dict] = None,
-            noise_model: Optional[Any] = None
+            noise_model=None
```
luisg5: There was no `NoiseModel` type in the repository, so I decided to leave it unmarked instead of specifying `Any`.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -34,14 +34,6 @@
 from .job import IBMQJob
 from .utils import update_qobj_config
 
-# IBMQProvider is only used for type checking,
-# but importing it results in the circular dependency:
-#     ibmqbackend.py -> ibmqprovider.py -> ibmqsingleprovider.py
-# This conditional ensures the module is only imported during
-# type checking and not runtime, effectively handling the circular dependency.
-if TYPE_CHECKING:
-    from .ibmqprovider import IBMQProvider  # pylint: disable=cyclic-import
```
luisg5: Removed since we don't want to handle the cyclic import error at this stage. See the comment below for more details.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -36,7 +40,13 @@
 class IBMQBackend(BaseBackend):
     """Backend class interfacing with an IBMQ backend."""
 
-    def __init__(self, configuration, provider, credentials, api):
+    def __init__(
+            self,
+            configuration: BackendConfiguration,
+            provider,
+            credentials: Credentials,
+            api: IBMQConnector
```
diego-plan9: Can you double-check this declaration? As part of the transition between the old and the new api, a good rule of thumb is that, most of the time, the objects that expected an `IBMQConnector` also accept the equivalent in the new API (client or similar). In general, we cannot rely solely on the existing docstrings, as in this case it seems it was not updated.

luisg5: Yeah, thanks for that catch Diego! I'll double check to make sure it is of the right type in that case.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -36,14 +40,20 @@
...
+            api: Union[AccountClient, IBMQConnector]
+    ) -> None:
         """Initialize remote backend for IBM Quantum Experience.
 
         Args:
             configuration (BackendConfiguration): configuration of backend.
             provider (IBMQProvider): provider.
             credentials (Credentials): credentials.
-            api (IBMQConnector):
+            api (Union[AccountClient, IBMQConnector]):
```
luisg5: Updated type hint to ensure it supports api_v2.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -36,14 +40,20 @@
 class IBMQBackend(BaseBackend):
     """Backend class interfacing with an IBMQ backend."""
 
-    def __init__(self, configuration, provider, credentials, api):
+    def __init__(
+            self,
+            configuration: BackendConfiguration,
+            provider,
+            credentials: Credentials,
+            api: Union[AccountClient, IBMQConnector]
```
luisg5: Updated type hint to ensure compatibility with api_v2.

Issue #118, opened 2019-05-23
Allow assigning a name to a job


nonhermitian:
### What is the expected behavior?
Moved from: https://github.com/Qiskit/qiskit-terra/issues/2406

As I have mentioned before (but cannot find the Issue for), we really need the ability to give a job a name (`job.name`), and to be able to query the database by this name, or a subset of characters in this name. The job id is good as a unique identifier, but less than ideal for actually labeling results for future retrieval. It would also be nice to set the name of a job after it is run, so that if I have good results I can say: job.set_name('These are good results').

nonhermitian:
To quote @jaygambetta's last comment:

> I agree we should do this.

> job.name --> returns the name
> job.set_name() = "blah" (do we need to have a set?)

> one question I have that we have punted on but I think we need to address before we do this is do > the job options gets associated with a backend or a provider.

> currently

> backend.jobs() - returns a list of previous jobs executed by this user on this backend.
> backend.retrieve_job() - returns a job by a job_id.

> but I think this actually does not do it by the backend (@diego-plan9) is this correct? It really does it > by the provider. If so should this be

> provider.jobs() - returns a list of previous jobs executed by this user on this backend.
> provider.retrieve_job() - returns a job by a job_id.

> and then the

> job = provider.search(name = "name")

nonhermitian:
To answer @jaygambetta's question, the backend itself makes the query to the API for its jobs, not the provider.

diego-plan9:
This would likely need to coordinate with the API team, and probably needs to be tackled after #104 to set all the pieces in place. One of the underlying topics is that the concept of a `Job` in Qiskit is not too heavily mapped to what a `Job` is for the API - ideally, we should make an effort for reconciling them and formalize them to a degree (ie. specs!) _before_ adding extra fields and functions, in order to try to avoid further disparities and their related side effects.

pacomf:
@nonhermitian the name needs to be unique in all jobs for an user? Or can be the same and you can return several jobs with the same name?

nonhermitian:
The job ID is the unique identifier.  The job name need not be.

pacomf:
ok!

nonhermitian:
@pacomf said that the API will already allow this, so I think the API action label can be removed.

ajavadia:
I agree this is extremely nice to have. Currently it is hard to reverse engineer what a job was supposed to do.

diego-plan9:
[...]

Issue #104, opened 2019-05-13
Refactor IBMQJob class


diego-plan9:
### What is the expected behavior?

Over time, `IBMQJob` has grown in complexity, and in particular the logic has grown a bit convoluted - the amount of indirection (specially for methods around the `status`) and corner cases is probably above the recommended threshold.

It would be worth exploring a refactor of its _internals_ post 0.3 release, focused on maintainability.

diego-plan9:
Closing after the final bit was included in #426 - great work team!

PR #300, opened 2019-08-08
Allow custom job name in backend.run()

jyu00:
### Summary
This is part of #118 . It allows a user to specify a job name when calling `backend.run()` and subsequently use it as a filter to `backend.jobs()`.

### Details and comments
I ended up adding a new attribute `_job_name` to `IBMQJob` since that's the most logical place. The downside is it might make people think that attribute is always available when it's not.

### DIFF
--- a/qiskit/providers/ibmq/api_v2/clients/account.py
+++ b/qiskit/providers/ibmq/api_v2/clients/account.py
@@ -124,7 +124,7 @@ class AccountClient(BaseClient):
                                     extra_filter=extra_filter)
 
     def job_submit(self, backend_name: str, qobj_dict: Dict[str, Any],
-                   job_name: str = None) -> Dict[str, Any]:
+                   job_name: Optional[str] = None) -> Dict[str, Any]:
         """Submit a Qobj to a device.
 
         Args:
@@ -138,7 +138,7 @@ class AccountClient(BaseClient):
         return self.client_api.submit_job(backend_name, qobj_dict, job_name)
 
     def job_submit_object_storage(self, backend_name: str, qobj_dict: Dict[str, Any],
-                                  job_name: str = None) -> Dict:
+                                  job_name: Optional[str] = None) -> Dict:
         """Submit a Qobj to a device using object storage.
 
         Args:
--- a/qiskit/providers/ibmq/api_v2/clients/account.py
+++ b/qiskit/providers/ibmq/api_v2/clients/account.py
@@ -123,30 +123,34 @@ class AccountClient(BaseClient):
         return self.client_api.jobs(limit=limit, skip=skip,
                                     extra_filter=extra_filter)
 
-    def job_submit(self, backend_name: str, qobj_dict: Dict[str, Any]) -> Dict[str, Any]:
+    def job_submit(self, backend_name: str, qobj_dict: Dict[str, Any],
+                   job_name: str = None) -> Dict[str, Any]:
         """Submit a Qobj to a device.
 
         Args:
             backend_name (str): the name of the backend.
             qobj_dict (dict): the Qobj to be executed, as a dictionary.
+            job_name (str): custom name to be assigned to the job.
 
         Returns:
             dict: job status.
         """
-        return self.client_api.submit_job(backend_name, qobj_dict)
+        return self.client_api.submit_job(backend_name, qobj_dict, job_name)
 
-    def job_submit_object_storage(self, backend_name: str, qobj_dict: Dict[str, Any]) -> Dict:
+    def job_submit_object_storage(self, backend_name: str, qobj_dict: Dict[str, Any],
+                                  job_name: str = None) -> Dict:
         """Submit a Qobj to a device using object storage.
 
         Args:
             backend_name (str): the name of the backend.
             qobj_dict (dict): the Qobj to be executed, as a dictionary.
+            job_name (str): custom name to be assigned to the job.
 
         Returns:
             dict: job status.
         """
         # Get the job via object storage.
-        job_info = self.client_api.submit_job_object_storage(backend_name)
+        job_info = self.client_api.submit_job_object_storage(backend_name, job_name=job_name)
 
         # Get the upload URL.
         job_id = job_info['id']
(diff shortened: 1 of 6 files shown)


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -153,7 +156,7 @@ def defaults(self, refresh=False):
 
         return self._defaults
 
-    def jobs(self, limit=10, skip=0, status=None, db_filter=None):
+    def jobs(self, limit=10, skip=0, status=None, db_filter=None, job_name=None):
```
diego-plan9: Can you reorder the arguments a bit, leaving `db_filter` as the last one? It has some penalty in regards to the signature of the function being backwards-compatible, but in this case I think we gain enough clarity by keeping the function parameters `(pagination, specific small filters, most generic filter)` (specially as the specific small filters are likely to be expanded in the future).

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -161,10 +165,12 @@ def __init__(self, backend, job_id, api, qobj=None,
             validate_qobj_against_schema(qobj)
             self._qobj_payload = qobj.to_dict()
             self._status = JobStatus.INITIALIZING
+            self._job_name = job_name
```
diego-plan9: Instead of setting an attribute, can you explore just passing the name to `.submit()`? As you mention, introducing a class-level attribute can be a bit confusing both for users ... and for maintainers (specially since the attribute is currently _not_ set for retrieved jobs, and the job class is quite convoluted). This hopefully will be enough for covering the requirements for this iterations, and we can include the name in a more controlled way during #104 .

test/ibmq/test_ibmq_job.py
```diff
@@ -433,6 +433,37 @@ def test_pulse_job(self, qe_token, qe_url):
...
+        # Use a unique job name
+        job_name = str(time.time()).replace('.', '')
+        job_ids = []
+        for _ in range(2):
+            job_ids.insert(0, backend.run(qobj, job_name=job_name).job_id())
+
+        job_list = backend.jobs(job_name=job_name)
+        self.assertEqual(len(job_list), 2)
+        job_list_ids = [job.job_id() for job in job_list]
+        self.assertEqual(job_ids, job_list_ids)
```
diego-plan9: Can you tweak the test for comparing using sets instead of list (or a similar approach), to prevent failing in some edge cases?

qiskit/providers/ibmq/api_v2/clients/account.py
```diff
@@ -123,30 +123,34 @@ def list_jobs_statuses(
         return self.client_api.jobs(limit=limit, skip=skip,
                                     extra_filter=extra_filter)
 
-    def job_submit(self, backend_name: str, qobj_dict: Dict[str, Any]) -> Dict[str, Any]:
+    def job_submit(self, backend_name: str, qobj_dict: Dict[str, Any],
+                   job_name: str = None) -> Dict[str, Any]:
```
diego-plan9: Pinging @luisg5 for recommendations on the type hints - `Optional`?

yeralin:
What happens if you pass an empty string as a `job_name`?

jyu00:
> What happens if you pass an empty string as a `job_name`?

It will be ignored per
```
        if job_name:
            api_filter['name'] = job_name
```

yeralin:
I guess what I was wondering is what are the constraints on the `job_name` parameter? (length, character encoding, allows special characters, etc.) 

I also could not find it on the API side. The API team hasn't implemented the feature yet?

diego-plan9:
Thanks @jyu00 ! During the merge at e0e50813020197809ef79f2cbea5d216ef07de7f, I had to solve some conflicts in the signatures which were a result of the merging of #299 - which seem to have gone nicely!

> I guess what I was wondering is what are the constraints on the job_name parameter? (max/min length, character encoding, allows special characters, etc.)

Good question - albeit something that can be answered at the API side: an error will be returned during the call if the parameters (including the job name) do not fulfill the constraints set by the API. If it ends up being too generic, we can revise and add some informative handling, but for the time being let's just rely on the API.

qiskit/providers/ibmq/api_v2/clients/account.py
luisg5: Thanks Diego! Yes, to be consistent with other type hints already defined, it might be best to specify the type hint for `job_name` as an `Optional[str]`. This would go for other function declarations as well. Essentially, it means the same thing, but abides to PEP 484, which we are trying to follow.

aasfaw:
@diego-plan9 I'm working on the latest master (see version below) and getting an error when I run `execute` without passing a `job_name`. Is the plan to make it mandatory to pass an empty string when `job_name` is unneeded? For the record, it works fine when I use a simulator as the backend, but breaks when I use a real quantum system as the backend. So seems like IBMQjob is the issue.

Code:
```
job = execute(circuit, backend = quantumsystem, shots=1024)
```

Version:
```
{‘qiskit-terra’: ‘0.9.0.dev0+b5d1511’,
‘qiskit-ignis’: ‘0.2.0.dev0+d5e7c7b’,
‘qiskit-aqua’: ‘0.5.6.dev0+01d5b1a’,
‘qiskit-ibmq-provider’: ‘dev-93fdcdb246a2b4ff4715a11f027c048b772e38ca’,
‘qiskit’: None,
‘qiskit-aer’: ‘0.2.3’}
```

Error:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-23-3909d03ec104> in <module>
----> 1 job = execute(circuit, backend = quantumsystem, shots=1024)
      2 from qiskit.tools.monitor import job_monitor
      3 job_monitor(job)

~/qiskit-terra/qiskit/execute.py in execute(experiments, backend, basis_gates, coupling_map, backend_properties, initial_layout, seed_transpiler, optimization_level, pass_manager, qobj_id, qobj_header, shots, memory, max_credits, seed_simulator, default_qubit_los, default_meas_los, schedule_los, meas_level, meas_return, memory_slots, memory_slot_size, rep_time, parameter_binds, **run_config)
    220 
    221     # executing the circuits on the backend and returning the job
--> 222     return backend.run(qobj, **run_config)

TypeError: run() missing 1 required positional argument: 'job_name'
```

jyu00:
@aasfaw Thank you for catching this! `job_name` is meant to be an optional parameter and was made required incorrectly during code merge. It will be fixed by #309.

PR #301, opened 2019-08-08
Added type hints to api_v2/rest

luisg5:
### Summary

Added type hints to the api_v2/rest directory.

### DIFF
--- a/qiskit/providers/ibmq/api_v2/rest/job.py
+++ b/qiskit/providers/ibmq/api_v2/rest/job.py
@@ -136,7 +136,10 @@ class Job(RestAdapterBase):
         return self.session.get(url, bare=True).json()
 
 
-def build_url_filter(excluded_fields: List[str], included_fields: List[str]) -> Dict[str, Dict[str, bool]]:
+def build_url_filter(
+        excluded_fields: List[str],
+        included_fields: List[str]
+) -> Dict[str, Dict[str, bool]]:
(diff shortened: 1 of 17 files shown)


qiskit/providers/ibmq/api_v2/rest/job.py
```diff
@@ -127,7 +130,7 @@ def get_object_storage(self, url):
         return self.session.get(url, bare=True).json()
 
 
-def build_url_filter(excluded_fields, included_fields):
+def build_url_filter(excluded_fields: List[str], included_fields: List[str]) -> Dict[str, Any]:
```
diego-plan9: Can you check if it is possible to be more specific about the return type? This function has bitten us in a number of issues and is a bit special compared to the rest of functions in the module (as in, it does not rely on a generic dict return by the api, but something we generate), and actually a good example where typing can be useful.

qiskit/providers/ibmq/api_v2/rest/job.py
```diff
@@ -133,7 +136,7 @@ def get_object_storage(self, url):
         return self.session.get(url, bare=True).json()
 
 
-def build_url_filter(excluded_fields, included_fields):
+def build_url_filter(excluded_fields: List[str], included_fields: List[str]) -> Dict[str, Dict[str, bool]]:
```
luisg5: Updated the dictionary type hint with the correct return type.

Issue #294, opened 2019-08-02
Revise compatibility with qiskit-terra 0.9


diego-plan9:
### What is the expected behavior?

With the upcoming release of `qiskit-terra`, we should ensure compatibility with the release.

jyu00:
I ran all the non-slow tests with `qiskit-terra` master branch and provider `master` and `0.3.1` branches. Everything passed except the ones using old api. Will run them again when we get closer to the release day.

diego-plan9:
Thanks for checking! Let's hope there are no big surprises as the release date approaches.

PR #303, opened 2019-08-12
Autocomplete of backend names

### LINKED ISSUE #205: Consider having autocomplete of backend names

As per suggestion by @ajavadia , it might be nice for usability purposes to allow the names of the backends to be auto-completed in IDEs and interpreters.

For example, we could end up with `provider.backends.poughkeepsie`, etc. There are a number of subtleties involved (forcing the backend names to be valid Python member identifiers, conventions, etc) that need to be looked into and balance - hence labelling as `discussion`!

jyu00:
### Summary
Fixes #205 .


qiskit/providers/ibmq/accountprovider.py
```diff
@@ -151,3 +166,50 @@ def __repr__(self):
 
         return "<{} for IBMQ({})>".format(
             self.__class__.__name__, credentials_info)
+
+
+class ProviderBackends(SimpleNamespace):
```
diego-plan9: Can you move this class to a separate file, for clarity?

diego-plan9:
Thanks @jyu00 ! Labelled as `on hold` until making sure #294 is closed, mostly to avoid including it if a new release of the provider needs to be done as a result. As soon as we have confirmation that the terra release went through and the provider is working correctly, feel free to merge!

PR #308, opened 2019-08-13
Introduce version response schema

yeralin:
### Summary
This PR introduces another schema used for `/version` endpoint.

### Details and comments
This schema contains a custom non-required field `new_api` that's used internally in IBMQ Provider component and defaults to `False`.


jyu00:
Since the main purpose of the schemas is to document communications with the api, the schemas should reflect the expected request/response contents. `new_api` is something the provider creates and not part of the actual response.

yeralin:
Removing `new_api`, I was expecting to be asked to remove it. Added it for awareness and tracing.

PR #319, opened 2019-08-21
Add accountclient test v2

luisg5:
### Summary

Added test case for AccountClient.backend_pulse_defaults.


test/ibmq/test_account_client_v2.py
```diff
@@ -145,6 +145,23 @@ def test_backend_properties(self):
         properties = api.backend_properties(backend_name)
         self.assertIsNotNone(properties)
 
+    def test_backend_pulse_defaults(self):
+        """Check the backend pulse defaults of each backend."""
+        api = self._get_client()
+        api_backends = [backend_info for backend_info
+                        in api.list_backends()]
```
jyu00: doesn't `api.list_backends()` already return a list?

luisg5: Yes, I don't know how that slipped past me 😅. Thank you Jessie.

Issue #289, opened 2019-07-31
Add schema for rest/auth endpoints


diego-plan9:
### What is the expected behavior?

This issue is for being able to kick off the documenting of the communication with the API effort: the endpoints at `rest/auth.py` are a good candidate, being quite restricted and simple in their scope, yet useful enough for setting the rest of the pieces in motion.

For this issue, let's:
* for each endpoint, add [`BaseSchema`](https://github.com/Qiskit/qiskit-terra/blob/master/qiskit/validation/base.py#L92) subclasses (one for the output, other for the input, if needed) that formalizes the messages. During this iteration, let's just focus on the fields the provider actually uses (ie. some endpoints will likely return more information).
* find the right place and naming convention, with the thinking that eventually will be scaled to all endpoints (probably simply `rest/schemas/auth.py` will work for the time being).
* make the pertinent changes in the internal documentation project.

yeralin:
Reading Marshmallow docs

diego-plan9:
Closed via #295 and others.

PR #321, opened 2019-08-21
Introduce root endpoints' schemas

yeralin:
### Summary
Addresses #289
Adds schemas that describe request and response payloads for "root" endpoints.

### Details and comments
I found couple problems while preparing this PR:
* Since we are using Marhsmallow 2.0, `marshmallow.fileds.Dict` is not fully supported
If you look at the official doc for `Dict` field under Marshmallow 2.0: https://marshmallow.readthedocs.io/en/2.x-line/api_reference.html#marshmallow.fields.Dict
It states: "This field is only appropriate when the structure of nested data is not known." i.e. if you use `mashmallow.fields.Dict` field under your schema, it will be deserialized as a simple `dict` object, no support for `marshmallow.fields.Nested`.
However, if you look at docs for `Dict` in Marshmallow 3.0: 
https://marshmallow.readthedocs.io/en/stable/api_reference.html#marshmallow.fields.Dict
You can explicitly set a nested schema for keys and values like: 
```
fields.Dict(keys=fields.Str(), values=fields.Float())
```
They discussed the implementation of this feature under https://github.com/marshmallow-code/marshmallow/issues/483
Because of this limitation schemas like `HubsResponseSchema` and `CircuitResponseSchema` can only be deserialized partially.

Also because of this issue, `marshmallow-jsonschema` package cannot generate a proper JSON schema for nested dicts, and rendered as `<not serializable>`:
image

> TL;DR we need to upgrade Marshmallow version to 3.0, if we want to fully deserialize schemas like `HubsResponseSchema` and `CircuitResponseSchema`

* Deserialization of `HubsResponseSchema` will always fail bc `@bind_schema` does not allow passing `many = True`

The issue was raised under `qiskit-terra` repo: https://github.com/Qiskit/qiskit-terra/issues/3021
If we were to create a model for `HubsResponseSchema` using `@bind_schema` decorator, and then use it to deserialize hubs response payload, the deserialization will always fail with `ModelValidationError`. The reason is bc the payload is of the following format: `[<payload>]` i.e. the payload is wrapped in an array. This could be easily fixed by passing `many = True` while creating an instance of `HubsResponseSchema(many=True)`, **but** the instance is created internally in `@bind_schema` decorator, and we *cannot*  pass any arguments inside.


qiskit/providers/ibmq/api_v2/rest/schemas/root.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # Required properties
+    order = String(required=True, default="creationDate DESC")
+    limit = Number(required=True, description="maximum number of items to return.")
+    skip = Number(required=True, description="offset for the items to return.")
+
+
+# Endpoint schemas.
+
+class HubsResponseSchema(BaseSchema):
```
diego-plan9: Can you double-check this schema? It seems to be missing one nesting level - the response contains a _list_ of hubs (each one containing a dict of groups, each one containing a dict of projects).

In this case, you can probably just use `many` (it being a list):

```python
hubs = Nested(SomethingAboutHubs, required=True, many=True)
```

qiskit/providers/ibmq/api_v2/rest/schemas/root.py
```diff
@@ -0,0 +1,167 @@
...
+    """Schema for HubsResponse"""
+
+    # Required properties.
+    name = String(required=True)
+    groups = Dict(Nested(GroupsResponseSchema), required=True)
+
+
+class VersionResponseSchema(BaseSchema):
+    """Schema for VersionResponse"""
+    pass
```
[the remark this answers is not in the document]
diego-plan9: No worries - actually, this should have the same request and response semantics as `auth.VersionResponseSchema` - can you double check it is the case, and if so, just remove it from this file?

qiskit/providers/ibmq/api_v2/rest/schemas/root.py
```diff
@@ -0,0 +1,167 @@
...
+    id = String(required=True, description="the job ID of an already submitted job.")
+    creationDate = String(required=True, description="when the job was run.")
+    status = String(required=True, description="`status` field directly from the API response.")
+
+
+class BackendsResponseSchema(BaseSchema):
+    """Schema for BackendResponse"""
+
+    # Required properties
+    backends = List(Nested(BackendConfigurationSchema, required=True))
```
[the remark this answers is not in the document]
diego-plan9: The best way is ... to rely on terra facilities, combined with relying on some bits of the schemas already defined elsewhere! Just using `BackendConfigurationSchema`, as you did, is fine - in theory, "schema inheritance" should take care of it  :+1:

qiskit/providers/ibmq/api_v2/rest/schemas/root.py
```diff
@@ -0,0 +1,167 @@
...
+    id = String(required=True)
+    status = String(required=True, validate=OneOf([status.value for status in JobStatus]))
+    creationDate = String(required=True)
+
+
+class JobsStatusRequestSchema(BaseSchema):
+    """Schema for JobsStatusRequest"""
+
+    # Required properties
+    filter = Nested(JobsStatusFilterRequestSchema, required=True)
```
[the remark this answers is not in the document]
yeralin: How about smth like `JobsStatusFilterQueryParamRequestSchema`?

qiskit/providers/ibmq/api_v2/rest/schemas/root.py
```diff
@@ -0,0 +1,167 @@
...
+
+class JobsRequestSchema(BaseSchema):
+    """Schema for JobsRequest"""
+
+    # Optional properties
+    name = String(required=False, description="custom name to be assigned to the job.")
+
+    # Required properties
+    qObject = Dict(required=True, description="the Qobj to be executed, as a dictionary.")
+    backend = Nested(BackendConfigurationSchema, required=True)
```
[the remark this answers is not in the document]
yeralin: Yep, you are right. Made a typo!

qiskit/providers/ibmq/api_v2/rest/schemas/root.py
yeralin: Exactly, that what I pointed out in the PR message:

* Deserialization of `HubsResponseSchema` will always fail bc `@bind_schema` does not allow passing `many = True`

The issue was raised under `qiskit-terra` repo: https://github.com/Qiskit/qiskit-terra/issues/3021
If we were to create a model for `HubsResponseSchema` using `@bind_schema` decorator, and then use it to deserialize hubs response payload, the deserialization will always fail with `ModelValidationError`. The reason is bc the payload is of the following format: `[<payload>]` i.e. the payload is wrapped in an array. This could be easily fixed by passing `many = True` while creating an instance of `HubsResponseSchema(many=True)`, **but** the instance is created internally in `@bind_schema` decorator, and we *cannot*  pass any arguments inside.

yeralin: Otherwise, I'd have to have *another* level of nestness. Like:
```
class HubResponseSchema(BaseSchema):
    """Schema for HubResponse"""

    # Required properties.
    name = String(required=True)
    groups = Dict(Nested(GroupsResponseSchema), required=True)

class HubsResponseSchema(BaseSchema):
    """Schema for HubsResponse"""

    # Required properties.
    huns = Nested(HubResponseSchema, required=True, many=True)
```

yeralin: So, it is either another schema *or* upgrading `@bind_schema` to support `many=True`

yeralin: @diego-plan9 Here is what I found.

Two schemas approach won't work! It will only work for payloads of the type:
```
{
    "hubs": [ 
        {<hub_1>},
        ...
    ]
}
```

But in our case, the payload is:
```
[
{<hub_1>},
...
]
```

That's why the only option is still having `@bind_schema(HubsResponseSchema, many=True)`
In order to deserialize "raw" collection.

Let me know if you need any clarification.

I'll try to create a quick code sample to reproduce this

yeralin: Here is the code sample:
```
from unittest import TestCase

from marshmallow import Schema
from marshmallow.fields import Nested, String, List



class NestedTestSchema(Schema):
    test = String(required=True)

class ManyTestSchema(Schema):
    nested = Nested(NestedTestSchema, many=True)

class ListTestSchema(Schema):
    nested = List(Nested(NestedTestSchema))

class SchemaTest(TestCase):

    def test_many_and_list_schemas_same(self):
        payload = { 'nested': [
                        {'test': 'something'}
                    ]
                  }
        many_deserialized = ManyTestSchema().load(payload).data
        list_deserialized = ListTestSchema().load(payload).data
        self.assertEqual(many_deserialized, list_deserialized)

    def test_cannot_deserialize_raw_list(self):
        payload = [
                    {'test': 'something'}
                  ]
        self.assertTrue(ManyTestSchema().load(payload).errors)
        self.assertTrue(ListTestSchema().load(payload).errors)
        self.assertTrue(ManyTestSchema(many=True).load(payload).data)
        self.assertTrue(ListTestSchema(many=True).load(payload).data)
```

For `ManyTestSchema` and `ListTestSchema` I wanted to point out that there is no difference in using either `Nested(NestedTestSchema, many=True)` or `List(Nested(NestedTestSchema))`.

diego-plan9:
Following up on `HubsResponseSchema`: let's decouple the issue with that particular response in order to move forward, as at this point in time there is no need for worrying about serialization and we can handle that response via other means.

Can you make `HubsResponseSchema` a dummy schema (with no fields), and add a small note in the class to remind us that it needs to be treated slightly differently for the time being?

PR #323, opened 2019-08-22
Introduce job endpoints schemas

yeralin:
### Summary
Addresses #289
Adds schemas that describe response payloads for "job" endpoints.

### Details and comments
Dummy schemas were created for `job.properties` (`PropertiesResponseSchema`) and `job.callback_download` (`CallbackDownloadResponseSchema`) since we are not utilizing returned payloads for those endpoints.


qiskit/providers/ibmq/api_v2/rest/schemas/job.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+# Helper schemas.
+
+class InfoQueueResponseSchema(BaseSchema):
+    """Nested schema for StatusResponseSchema"""
+
+    # Required properties
+    status = String(required=True)
+    position = Number(required=True, missing=0)
```
jyu00: Can it be both `required` and have a default (`missing`) value?

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class StatusResponseSchema(BaseSchema):
+    """Schema for StatusResponse"""
+
+    # Optional properties
+    infoQueue = Nested(InfoQueueResponseSchema, required=False)
+
+    # Required properties
+    status = String(required=True, validate=OneOf([status.name for status in ApiJobStatus]))
```
jyu00: While it makes no difference in this care (`ApiJobStatus`), it's better to do `status.value` instead of `status.name`, since we do `ApiJobStatus(api_response['status'])`, which requires the _value_ of the enum

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class CancelResponseSchema(BaseSchema):
+    """Schema for CancelResponse"""
+
+    # Optional properties
+    error = String(required=False)
+
+    # Required properties
+    cancelled = Boolean(required=True, description="True if job can be cancelled, else False. "
```
jyu00: I don't think we specifically look for this property. Yes it's weird to have just `error` as the only one. Perhaps we can reintroduce this after confirming with api *and* update the mainline code.

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
```diff
@@ -0,0 +1,100 @@
...
+class PropertiesResponseSchema(BaseSchema):
+    """Schema for PropertiesResponse"""
+    pass
+
+
+class StatusResponseSchema(BaseSchema):
+    """Schema for StatusResponse"""
+
+    # Optional properties
+    infoQueue = Nested(InfoQueueResponseSchema, required=False)
```
jyu00: I don't know why some have `description` set and others don't, but this one could probably use one since it's not obvious what this property is for.

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
```diff
@@ -0,0 +1,100 @@
...
+
+class InfoQueueResponseSchema(BaseSchema):
+    """Nested schema for StatusResponseSchema"""
+
+    # Required properties
+    status = String(required=True)
+    position = Number(required=True, missing=0)
+
+
+class JobResponseSchema(BaseSchema):
```
jyu00: If this serves strictly as a nested schema for `CallbackUploadResponseSchema`, then `JobResponseSchema` is not a very good name since there are a lot fo `/Jobs` related endpoints (and hence responses)

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
```diff
@@ -0,0 +1,100 @@
...
+    """Nested schema for CallbackUploadResponseSchema"""
+    # pylint: disable=invalid-name
+
+    # Required properties
+    id = String(required=True)
+    kind = String(required=True)
```
jyu00: For `callback_upload` path, I can also see `error`, `creationDate`, and `id` as properties used.

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
yeralin: According to @diego-plan9 I can only set descriptions *iff* there is already one in the docs or code. That's why you don't see them everywhere. I wanted to compose "my descriptions", but Diego asked me to remove them :) He mentioned that we will address them later

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
yeralin: Interestingly, it can be! I just ran a quick test. If `position` is missing, it is set to `0` which is a little counterintuitive since if we remove `missing = 0` and pass a payload w/o `position`, it will raise `ValidationError`.

https://marshmallow.readthedocs.io/en/2.x-line/api_reference.html#marshmallow.fields.Field

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
yeralin: We inexplicably do! 

Under `qiskit.providers.ibmq.job.ibmqjob:288` which checks for `cancelled` field in a response of the form `{'cancelled': True}`.

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
yeralin: Hmmm, yes I agree. Maybe we should come up with some better naming policy? Nothing good comes to mind.

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
yeralin: Hmmm I could not find anything. The only place I saw is `test_job_submit_object_storage` under `test_account_client_v2` where it only uses: `job['id']` and `job['kind']`.

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
jyu00: Are we looking at the same code? To me that [block](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/job/ibmqjob.py#L288) has 
```
        try:
            response = self._api.cancel_job(self._job_id)
            self._cancelled = 'error' not in response
```

yeralin: Yep, I misread sorry. I was reading it as `'cancelled' in response`, I'll remove this field from the schema. Thank you! :)

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
jyu00: It's line [487-495](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/job/ibmqjob.py#L487) in ibmqjob. The `job` in `callback_upload` response is returned as `submit_info` by `job_submit_object_storage`

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
jyu00: how about `CallbackUploadResponseJobSchema` for now? we'll likely combine this with other response schemas later since multiple endpoints return the same "job" results.

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
jyu00: Do you want to make them not required then?

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
yeralin: That is so strange, my IDE doesn't show the call in hierarchy!
image

However, yes it is called in `_submit_callback` method under `ibmqjob`.

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
yeralin: `CallbackUploadResponseNestedJobSchema`?

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
yeralin: Also, I see these two calls:
```
submit_info.get('creationDate')
```
Since it uses `get` method instead of `[]` operator, should I make `creationDate` field optional? Or am I overthinking it?

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+# Helper schemas.
+
+class InfoQueueResponseSchema(BaseSchema):
+    """Nested schema for StatusResponseSchema"""
+
+    # Optional properties
+    position = Integer(required=False, missing=0)
+    status = String(required=False, missing=False)
```
jyu00: The default status, of type `String`, is `False`? I think you misread the code. It assigns the default return value (_not_ the default `queue_status`) to False.

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
```diff
@@ -0,0 +1,101 @@
...
+# Helper schemas.
+
+class InfoQueueResponseSchema(BaseSchema):
+    """Nested schema for StatusResponseSchema"""
+
+    # Optional properties
+    position = Integer(required=False, missing=0)
+
+    # Required properties
+    status = String(required=True)
```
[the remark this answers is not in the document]
yeralin: Totally right, I rushed with this commit.

qiskit/providers/ibmq/api_v2/rest/schemas/job.py
yeralin: Totally right, I rushed this change. Fixed!

Issue #322, opened 2019-08-21
Retry websocket if connection closed


jyu00:
### What is the expected behavior?
Websocket connection may get closed due to the server pod being restarted, client machine going to sleep, etc. It'd be nice to retry once or twice. This is requested by the api team.

PR #324, opened 2019-08-23
Add retry logic in case of websocket connection closing.

lgarc15:
### Summary

Addresses #322 

### Details and comments

Added logic to retry getting the status of a job, in case of the websocket connection closing. A test still needs to be added, but pushed in order to receive feedback.

### DIFF
--- a/test/ibmq/websocket/test_websocket.py
+++ b/test/ibmq/websocket/test_websocket.py
@@ -30,7 +30,7 @@ from ...ibmqtestcase import IBMQTestCase
 from .websocket_server import (
     TOKEN_JOB_COMPLETED, TOKEN_JOB_TRANSITION, TOKEN_WRONG_FORMAT,
     TOKEN_TIMEOUT, TOKEN_WEBSOCKET_RETRY_SUCCESS,
-    TOKEN_WEBSOCKET_RETRY_LIMIT_EXCEEDED, TOKEN_WEBSOCKET_JOB_NOT_FOUND,
+    TOKEN_WEBSOCKET_RETRY_FAILURE, TOKEN_WEBSOCKET_JOB_NOT_FOUND,
     websocket_handler)
 
 TEST_IP_ADDRESS = '127.0.0.1'
@@ -127,7 +127,7 @@ class TestWebsocketClientMock(IBMQTestCase):
     def test_websocket_retry_failure(self):
         """Test exceeding the retry limit for retrieving a job status."""
         client = WebsocketClient('ws://{}:{}'.format(
-            TEST_IP_ADDRESS, VALID_PORT), TOKEN_WEBSOCKET_RETRY_LIMIT_EXCEEDED)
+            TEST_IP_ADDRESS, VALID_PORT), TOKEN_WEBSOCKET_RETRY_FAILURE)
         with self.assertRaises(WebsocketError):
             _ = asyncio.get_event_loop().run_until_complete(
                 client.get_job_status('job_id'))
--- a/test/ibmq/websocket/websocket_server.py
+++ b/test/ibmq/websocket/websocket_server.py
@@ -25,7 +25,7 @@ TOKEN_JOB_TRANSITION = 'token_job_transition'
 TOKEN_TIMEOUT = 'token_timeout'
 TOKEN_WRONG_FORMAT = 'token_wrong_format'
 TOKEN_WEBSOCKET_RETRY_SUCCESS = 'token_websocket_retry_success'
-TOKEN_WEBSOCKET_RETRY_LIMIT_EXCEEDED = 'token_websocket_retry_limit_exceeded'
+TOKEN_WEBSOCKET_RETRY_FAILURE = 'token_websocket_retry_failure'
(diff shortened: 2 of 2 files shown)


qiskit/providers/ibmq/api_v2/clients/websocket.py
```diff
@@ -150,64 +153,81 @@ def get_job_status(self, job_id, timeout=None):
...
+                        raise WebsocketError('Connection with websocket closed '
+                                             'unexpectedly: {}'.format(message)) from ex
+            except WebsocketError as ex:
+                # Maxed out the number of retries.
+                if current_retry >= retry_max:
+                    raise ex
+                # Job id not found. No point of retrying.
+                if connection_closed_ex_code == 4003:
+                    raise ex
+
```
luisg5: Most logical way I could think of retrying was to catch the exception and decide whether a retry attempt should be made.

qiskit/providers/ibmq/api_v2/clients/websocket.py
```diff
@@ -150,64 +153,81 @@ def get_job_status(self, job_id, timeout=None):
...
+                current_retry = current_retry + 1
+                # Retry: Continue next iteration after finally is executed.
+                continue
```
luisg5: Interestingly, a `finally` clause is executed "on the way" if it encounters certain statements, such as a `continue` statement within a `try` or `except` clause. Essentially, this means that the `finally` clause will be executed right after the `continue` statement and right before the next iteration of the loop it lives within. In this case, the websocket connection will be closed before attempting to grab another websocket connection and retrying.

qiskit/providers/ibmq/api_v2/clients/websocket.py
```diff
@@ -140,6 +140,9 @@ def get_job_status(self, job_id, timeout=None):
         Args:
             job_id (str): id of the job.
             timeout (int): timeout, in seconds.
+            retry_max (int): max number of retries to
```
jyu00: I don't think the retry needs to be a parameter since we don't want to retry more than once anyway

qiskit/providers/ibmq/api_v2/clients/websocket.py
```diff
@@ -150,64 +153,81 @@ def get_job_status(self, job_id, timeout=None):
...
-                warnings.filterwarnings("ignore", category=DeprecationWarning)
-                yield from websocket.close()
-
-        return last_status
+        current_retry = 0
+        while current_retry <= retry_max:
+            websocket = yield from self._connect(url)
+
+            connection_closed_ex_code = None
+            original_timeout = timeout
```
jyu00: You're essentially doubling the user specified timeout value by resetting it every time

qiskit/providers/ibmq/api_v2/clients/websocket.py
jyu00: You should be able to just check for the error code inside the `except` block of 197 and do a reconnect instead of raising. If you still end up doing this additional `try`, you might want to break the block into helper functions to make the code more readable.

qiskit/providers/ibmq/api_v2/clients/websocket.py
luisg5: I thought having it specified as a parameter would allow for more flexibility in case we decide to retry more than once. But I agree, once should be enough 👍 . I'll go ahead an make the change.

qiskit/providers/ibmq/api_v2/clients/websocket.py
luisg5: Didn't actually catch that, thanks Jessie!

qiskit/providers/ibmq/api_v2/clients/websocket.py
luisg5: I was thinking of reconnecting within the `except` block of 197. But if the connection is reestablished inside the `except` block, wouldn't the `finally` block close the connection right after, since it always gets executed after?

qiskit/providers/ibmq/api_v2/clients/websocket.py
```diff
@@ -150,64 +153,81 @@ def get_job_status(self, job_id, timeout=None):
...
+                        # 4001: closed due to an internal errors
+                        # 4002: closed on purpose (no more updates to send)
+                        # 4003: closed due to job not found.
+                        connection_closed_ex_code = ex.code
+                        message = 'Unexpected error'
+
+                        if ex.code == 4001:
+                            message = 'Internal server error'
+                        elif ex.code == 4002:
+                            break
```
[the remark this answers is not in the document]
luisg5: That's a good catch, thanks @yeralin. 

I think we might shy away from having multiple retry attempts and only allow for one retry. So, it makes sense to go with something like the latter (a boolean marker), as you specified.

qiskit/providers/ibmq/api_v2/clients/websocket.py
luisg5: Yes, that makes perfect sense! I was thinking of things from a different perspective, but I think the way you mentioned might keep things more simplified and tidy.

PR #329, opened 2019-09-03
IBMQJob refactor

jyu00:
### Summary
First pass to address #104. Main changes are:
- Move the decision and call of using object storage and websockets to `AccountClient`
- Bind `IBMQJob` to a schema
- Move job submit to `IBMQBackend` and make it synchronous
- Simplify state changes

### Details and comments
Some follow-ups needed:
- Make `IBMQJob` constructor parameters adhere to Python standards
- Provide the ability to refresh a job's attributes


qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -49,157 +51,149 @@ class IBMQJob(BaseJob):
...
-    Attributes:
-        _executor (futures.Executor): executor to handle asynchronous jobs
     """
-    _executor = futures.ThreadPoolExecutor()
 
-    def __init__(self, backend, job_id, api, qobj=None,
-                 creation_date=None, api_status=None,
-                 use_object_storage=False, use_websockets=False):
+    def __init__(self,
+                 backend_obj: Optional[BaseBackend],
```
luisg5: Trying to figure out if having `Optional` is a typo or maybe I'm missing something. In what cases would `backend_obj` be `None`?

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -49,157 +51,149 @@ class IBMQJob(BaseJob):
...
+                 job_id: Optional[str],
```
luisg5: Same as above regarding "defaulting to `None`"

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -49,157 +51,149 @@ class IBMQJob(BaseJob):
...
+                 api: AccountClient,
+                 qobj: Optional[Qobj] = None,
+                 name: Optional[str] = None,
+                 use_websockets: bool = True,
+                 **kwargs: Dict) -> None:
```
luisg5: The type annotation should be `Any` for `kwargs`. It might be a bit unintuitive, but mypy expands `Any` to be of type `Dict[Any, Any]`.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -391,249 +344,127 @@ def error_message(self):
...
                     'Timeout while waiting for job {}'.format(self._job_id))
+        self._update_status_position(status_response)
+        return self._status in required_status
 
-            logger.info('status = %s (%d seconds)', self._status, elapsed_time)
-            time.sleep(wait)
-
-    def _wait_for_final_status_websocket(self, timeout=None):
-        """Wait until the job progress to a final state using websockets.
+    def _init_job_model(self, **kwargs: Dict) -> None:
```
luisg5: Similar to an above comment. The actual type for `kwargs` should be `Any`.

luisg5:
The `status` and `_get_job` methods in `ibm/job/circuitjob.py` are missing return type annotations:
- `status` should be returning `str`
- `_get_job` should be returning `Dict[str, Any]`

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -391,249 +344,127 @@ def error_message(self):
 
         return self._api_error_msg
 
-    def queue_position(self):
+    def queue_position(self) -> int:
```
luisg5: Maybe a small mishap 😅. But, since `None` could be returned, the the type should be `Optional[int]`

qiskit/providers/ibmq/job/utils.py
```diff
@@ -63,3 +84,24 @@ def build_error_report(results: List[Dict[str, Any]]) -> str:
...
+        api_status (ApiJobStatus): API job status
+
+    Returns:
+        JobStatus: job status
+    """
+    return API_TO_JOB_STATUS[api_status]
+
+
+@contextmanager
+def api_to_job_error():
```
luisg5: The function is missing a return type annotation. Since nothing is returned, `None` should do.

qiskit/providers/ibmq/job/ibmqjob.py
jyu00: So that's how it was before... `job_id` is documented (and expected) to be either `str` or `None`, yet there is no default provided. If I give it a default I'd have to change the argument positions (from `job_id`, `api` to `api`, `job_id`), but that _should_ be ok since `IBMQJob` is not meant to be instantiated externally.

qiskit/providers/ibmq/job/ibmqjob.py
jyu00: It's `Optional` because of `CircuitJob`, which is a child of `IBMQJob`. In `CircuitsManager`, `CircuitJob` is initialized without a backend. I don't want to make it optional since it's expected for all but circuit jobs. Let me discuss with @diego-plan9 .

jyu00:
> The `status` and `_get_job` methods in `ibm/job/circuitjob.py` are missing return type annotations:
> 
> * `status` should be returning `str`
> * `_get_job` should be returning `Dict[str, Any]`

`circuitjob.py` never had type annotations, which I suspect is because the feature is still incomplete.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -76,19 +80,58 @@ def run(self, qobj: Qobj, job_name: Optional[str] = None) -> IBMQJob:
...
+            'backend_obj': self,
+            'api': self._api,
+            'qobj': qobj
+        })
+        try:
+            job = IBMQJob.from_dict(submit_info)
+        except ModelValidationError as err:
+            raise IBMQBackendError('Unexpected return value from the server when '
+                                   'submitting job: {}'.format(str(err)))
+        Publisher().publish("ibmq.job.start", job)
```
diego-plan9: Can you add a note to the docstring about the event, moving the previous one at https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/job/ibmqjob.py#L427?

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -391,249 +317,80 @@ def error_message(self):
...
 
         Raises:
-            JobError: If we have already submitted the job.
+            JobError: If an error occurred during job submit.
         """
-        # TODO: Validation against the schema should be done here and not
-        # during initialization. Once done, we should document that the method
-        # can raise QobjValidationError.
-        if self._future is not None or self._job_id is not None:
+        if self.job_id() is not None:
```
diego-plan9: Can you check if we can remove this `if` (and a similar one at `.error_message()`)? Hopefully we can now guarantee that all jobs have an `id`.

qiskit/providers/ibmq/api/clients/schema.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
```
diego-plan9: Non blocking, and mostly a note for follow-ups (as I think the decision had to do with the circular import quick fix) - for organizational purposes, we should find a different place for this file, keeping the `clients/` module just for clients.

qiskit/providers/ibmq/api/clients/account.py
```diff
@@ -67,7 +76,8 @@ def list_backends(self, timeout: Optional[float] = None) -> List[Dict[str, Any]]
         Returns:
             list[dict]: a list of backends.
         """
-        return self.client_api.backends(timeout=timeout)
+        b = self.client_api.backends(timeout=timeout)
```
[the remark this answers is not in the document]
jyu00: Oops, good catch

qiskit/providers/ibmq/api/clients/schema.py
jyu00: Yes it was moved here to resolve circular import issue, as it's used by `AccountClient`. I'll open a follow up issue to address this.

qiskit/providers/ibmq/job/ibmqjob.py
jyu00: This is to catch the off chance someone creates an `IBMQJob` manually and try to submit it.

qiskit/providers/ibmq/job/ibmqjob.py
diego-plan9: To be reevaluated during #363

PR #330, opened 2019-09-04
Resolve simpler mypy errors

luisg5:
### Summary

Adresses #265 

### Details and comments

This is the first round of resolving mypy errors, in the effort of soon making mypy a prerequisite to builds. The resolved errors in this PR are simple fixes such as:

1. Adding type hints to untyped functions.
2. Fixing previously, wrongly typed variables.
3. Ignoring functions that will soon be deprecated.

I chose to ignore functions that will soon be deprecated since we will not be adding type hints to them. Ignoring these functions also ensures they are not reported as errors when running `make mypy`.  This also makes it easier to keep track of the actual mypy errors we would like to address.

### DIFF
--- a/qiskit/providers/ibmq/ibmqbackend.py
+++ b/qiskit/providers/ibmq/ibmqbackend.py
@@ -25,6 +25,8 @@ from qiskit.qobj import Qobj
 from qiskit.providers import BaseBackend, JobStatus
 from qiskit.providers.models import (BackendStatus, BackendProperties,
                                      PulseDefaults, BackendConfiguration)
+from qiskit.providers.ibmq.accountprovider import AccountProvider
+from qiskit.providers.ibmq.ibmqprovider import IBMQProvider
 
 from .api import ApiError, IBMQConnector
 from .api_v2.clients import BaseClient, AccountClient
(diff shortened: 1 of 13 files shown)


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -43,7 +43,7 @@ class IBMQBackend(BaseBackend):
     def __init__(
             self,
             configuration: BackendConfiguration,
-            provider,
+            provider: Union['AccountProvider', 'IBMQProvider'],
```
luisg5: Temporary forward reference marker. Needs to be dealt with after addressing the circular dependency that arises when the class is imported.

qiskit/providers/ibmq/providerbackends.py
```diff
@@ -16,15 +16,17 @@
...
 
+from .ibmqbackend import IBMQBackend
 from .api_v2.exceptions import RequestsApiError
 
 
 class ProviderBackends(SimpleNamespace):
     """Backend namespace for an IBM Quantum Experience account provider."""
 
-    def __init__(self, provider):
+    def __init__(self, provider: 'AccountProvider') -> None:
```
luisg5: Temporary forward reference marker. Needs to be dealt with after addressing the circular dependency that arises when the class is imported.

PR #334, opened 2019-09-06
Include qiskit.providers.ibmq.api.rest.schemas into the build

yeralin:
### Summary
Add `qiskit.providers.ibmq.api_v2.rest.schemas` package to be part of the `qiskit-ibmq-provider` build.


diego-plan9:
Thanks @yeralin ! This will be needed, but let's hold off merging until the schemas are either verified or in use.

diego-plan9:
After the latest changes and re-evaluating, the usage of the schemas will not be included in the next release - so let's close the PR for the time being to avoid confusion, and we can re-issue it when we are fully ready to integrate them, along with the next round of updates.

PR #340, opened 2019-09-10
Resolve finer mypy errors

luisg5:
### Summary

Adresses #265 

### Details and comments

This PR is a follow up to #330. Besides adding type hints to untyped methods, there are also more specific fixes. All in all, the fixes addressed include:

1. Adding type hints to untyped methods.
2. Adding a mypy flag to allow variable redefinitions in certain contexts.
3. Add inline type hints.

### DIFF
--- a/qiskit/providers/ibmq/api_v2/clients/websocket.py
+++ b/qiskit/providers/ibmq/api_v2/clients/websocket.py
@@ -19,10 +19,12 @@ import json
 import logging
 import time
 from concurrent import futures
+from typing import Dict, Generator, Union, Optional, Any
 import warnings
 
 import nest_asyncio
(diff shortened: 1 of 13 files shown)


qiskit/providers/ibmq/api_v2/clients/websocket.py
```diff
@@ -130,7 +136,11 @@ def _connect(self, url):
         return websocket
 
     @asyncio.coroutine
-    def get_job_status(self, job_id, timeout=None):
+    def get_job_status(
+            self,
+            job_id: str,
+            timeout: Optional[int] = None
```
jyu00: The timeout value is actually `float`, not `int`. Can you also correct the docstring?

Issue #331, opened 2019-09-04
Websocket should be retried after failure


sdague:
There are many reasons why the websocket service might break a connection. Today, if the websocket for status fails for any reason, qiskit falls back to polling.

This should be avoided. The websocket should be reconnected to after any hangup.

diego-plan9:
Possible duplicate of #322 - can you check if the related issue and PR covers the issue?

sdague:
The fix listed above a single retry, which I think is the wrong fix. This should be done per a normal retry loop, try a couple of times, and if successful, reset the retry count. Falling back to polling should be discouraged.

jyu00:
It's easy enough to change the retry in the fix from a boolean to a count. However, how many time is considered "sufficient retries"? We picked retrying once because if it's just a blip then once should be enough. Do you have specific scenarios in mind that would require multiple retries?

sdague:
On the REST side the retry count is 5 - https://github.com/Qiskit/qiskit-ibmq-provider/blob/4a0c29f6f1ec4eff7a079d355d5d2cc34e7302b7/qiskit/providers/ibmq/api_v2/session.py#L48, with a delay between connects to handle network issues. That's probably a good baseline.

The other thing about websockets being long lived connections, is that a successful connection should reset the counter. From what I can tell in looking through the code, the retry flag is only set at the beginning - https://github.com/Qiskit/qiskit-ibmq-provider/blob/4a0c29f6f1ec4eff7a079d355d5d2cc34e7302b7/qiskit/providers/ibmq/api_v2/clients/websocket.py#L158 

If you make it here https://github.com/Qiskit/qiskit-ibmq-provider/blob/4a0c29f6f1ec4eff7a079d355d5d2cc34e7302b7/qiskit/providers/ibmq/api_v2/clients/websocket.py#L178 I think the retry counter should be reset.

Because the servers are running in pods, there are many reasons they might go down, including rolling upgrades. And for long running jobs that are taking an hour to return, I think seeing a couple of interuptions is not unexpected.

luisg5:
Closed via #341.

PR #341, opened 2019-09-12
Allow websocket retries with backoff time in between

luisg5:
### Summary

Adresses #331 

### Details and comments

The `WebsocketClient` class method `get_job_status` currently allows retrying once in the case of a connection closing. This PR allows the method to retry up to five times, with time in between.


qiskit/providers/ibmq/api_v2/clients/websocket.py
```diff
@@ -203,12 +224,24 @@ def get_job_status(self, job_id, timeout=None):
...
+                                       'Retrying get_job_status after %s seconds: '
+                                       'Attempt #%s.',
+                                       message, ex.code, backoff_time, current_retry_attempt)
+
+                        # Block asyncio loop for a given backoff_time.
+                        yield from asyncio.sleep(backoff_time)
+
+                        try:
+                            websocket = yield from self._connect(url)
+                        except (WebsocketError, WebsocketAuthenticationError):
```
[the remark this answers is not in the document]
luisg5: I've added that to the PR, thanks for the catch Jessie 👍

qiskit/providers/ibmq/api/clients/websocket.py
```diff
@@ -161,75 +180,119 @@ def get_job_status(
...
+                        # Successfully received a message, reset retry counter.
+                        current_retry_attempt = 0
+
+                        response = WebsocketMessage.from_bytes(response_raw)
+                        last_status = response.data
+
+                        job_status = response.data.get('status')
+                        if (job_status and
+                                ApiJobStatus(job_status) in API_JOB_FINAL_STATES):
+                            break
```
jyu00: With nested `while` and multiple `break` and `continue`, it'd probably be cleaner to just do a `return` here since the task is done.

qiskit/providers/ibmq/api/clients/websocket.py
```diff
@@ -161,75 +180,119 @@ def get_job_status(
...
+                    except ConnectionClosed as ex:
+                        # From the API:
+                        # 4001: closed due to an internal errors
+                        # 4002: closed on purpose (no more updates to send)
+                        # 4003: closed due to job not found.
+                        message = 'Unexpected error'
+                        if ex.code == 4001:
+                            message = 'Internal server error'
+                        elif ex.code == 4002:
+                            break
```
jyu00: Same as above.

qiskit/providers/ibmq/api/clients/websocket.py
```diff
@@ -161,75 +180,119 @@ def get_job_status(
...
+                            attempt_retry = False  # No point in retrying.
+                            message = 'Job id not found'
+
+                        raise WebsocketError('Connection with websocket closed '
+                                             'unexpectedly: {}(status_code={})'
+                                             .format(message, ex.code)) from ex
+
+            # Errors unrelated to websocket connection, not worth retrying.
+            except (WebsocketIBMQProtocolError, WebsocketTimeoutError) as ex:
+                raise ex
```
jyu00: You don't need to `except` the errors if you're just going to raise it? Or are you trying to document the possible exceptions?

qiskit/providers/ibmq/api/clients/websocket.py
```diff
@@ -161,75 +180,119 @@ def get_job_status(
...
+
+            # Errors related to websocket connection, attempt retrying.
+            except (WebsocketError, WebsocketAuthenticationError) as ex:
+                logger.warning('%s', ex)
+
+                current_retry_attempt = current_retry_attempt + 1
+                if current_retry_attempt > retries:
+                    break  # Attempted maximum number of retries already.
```
jyu00: It's probably worth keeping `WebsocketAuthenticationError` instead of masking it as `WebsocketError`, just to know that the error occurred during authentication. Or at least do a `from ex` so the previous error is preserved.

qiskit/providers/ibmq/api/clients/websocket.py
luisg5: I agree! I think it would be much cleaner to simply remove them, since they are going to be raised regardless.

qiskit/providers/ibmq/api/clients/websocket.py
```diff
@@ -82,13 +84,14 @@ class WebsocketClient(BaseClient):
...
     """
+    BACKOFF_MAX = 8  # Maximum time to wait between retries.
 
     def __init__(self, websocket_url: str, access_token: str) -> None:
         self.websocket_url = websocket_url.rstrip('/')
         self.access_token = access_token
 
     @asyncio.coroutine
-    def _connect(self, url: str) -> Generator[Any, None, Connect]:
+    def _connect(self, url: str) -> Generator[Any, None, WebSocketClientProtocol]:
```
jyu00: Can you also update the docstring for the return type (from `Connect` to `WebSocketClientProtocol`)?

Issue #304, opened 2019-08-12
Use backend configuration for determining object storage support


diego-plan9:
### What is the expected behavior?

Currently, when submitting a job, the provider [always attempt to use object storage](https://github.com/Qiskit/qiskit-ibmq-provider/blob/d9eba39548fddda529c5e3982f5b93d0441490a9/qiskit/providers/ibmq/job/ibmqjob.py#L458) - if the submission fails, it falls back to submitting the job via a regular POST.

The API team mentioned that a new attribute will soon be added to the backend configuration that signals unequivocally if the backend supports object storage or not - when that happens, we should remove the fallback and instead decide if submitting via object storage (or not) directly.

diego-plan9:
Closed via #343

diego-plan9:
Re-fixed by #362

PR #343, opened 2019-09-18
Refactor object storage support

luisg5:
### Summary

Addresses #304 

### Details and comments

By default, the v2 provider tries to submit a job using object storage, which may not be supported. If submitting the job via object storage fails, there is a fallback to submit the job via HTTP. This PR removes the fallback by deciding beforehand whether object storage is supported or not.


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -85,7 +85,9 @@ def run(self, qobj: Qobj, job_name: Optional[str] = None) -> IBMQJob:
         kwargs = {}
         if isinstance(self._api, BaseClient):
             # Default to using object storage and websockets for new API.
-            kwargs = {'use_object_storage': True,
+
+            use_object_storage = self._configuration['allow_object_storage']
```
luisg5: Note: `allow_object_storage` may not be the actual attribute name. It is simply a placeholder as of right now. It will be changed once the attribute is added to the backend configurations.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -81,8 +81,10 @@ def run(self, qobj: Qobj, job_name: Optional[str] = None) -> IBMQJob:
             IBMQJob: an instance derived from BaseJob
         """
         # pylint: disable=arguments-differ
-        kwargs = {'use_object_storage': True,
+        use_object_storage = self.configuration().allow_object_storage
+        kwargs = {'use_object_storage': use_object_storage,
```
diego-plan9: We can simplify things a bit after the deprecation of `v1` - instead of using the temporary variables, can you simply pass `use_object_storage` to the constructor (and just drop `use_websockets`, to make it rely on the default value)?

```
job = IBMQJob(self, None, self._api, qobj=qobj, use_object_storage=self.configuration....)
```

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -450,33 +450,26 @@ def _submit_callback(self, job_name=None):
         backend_name = self.backend().name()
 
         submit_info = None
-        if self._use_object_storage:
-            # Attempt to use object storage.
-            try:
+        try:
+            if self._use_object_storage:
+                # Submit Qobj via object storage.
```
diego-plan9: Small nit - can yo use ` Submit Job ...` instead of `Qobj`, for clarity? (also on the else)

qiskit/providers/ibmq/ibmqbackend.py
diego-plan9: Seems it is the final name!

qiskit/providers/ibmq/ibmqbackend.py
luisg5: Yeah, simplifying would be a good idea, will do! 👍

Issue #267, opened 2019-07-18
Revise usages of "IBMQConnector or BaseClient"


diego-plan9:
### What is the expected behavior?

During the phase were both v2 and v1 were supported, some of the classes were modified in order to support both kinds of object that connects to the API. When v1 reaches EOL, we should simplify them accordingly:

* [ ] `IBMQBackend`
  * [ ] constructor + `self._api`
  * [ ] `run()` - `isinstance`
  * [ ] `jobs()` - `isinstance`
  * [ ] `retrieve_job()` - `isinstance`
* [ ] `IBMQJob`
  * [ ] constructor + `self._api`
  * [ ] revise `self._use_websockets` and `use_websockets`

PR #346, opened 2019-09-20
Revise api client usage

luisg5:
### Summary

Adresses #267 

### Details and comments

With #345, the `IBMQConnector` class is deprecated. As a result, the `_api` member variables of `IBMQBackend` and `IBMQJob` could only be of the type `AccountClient`


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -113,12 +109,6 @@ def properties(
         """
         # pylint: disable=arguments-differ
         if datetime:
-            if not isinstance(self._api, BaseClient):
-                warnings.warn('Retrieving the properties of a '
-                              'backend in a specific datetime is '
-                              'only available when using IBM Q v2')
-                return None
```
diego-plan9: Good catch!

PR #354, opened 2019-09-30
Changes to provider.backends

diego-plan9:
### Summary

Update `provider.backends` according to the recent design document:
* use `ProviderBackends` (now `IBMQBackendService`) as `provider.backends()`
* move `.jobs()` and `.retrieve_job()` to `provider.backends`


qiskit/providers/ibmq/ibmqbackendservice.py
```diff
@@ -0,0 +1,296 @@
...
+                job_kind = ApiJobKind(job_info.get('kind', None))
+            except ValueError:
+                # Discard pre-qobj jobs.
+                break
+
+            kwargs['use_websockets'] = True
+            if job_kind == ApiJobKind.QOBJECT_STORAGE:
+                kwargs['use_object_storage'] = True
+
+            job = IBMQJob(self, job_info.get('id'), self._provider._api,
```
jyu00: `IBMQjob` constructor takes a `BaseBackend` as its first parameter. With this change I'm guessing it'll be updated to take a backend name instead of a backend instance? And `job.backend()` will just return the name string.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -221,83 +218,12 @@ def jobs(
...
-                kwargs['use_object_storage'] = True
-
-            job = IBMQJob(self, job_info.get('id'), self._api,
-                          creation_date=job_info.get('creationDate'),
-                          api_status=job_info.get('status'),
-                          **kwargs)
-            job_list.append(job)
-
-        return job_list
+        warnings.warn('backend.jobs() is deprecated and will be removed after '
```
jyu00: If this is being deprecated then the test cases should be updated to use the new methods.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -221,83 +218,12 @@ def jobs(
...
+                      '0.5. Please use provider.backends.jobs(backend_name='
+                      '"{}") instead.'.format(self.name()), DeprecationWarning)
+
+        return self._provider.backends.jobs(
```
jyu00: If the intent of this change is just to allow job retrieval without a backend, then `provider.jobs()` seem to make more sense than `provider.backends.jobs()`.

qiskit/providers/ibmq/ibmqbackendservice.py
diego-plan9: Correct - this goes along the lines of a comment by @luisg5 too, and the first parameter indeed needs to be fixed! It will likely spawn a follow-up to be tackled at `job.backend()` - as in the original document, it was not contemplated what should happen for jobs whose backend does not exist anymore.

qiskit/providers/ibmq/ibmqbackend.py
diego-plan9: Unfortunately `provider.jobs()` is explicitly not defined in the proposal, right?

qiskit/providers/ibmq/ibmqbackendservice.py
diego-plan9: In the end decided to leave it empty, and tackle it in a follow-up 612e512 when we have a better picture (ie. after refactoring).

qiskit/providers/ibmq/ibmqbackend.py
diego-plan9: Correct! 
cbfd400

qiskit/providers/ibmq/ibmqbackend.py
nonhermitian: There will be more than one job type going forward depending on the service. This is why this is the way it is at service level

Issue #129, opened 2019-05-30
Run slow_tests periodically


diego-plan9:
### What is the expected behavior?

As a spin-off from  #69, a number of tests in the project are `slow_test` that rely on individual developers running them manually. We would benefit from running them periodically in the CI to ensure they are being executed - might involve a round of cleaning them up in the process.

delapuente:
Although #136 is merged. Let's leave this open until we are able to pass them all at least once. Also:

- [ ] Investigate how to keep Travis running for more than 10 min at the same time we have the output of `make test` availabe.
- [ ] In the spirit of "master is failing" in Qiskit, add the infrastructure to inform that slow tests are failing without activating it.

diego-plan9:
On a minor note - it seems that, when a cron job is run, it "overwrites" the last `master` job in some places - most notably, in the main [commits](https://github.com/Qiskit/qiskit-ibmq-provider/commits/master) tab. 

For example, at the time of writing, the red x in the latest commit (a8e673eecb9154f8af1a3ec00ff3402a6ab96480) points to:
https://github.com/Qiskit/qiskit-ibmq-provider/runs/143526463 (the one from the cron job)
as opposed to:
https://travis-ci.com/Qiskit/qiskit-ibmq-provider/builds/114630308 (the one from the commit)
 
This is probably a Travis oddity, but can we look into it? Mostly as the badge we are using in the `README` and in other projects is linked to "the latest master build", and also for easiness of tracking real master failures and inspection.

diego-plan9:
Updated the cron job to make it run weekly while solving the issue mentioned on the previous comment.

delapuente:
According to my research, what happends is that the links from the commits to the Travis builds are intended to point to the most recent build for that commit. I'm looking into this specifically. A possible workaround is to make the cron jobs never to fail so it would not show like a failure in the `REDME` shield but that would also mask the failures in the commits.

jyu00:
Slow tests now run weekly.

PR #358, opened 2019-10-01
Run some slow tests on staging

jyu00:
### Summary
Partially addresses #129.


jyu00:
Closing due to wrong credentials used.

PR #367, opened 2019-10-03
Refactor websocket message class

luisg5:
### Summary

Related to #265 and #361.

### Details and comments

The `WebsocketMessage` class has conflicting types for its `data` class variable. As a result, its use within `WebsocketClient` is not always clear or consistent. These were issues raised by mypy and rather than ignoring them, a refactor was done instead. On top of resolving `mypy` issues, a refactor ensures the inconsistencies are dealt with and clarifies its use.

### DIFF
--- a/qiskit/providers/ibmq/api/clients/websocket.py
+++ b/qiskit/providers/ibmq/api/clients/websocket.py
@@ -54,7 +54,7 @@ class WebsocketMessage:
     def __init__(
             self,
             type_: str,
-            data: Optional[Union[str, Dict[str, str]]] = None
+            data: Optional[Dict[str, str]] = None
     ) -> None:
         self.type_ = type_
         self.data = data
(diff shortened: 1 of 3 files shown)


jyu00:
I actually like how it was done before, that `WebsocketMessage` is simply a container of messages sent/received via websockets. The protocol with the api server does not impose a restriction on the type of data being passed. I don't think it's necessary to artificially convert the data just to make `mypy` happy. @diego-plan9 what do you think?

diego-plan9:
Actually you are both correct - Luis in the spirit of the PR, and Jessie in the sense that, indeed, we should not just make changes solely to make `mypy` happy.

In this case, it seems to come from a potential bug discovered thanks to `mypy`:
```
.../websocket.py:224: error: Item "str" of "Union[str, Dict[str, str], None]" has no attribute "get"
```

Which boils down to giving us a warning that [this `response.data.get('status')`](
https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/api/clients/websocket.py#L224) might fail ungracefully for messages that do not contain a dict. If that is the case, and considering that we have quite some guarantees over what to expect from the websocket server, I'd say tuning up the `WebsocketMessage` is reasonable (albeit note I did not really review!).

diego-plan9:
After a more detailed review, can we try another approach?

Since it seems the authentication message is the only response that will _not_ contain a dict in `data`, let's try subclassing and having a different class for `WebsocketUsedForAuthenticationMessage` and `WebsocketUserForAllTheOtherMessage`.

This way we can restrict the parameters and handling for each one (authentication would expect `str`, and for the rest a `dict`), and at the same time ensure we get strict checks for the types. Actually, can you check if we can drop the `Optional`, mostly to avoid unexpected crashes during `response.data.get('status')` if it is `None`?

Issue #176, opened 2019-06-19
Make error_message() return the message


ryan-woo:
### What is the expected behavior?
The error message will begin to come up from backends soon. IBMQ Provider needs to accommodate them.

The error message will be found in the `qObjectResult` in json as:

``` 
"qObjectResult": {
    <normal qObjectResult fields>
    "error": {
        "message": "Error message example",
        "code": 9999
    }
}
```

The error field will only appear if there is an error, so the job status should still be checked before attempting to access the `"error"`.

This should be a simple enough matter of just digging a bit deeper into the job_response.

nonhermitian:
A bit confused here. The error message, if any, should be returned in the status check so that one does not need to retrieve the results In order to get it.

ryan-woo:
Yes the `error` object can also be found in the "job" (the object that holds the `qObjectResult`). I think what you are asking for, where the status check returns the message, @pacomf should be able to help with.

Right now it looks like `error_message` is written in a way that it wants the full job result. That of course can change as well so that the error message comes from the status, but the API level needs to implement it.

jyu00:
I think what @nonhermitian is asking is to have `job.status()` return the error message if an error has occurred. This would require api changes since the provider gets the status from the api. 

In the meantime, we can extract and return the error message in `job.result()` and `job.error_message()`. So people who use `job.result()` can see more than just `Invalid job state`.

jyu00:
Put it on hold until backend error message code is deployed.

nonhermitian:
This should be taken off hold as it is implemented on some devices now.

jyu00:
I think the provider's work on this is done for now. Just to do a quick recap, `job.error_message()` will retrieve the job result, and either
1. produce an error report if individual error message is available for each experiment 
2. give a one-line error message if individual results are not available

Some parts of the stack might not be producing error messages right now, but when they do so in the correct format, the provider will automatically pick them up.

PR #370, opened 2019-10-04
Job error message

jyu00:
### Summary
Addresses #176. This PR focuses on extracting `error` from `qObjectResult` when partial results are not returned. The latter will be revisited when partial result is schema compliant. 

### Details and comments
`job.error_message()` will now return
```
Instruction not in basis gates: instruction: bad_instruction, qubits: [0, 1], params: []
```
or 
```
The following experiments failed:
Experiment 1: ERROR: Circuit contains invalid instructions ( invalid gate instructions: {bad_instruction})
```
if individual experiment status is known.


quantumjim:
Does this give a nice error message in the case that `shots` is too high?

jyu00:
> Does this give a nice error message in the case that `shots` is too high?

Right now it just gives `ERROR_VALIDATING_JOB. However, the API team is defining more granular error message, so that should be coming soon!

PR #381, opened 2019-10-09
Update job.properties() docstring

jyu00:
### Summary
As discussed in #116, API now returns backend properties at job creation time if the job hasn't finished. The existing docstring is misleading. The PR left the dostring ambiguous since the behavior is likely to change again.

### DIFF
--- a/qiskit/providers/ibmq/job/ibmqjob.py
+++ b/qiskit/providers/ibmq/job/ibmqjob.py
@@ -160,12 +160,8 @@ class IBMQJob(BaseModel, BaseJob):
     def properties(self) -> Optional[BackendProperties]:
         """Return the backend properties for this job.
 
-        The properties might not be available if the job hasn't completed,
-        in which case None is returned.
-
         Returns:
-            BackendProperties: the backend properties used for this job, or None if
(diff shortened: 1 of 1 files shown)


qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -160,23 +160,15 @@ def qobj(self) -> Qobj:
...
 
         Raises:
             JobError: if there was some unexpected failure in the server.
         """
         with api_to_job_error():
             properties = self._api.job_properties(job_id=self.job_id())
 
-        # Backend properties of a job might not be available if the job hasn't
-        # completed. This is to ensure the properties returned are up to date.
-        if not properties:
```
diego-plan9: Can you double-check if we can really remove this check? In particular:
* for jobs against a simulator, properties is probably empty
* for older jobs, can we be sure that 100% properties will be available (ie. edge cases such as errored jobs or some odd combinations)

jyu00: You were right. Forgot about the simulator. Put the check back.

Issue #333, opened 2019-09-05
A workflow engine


ajavadia:
IBMQ needs a workflow engine to manage the execution and retrieval of jobs.

This would be a higher level construct than `Job`, which is the unit of interaction with the backends (from the API's point of view). But jobs have arbitrary limitations that should be masked from the high-level user. For example each backend places its own limit on the number of experiments contained within a job. So currently you have to manually batch many circuits into some jobs, manually store the jobs into some datastructure, and manually retrieve/post-process the results.  If one job fails, it can be difficult to recover the total results without manually resubmitting that job and replacing the entry in the database. 

The workflow engine automates this process. It can be called a `JobManager`. At a high level, the user specifies which "tasks" they would like to run. Even if a task has a million circuits, the workflow engine will take care of splitting them, farming them out to the backends, reporting on the progress to the user, and joining the reuslts into a single Result for the task.

Simple pseudo-code (up for discussion):
```python
from qiskit.providers.IBMQ import JobManager
jm = JobManager()

# build a million circuits
from qiskit.circuit.random import random_circuit
circs = []
for _ in range(1000000):
    circs.append(random_circuit())

# farm out the jobs.. 
from qiskit import IBMQ
p = IBMQ.load_account()
device = p.backends.tenerife()
jm.run(circs, backend=device)

# inquire about the status of my runs
jm.status()  # says how many jobs, and the status of each

# join result of (potentially) multiple jobs
res = jm.result()
```

woodsp-ibm:
This would be great if we can alleviate Aqua from these type of transport/payload concerns and have it more widely available. Having a reliable guaranteed operation would allow us to rely on such an engine instead of requiring to have such code in Aqua. I am cross-referencing this PR as more thought for the engine. While we did some level splitting it turns out that it can become less than optimal as the PR outlines. https://github.com/Qiskit/qiskit-aqua/pull/694

ajavadia:
Long term it would be nice to be able to somehow record this in the online database too. One very nice feature of IBMQProvider is that it allows you to retrieve results/qobj/properties from past executions and analyze them. If, for some big experiment, your circuits had to be split into 5 jobs, it can be hard to relate those 5 jobs later on. So it would be nice if the `JobManager` somehow recorded this relationship between the jobs in the database and presented them to the user in one unified form upon later retrieval.

jyu00:
@ajavadia a few design questions:

[...]

PR #389, opened 2019-10-11
Job manager

jyu00:
### Summary
Addresses #333 .

### Details and comments
A job manager that takes a list of circuits/schedules, splits them into multiple jobs and submits the jobs. A few notes:
- `run()` only takes a list of circuits/schedules, not a single one. The rationale is the manager is meant to manage a large number of jobs.
- `run()` takes `IBMQBackend`, not `BaseBackend` as the `backend` input. Local backends don't have circuit limits and thus shouldn't need a job manager.
- This does not split the circuits by max gates, since the "payload too large" error should go away once object storage is deployed everywhere (soon).


qiskit/providers/ibmq/jobmanager.py
```diff
@@ -0,0 +1,332 @@
...
+                If the specified value is greater the maximum allowed by the
+                backend, the default is used.
+
+        Returns:
+            Number of jobs to be run.
+
+        Raises:
+            JobError: If a job cannot be submitted.
+        """
+        if (all(isinstance(exp, ScheduleComponent) for exp in experiments) and
```
diego-plan9: Double-checking - `all` or `any`?

qiskit/providers/ibmq/jobmanager.py
```diff
@@ -0,0 +1,332 @@
...
+                cannot be retrieved due to server error.
+        """
+        statuses = []
+        for i, job in enumerate(self._jobs):
+            try:
+                statuses.append(job.status())
+            except JobError as err:
+                logger.warning("Unable to retrieve status for job %d (job ID=%s): %s",
+                               i, job.job_id(), str(err))
+                statuses.append(None)
```
diego-plan9: Would this cause some ripples in `report()`? It seems it would not account for the job in the non-detailed output (which can be considered "by design"), but more importantly can you check it does not cause an error in the function (ie. when accessing `status.value` or similar cases)?

test/ibmq/test_jobmanager.py
```diff
@@ -0,0 +1,216 @@
...
+        circs = [transpile(self._qc, backend=backend), bad_qc]
+        self._jm.run(circs, backend=backend, max_experiments_per_job=1, skip_transpile=True)
+
+        results = self._jm.result()
+        self.assertIsNone(results[1])
+
+        error_report = self._jm.error_message()
+        self.assertIsNotNone(error_report)
+        print("JobManager.error_message(): \n{}".format(error_report))
+        print("\nJobManager.report(): \n{}".format(self._jm.report()))
```
diego-plan9: Can you do a pass for the `print`s, which seem to be leftovers?

qiskit/providers/ibmq/jobmanager.py
jyu00: `assemble` will fail if they're not all the same type, but `any` is probably more efficient here.

qiskit/providers/ibmq/jobmanager.py
jyu00: Not counting in the summary report is by design; fixed `status.value`. We really should have a `JobStatus.UNKNOWN`.

test/ibmq/test_jobmanager.py
jyu00: They were intentional to allow someone to "eyeball" them in the Travis log. But perhaps they just pollute the log. I'll remove them.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -23,9 +23,8 @@
 
 from marshmallow import ValidationError
 
-from qiskit.providers.basejob import BaseJob
-from qiskit.providers.basebackend import BaseBackend
-from qiskit.providers.exceptions import JobTimeoutError
+from qiskit.providers import (BaseJob, # type: ignore[attr-defined]
+                              JobTimeoutError, BaseBackend)
```
diego-plan9: Note I reverted this change from `master` during the conflict solving at the latest merge, as technically the `qiskit.providers` API is the one that is stable - we don't have that guarantee for `qiskit.providers.xxx.yyy` (at the expense of the mypy ignore).

PR #393, opened 2019-10-13
no breaking line in UPDATE_ACCOUNT_TEXT

1ucian0:
There is a `\n` in `UPDATE_ACCOUNT_TEXT`, which results in exception messages like this one:
```
qiskit.providers.ibmq.exceptions.IBMQApiUrlError: 'Invalid IBM Q Experience credentials found. Please update your accounts and programs by following the instructions here:\nhttps://github.com/Qiskit/qiskit-ibmq-provider#updating-to-the-new-ibm-q-experience'
```
Notice that `\n` shows up in the message, making the url unclickable in those terminals that support it.

This PR removes that `\n`. Now it looks like this one:

```
qiskit.providers.ibmq.exceptions.IBMQApiUrlError: 'Invalid IBM Q Experience credentials found. Please update your accounts and programs by following the instructions here: https://github.com/Qiskit/qiskit-ibmq-provider#updating-to-the-new-ibm-q-experience'
```

### DIFF
--- a/qiskit/providers/ibmq/ibmqfactory.py
+++ b/qiskit/providers/ibmq/ibmqfactory.py
@@ -30,9 +30,9 @@ from .exceptions import IBMQAccountError, IBMQApiUrlError, IBMQProviderError
 logger = logging.getLogger(__name__)
 
 QX_AUTH_URL = 'https://auth.quantum-computing.ibm.com/api'
-UPDATE_ACCOUNT_TEXT = (
-    "Please update your accounts and programs by following the instructions here:",
-    "https://github.com/Qiskit/qiskit-ibmq-provider#updating-to-the-new-ibm-q-experience")
+UPDATE_ACCOUNT_TEXT = "Please update your accounts and programs by following the " \
+                      "instructions here: https://github.com/Qiskit/qiskit-ibmq-provider#" \
+                      "updating-to-the-new-ibm-q-experience "
 
 
 class IBMQFactory:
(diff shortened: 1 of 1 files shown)


diego-plan9:
Thanks @1ucian0, nice catch !

Fixing the spurious `\n` is great, but can you revert the exception instantiation in order to pass a single string as their only argument, instead of multiple strings? The parenthesis when defining `UPDATE_ACCOUNT_TEXT` were only for being able to split the lines in two (ie. only for formatting), and even if the original `QiskitError` accepts tuples and concatenates them, we lean towards just using a single argument, which would help a bit #106 by not restricting the parameter usage too much.

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -28,13 +27,12 @@
...
 
-
 logger = logging.getLogger(__name__)
 
 QX_AUTH_URL = 'https://auth.quantum-computing.ibm.com/api'
 UPDATE_ACCOUNT_TEXT = (
-    'Please update your accounts and programs by following the instructions here:\n'
-    'https://github.com/Qiskit/qiskit-ibmq-provider#updating-to-the-new-ibm-q-experience')
+    "Please update your accounts and programs by following the instructions here:",
+    "https://github.com/Qiskit/qiskit-ibmq-provider#updating-to-the-new-ibm-q-experience")
```
diego-plan9: Small note - with the changes mentioned on the main comment, probably a ` ` needs to be introduced before the url.

1ucian0:
done

PR #395, opened 2019-10-15
Refactor RequestsApiError Exception

luisg5:
### Summary

Addresses #106 

### Details and comments

This PR is a step into unifying the exceptions within the provider. It removes an unnecessary initializer from `RequestsApiError` by using exception chaining to accomplish the same behavior.

### DIFF
--- a/qiskit/providers/ibmq/api/clients/auth.py
+++ b/qiskit/providers/ibmq/api/clients/auth.py
@@ -14,6 +14,7 @@
 
 """Client for accessing authentication features of IBM Q Experience."""
 from typing import Dict, List, Optional, Any
+from requests.exceptions import RequestException
 
 from ..exceptions import AuthenticationLicenseError, RequestsApiError
 from ..rest import Api, Auth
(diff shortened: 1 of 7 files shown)


qiskit/providers/ibmq/api/clients/auth.py
```diff
@@ -75,12 +75,12 @@ def _request_access_token(self) -> str:
             response = self.client_auth.login(self.api_token)
             return response['id']
         except RequestsApiError as ex:
-            response_obj = ex.original_exception.response
-            if response_obj is not None and response_obj.status_code == 401:
+            error_response = ex.__cause__.response    # pylint: disable=no-member
+            if error_response is not None and error_response.status_code == 401:
```
diego-plan9: Can you revise a bit the variable and the `if`?

Previously, we had strong assurances that `ex.original_exception` was coming from `requests` - now we still have them, although we are open to the more general case of the exception being `raise from` somewhere else (although in practice it doesn't seem it will happen yet). Can we do some defensive programming and try something along the lines of:
```
if isinstance(ex.__cause__, Blah) and ...
```

Issue #49, opened 2019-03-04
Handle partial success in jobs


delapuente:
### What is the expected behavior?
A job can partially succeed if not all their experiments fail. The current implementation does not allow the user to retrieve the successful experiments. Instead, it raises an error when trying to get the results.

```
>> job.status()
<JobStatus.ERROR: 'job incurred error'>
>> job.result()
JobError: 'Invalid job state. The job should be DONE but it is JobStatus.ERROR'
```

We should be able to browse a partial success, even if the global status of the job is `ERROR`.

jyu00:
This does seem to be a useful feature, but it would only apply to the simulators since real backends don't return partial results. The main issue is that the results returned by the simulators are not schema compliant if some of the experiments had failed. For example, a partially successful job result might look like 

```
 {'backend_name': 'ibmq_qasm_simulator',
 ...
 'results': [{'status': 'ERROR: Circuit contains invalid instructions ( '
                        'invalid gate instructions: {test_name})',
              'success': False},
             {'data': {'counts': {'0x0': 122, '0x3': 134}},
              'header': {...},
              'shots': 256,
              'status': 'DONE',
              'success': True],
 'status': 'COMPLETED',
 'success': False,
 'time_taken': 0.003798207}
```

The first experiment in the above example is missing required fields `data` and `shots`, therefore we cannot easily convert it to a `qiskit.Result` object.  If we want to implement this, we can either 1) manipulate the data to make it schema compliant or 2) ask the simulator teams to return schema compliant results for failed experiments or 3) loosen the schema to not require `data` and `shots`. 

@diego-plan9 Any thoughts?

jyu00:
The test case that generates an error can be found [here](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/test/ibmq/test_ibmq_job.py#L375)

diego-plan9:
@luisg5 , @jyu00 - can you check if we can close this issue?

luisg5:
The issue should not be closed yet. Although the qasm simulator from the latest Aer release supports partial results, the results returned from the API are still not schema complaint.

Since this is the case, running a job (where some experiments succeed and others fail) will currently cause an `IBMQJobFailure` exception.

diego-plan9:
Ok - changing the milestone accordingly, so we can retake the issue after the fixes are in place in the remote components.

luisg5:
[...]

PR #399, opened 2019-10-17
Return partial results

luisg5:
### Summary

Addresses #49 

### Details and comments

Placed `on hold` because the added test currently fails. The issue is that the result returned by `job_result` is not schema compliant. Will remove the label when this changes.


jyu00:
A few comments:
- Simulator is the only one that returns partial results today, but that's going to change. Instead of checking for partial results only for simulators, it'd be better to look for them if `partial=True`. So when it's implemented by non-simulators, people with down level qiskit can get them immediately, and we won't have to update the code to remove the check.
- While simulators normally support partial results, it's still better to a) verify results are returned (I've seen cases where they are not, not sure why) and b) catch and suppress any schema validation error raised by `from_dict()`, just to be nicer to the user
- Some results are "consumed" after the first call. So if a user wants to get both partial results and error messages, today it'd be 2 calls to the results endpoint. It needs to be cached somewhere.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -139,7 +139,9 @@ def __init__(self,
         self._update_status_position(_status, kwargs.pop('infoQueue', None))
 
         # Properties used for caching.
+        self._result = None
```
jyu00: `self._result` could have been set on line 130 (see `job/schema`). You don't want to override it here.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -226,6 +240,26 @@ def result(self, timeout: Optional[float] = None, wait: float = 5) -> Result:
...
+        """
+        if not self._result and not self._partial_result:  # type: ignore[misc]
+            # No results, nor partial results, have been attained for this job yet.
+            with api_to_job_error():
+                # Set partial results, in case validation fails they are cached.
+                self._partial_result = self._api.job_result(self.job_id(), self._use_object_storage)
+
+            try:
+                return Result.from_dict(self._partial_result)
+            except ValidationError:
```
jyu00: Did you verify if it really throws a `ValidationError`? I `from_dict()` converts that to a different error.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -226,6 +240,26 @@ def result(self, timeout: Optional[float] = None, wait: float = 5) -> Result:
...
+        """Attempt to retrieve partial results for a job.
+
+        Returns:
+            The partial results for a job, or ``None`` if schema validation fails.
+        """
+        if not self._result and not self._partial_result:  # type: ignore[misc]
+            # No results, nor partial results, have been attained for this job yet.
+            with api_to_job_error():
+                # Set partial results, in case validation fails they are cached.
+                self._partial_result = self._api.job_result(self.job_id(), self._use_object_storage)
```
jyu00: I think `result_response` is a better name for this, since the data returned by `job_result()` may or may not contain partial results. It's also a bit confusing about the differences between `_partial_result` and `_result` (since `_result` could also be partial).

qiskit/providers/ibmq/job/ibmqjob.py
luisg5: Thanks for that catch! It actually returns a `ModelValidationError` which is one of our custom exceptions. I'll update that.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -203,13 +203,27 @@ def result(
             results again in another instance or session might fail due to the
             job having been consumed.
 
+            In the case a job fails, but partial results exist, the result
+            method may return a `Result` object. If this is the case, precaution
```
jyu00: Can you clarify that partial results are only returned if `partial=True` is specified? And that one can verify whether some experiments failed by looking at the `result.success` attribute?

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -203,13 +203,27 @@ def result(
...
+            should be taken when accessing individual experiments, as doing so
+            would cause an exception.
+
+            For example:
+                If there is a job with two experiments (where one fails), getting
+                the counts of the unsuccessful experiment would raise an exception
+                since there are no counts to return for it,
+                i.e.
+                    > result = job.result()
+                    > counts = result.get_counts(failed_experiment) # Fails
```
jyu00: Can you change the second part of the example to something like
```
try:
  counts = result.get_counts("failed_experiment")
except QiskitError:
  print("Experiment failed!")
```

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -451,3 +482,27 @@ def _wait_for_completion(
             self.refresh()
 
         return self._status in required_status
+
+    def _retrieve_partial_result(self) -> Optional[Result]:
```
jyu00: I think you can simplify the code a bit by doing something like 
```
def _retrieve_result():
  if not self._result_response:
    with api_to_job_error():
      self._result_response = self._api.job_result(...)
    try:
      self._result = Result.from_dict(...)
    except ...
  return self._result
```
Then in `result()`:
```
if self._status is JobStatus.ERROR and not partial:
  raise
return self._retrieve_result()
```
and `error_message()`:
```
self._retrieve_result()
if self._result_response and self._result_response['results']: ...
```

test/ibmq/test_ibmq_job.py
```diff
@@ -386,6 +386,38 @@ def test_retrieve_failed_job_device(self, provider):
...
+
+        job = backend.run(qobj)
+        result = job.result(partial=True)
+
+        circuit_results = result['results']
+        self.assertTrue(circuit_results[0]['success'])
+        self.assertFalse(circuit_results[1]['success'])
+
+    @requires_provider
+    def test_retrieve_failed_job_device_partial(self, provider):
```
jyu00: - if this is meant to run on a real backend, it needs to say `@slow_test` or `run_on_staging`
- docstring needs to be updated 
- real backend doesn't return partial results, so `result['results']` will give an empty list.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -203,18 +203,22 @@ def result(
...
             job having been consumed.
 
-            In the case a job fails, but partial results exist, the result
-            method may return a `Result` object. If this is the case, precaution
-            should be taken when accessing individual experiments, as doing so
-            would cause an exception.
+
+            When `partials=True`, the result method may return a `Result` object
+            containing partial results. If partial results are returned, precaution
+            should be taken when accessing individual experiments, as doing so would
```
jyu00: I think you meant to say "as doing so _might_ cause an exception.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -451,3 +481,23 @@ def _wait_for_completion(
...
+
+        Raises:
+            IBMQJobApiError: if there was some unexpected failure in the server.
+        """
+        if not self._result_response:
+            with api_to_job_error():
+                self._result_response = self._api.job_result(self.job_id(),
+                                                             self._use_object_storage)
+            try:
+                return Result.from_dict(self._result_response)
```
jyu00: If you set `self._result` here instead of returning it, then you can consolidate lines 249-256 to just `return self._retrieve_result()`. It would also keep the api call to retrieve result to a single method, making the code cleaner.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -197,12 +202,31 @@ def result(self, timeout: Optional[float] = None, wait: float = 5) -> Result:
             results again in another instance or session might fail due to the
             job having been consumed.
 
+
+            When `partials=True`, the result method may return a `Result` object
```
diego-plan9: Small tweaks:
* probably `partial=True` instead of `partials`?
* can you have just one blank line at the top? 
* also, can you double-check if after the latest changes (exception instead of result) this sentence continues being accurate - specifically the "may" (it probably is fine - just wondering)?

> When `partials=True`, the result method **may** return a `Result` object

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -197,12 +202,31 @@ def result(self, timeout: Optional[float] = None, wait: float = 5) -> Result:
...
+
         Args:
            timeout: number of seconds to wait for job
            wait: time between queries to IBM Q server
+           partial: if true attempts to return partial results for the job.
 
         Returns:
-            Result object
+            Result object if there are valid results to return (including partial
+            results), else `None`.
```
diego-plan9: Also, can you double-check the last bit?

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -182,7 +182,12 @@ def properties(self) -> Optional[BackendProperties]:
 
         return BackendProperties.from_dict(properties)
 
-    def result(self, timeout: Optional[float] = None, wait: float = 5) -> Result:
+    def result(
+            self,
+            timeout: Optional[float] = None,
+            wait: float = 5,
+            partial: Optional[bool] = False
+    ) -> Optional[Result]:
```
jyu00: This should no longer be `Optional` as it'd either return `Result` or raise.

PR #400, opened 2019-10-18
Managed job sets

jyu00:
### Summary
Addresses #333 . 
Allows a job manager to manage multiple sets of experiments.

### Details and comments
This PR introduces 2 new classes - `ManagedJobSet`, which represents a set of managed jobs, and `ManagedJob`, which is simply `IBMQJob` with some management metadata.

Most of the functionalities provided by `JobManager` (now `IBMQJobManager`), were moved to `ManagedJobSet`, so a user can do something like
```
jm = IBMQJobManager()
job_set1 = jm.run(some_experiments)
job_set2  = jm.run(more_experiments)
job_set1.results()
job_set2.results()
```
Due to the number of files now associated with job manager, they're all moved to the `/managed` directory be more organized.


qiskit/providers/ibmq/managed/ibmqjobmanager.py
```diff
@@ -0,0 +1,196 @@
...
+                backend, the default is used.
+
+        Returns:
+            Managed job set.
+
+        Raises:
+            IBMQJobManagerInvalidStateError: If the backend does not support
+                the experiment type.
+        """
+        if (any(isinstance(exp, ScheduleComponent) for exp in experiments) and
```
ajavadia: I think the check here should be for `Schedule` and not `ScheduleComponent`, since the type hint says that the only acceptable inputs here are `QuantumCircuit` or `Schedule` -- so anything other than a `Schedule` should be caught by the hint. Here we are just trying to say that `Schedule` is supported by certain backends.

qiskit/providers/ibmq/managed/ibmqjobmanager.py
```diff
@@ -0,0 +1,196 @@
...
+        """Creates a new IBMQJobManager instance."""
+        self._job_sets = []  # type: List[ManagedJobSet]
+        self._executor = futures.ThreadPoolExecutor()
+
+    def run(
+            self,
+            experiments: Union[List[QuantumCircuit], List[Schedule]],
+            backend: IBMQBackend,
+            name: Optional[str] = None,
+            shots: int = 1024,
```
ajavadia: The `JobManager.run()` method has been limited here in the options that it can accept. Is this intentional? Since these will be directly passed to the `qiskit.compiler.assemble()` function, why not allow all options that are allowed in assemble?

qiskit/providers/ibmq/managed/ibmqjobmanager.py
```diff
@@ -0,0 +1,196 @@
...
+        """Return a report on the statuses of all jobs managed by this manager.
+
+        Args:
+            detailed: True if a detailed report is be returned. False
+                if a summary report is to be returned.
+
+        Returns:
+            A report on job statuses.
+        """
+        job_set_statues = [job_set.statuses() for job_set in self._job_sets]
```
ajavadia:
```suggestion
        job_set_statuses = [job_set.statuses() for job_set in self._job_sets]
```

qiskit/providers/ibmq/managed/ibmqjobmanager.py
```diff
@@ -0,0 +1,196 @@
...
+        if detailed:
+            report.append("\nDetail report:")
+            for i, job_set in enumerate(self._job_sets):
+                report.append(("  Job set {}:".format(job_set.name())))
+                report.extend(format_job_details(
+                    job_set_statues[i], job_set.managed_jobs()))
+
+        return '\n'.join(report)
+
+    def job_set(self, name: str) -> Optional[ManagedJobSet]:
```
ajavadia: is there no method to return all job sets? I think it should be easy to run a bunch of circuits and get then see all the jobs that were created. In your previous PR you could do:
```
jobmanager.jobs()
```
here it seems you should be able to do:
```
jobmanager.job_sets(0).jobs()
```

qiskit/providers/ibmq/managed/ibmqjobmanager.py
```diff
@@ -0,0 +1,196 @@
...
+        for job_set in self._job_sets:
+            if job_set.name() == name:
+                return job_set
+        return None
+
+    @classmethod
+    def previous_job_set(
+            cls,
+            name: str,
+            provider: AccountProvider,
```
ajavadia: Unrelated to this PR, but I'm still a bit confused by Account vs. Provider. What is `AccountProvider`? Is this what previously was `SingleProvider`? I thought Account was supposed to be all of the providers.

qiskit/providers/ibmq/managed/ibmqjobmanager.py
```diff
@@ -0,0 +1,196 @@
...
+        Returns:
+            A set of managed jobs or ``None`` if no named set found.
+        """
+        for job_set in self._job_sets:
+            if job_set.name() == name:
+                return job_set
+        return None
+
+    @classmethod
+    def previous_job_set(
```
ajavadia: I think this method should be called `retrieve_job_set`.

But I'm confused by its existence here. Can't you make it a method of `IBMQBackend` so that it functions similar to `backend.jobs()` and `backend.retrieve_job()`? Because those are the methods that users use to retrieve previous jobs, and I think job sets should not be different.

qiskit/providers/ibmq/managed/ibmqjobmanager.py
jyu00: The intent was to make the interface simpler. But you're right, people would want the `assemble` options. I added `**assemble_config` instead of listing them individually to avoid duplicating the options. However, this comes with the slight inconvenience to the user as they'd have to look up the `assemble()` options separately.

qiskit/providers/ibmq/managed/ibmqjobmanager.py
jyu00: An `AccountProvider` controls access by hub/group/project. An account can have multiple providers (e.g. if you belong in multiple projects).

qiskit/providers/ibmq/managed/ibmqjobmanager.py
jyu00: It's here because job sets are managed by job manager, so it seemed logical. But I agree `backend.retrieve_job_set()` makes more sense, especially since it disassociates the job set from an existing job manager instance. 

We'll probably save this feature for the next release though, since job names need not be unique and therefore is not a very good filter to find the right job set.

qiskit/providers/ibmq/managed/ibmqjobmanager.py
ajavadia: The assemble function calls these `run_config`. Can we mirror that?
https://github.com/Qiskit/qiskit-terra/blob/812e213a012859693a3432becc377ace41e9034e/qiskit/assembler/assemble_circuits.py#L21

Also it might be a good idea to add them to the docstring here for reference.

qiskit/providers/ibmq/managed/ibmqjobmanager.py
ajavadia: So one "AccountProvider" is a provider from one "h/g/p"?

qiskit/providers/ibmq/managed/ibmqjobmanager.py
jyu00: I added names of the options listed in `assemble()` to the docstring as examples, but I don't think it's necessary to duplicate their descriptions. Also would like to avoid having to keep them in sync.

qiskit/providers/ibmq/managed/ibmqjobmanager.py
jyu00: Yes!

qiskit/providers/ibmq/managed/managedjobset.py
```diff
@@ -0,0 +1,287 @@
...
+    def submit_results(self) -> None:
+        """Collect job submit responses.
+
+        Raises:
+            IBMQJobManagerInvalidStateError: If jobs have not been submitted.
+        """
+        if not self._managed_jobs:
+            raise IBMQJobManagerInvalidStateError("Jobs need to be submitted first!")
+
+        if all(mjob.future is None for mjob in self._managed_jobs):
```
diego-plan9: Can we find a different way of checking if the managed job has been submitted other than removing its `.future`?

Having the future around can be handy (for example, for some instances of cancelling) and I think we can assume the future is always in progress (either waiting for a spot or running) - maybe `future.done()` can help. If no good alternatives are found, we can have some sort of flag at the `ManagedJobSet` level (ie. `ManagedJobSet()._i_have_really_collected_everything = True`.

qiskit/providers/ibmq/managed/managedjobset.py
```diff
@@ -0,0 +1,287 @@
...
+                except JobError as err:
+                    warnings.warn(
+                        "Unable to retrieve status for job {}, job ID={}, "
+                        "for experiments {}-{}: {}".format(
+                            i, mjob.job.job_id(), mjob.start_index, mjob.end_index, err))
+            statuses.append(None)
+
+        return statuses
+
+    @requires_submit
```
diego-plan9: This `requires_submit` would make `report()` not really return immediately, right? Thinking that part of the usefulness of the manager will come from submitting a large number of jobs, I think we should try to just add a summary for "jobs that have been created but are waiting for the executor to have a free spot" - I don't mind squeezing it into `Pending jobs:` or have a new line entirely.

qiskit/providers/ibmq/managed/__init__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
diego-plan9: Can you add `qiskit.providers.ibmq.managed` to `setup.py` `packages`?

qiskit/providers/ibmq/managed/managedjob.py
```diff
@@ -0,0 +1,61 @@
...
+class ManagedJob:
+    """Job managed by job manager."""
+
+    def __init__(
+            self,
+            experiments: Union[List[QuantumCircuit], List[Schedule]],
+            start_index: Optional[int] = None,
+            future: Optional[Future] = None,
+            job: Optional[IBMQJob] = None
+    ):
```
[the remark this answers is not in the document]
jyu00: Yeah they were leftovers from `retrieve_job_set()`.

qiskit/providers/ibmq/managed/managedjobset.py
jyu00: If `report()` and `statuses()` don't wait for submit results, then a separate thread needs to wait and collect them. Since the users are likely to invoke `report()` or `statuses()` to check the progress of their jobs.

PR #407, opened 2019-10-23
Managed results

jyu00:
### Summary
Addresses #333. This is a follow up to #389 and #400. 

Introduces a `ManagedResults` class that have the same methods (`get_memory()`, `get_counts()`, etc) as `Result`. It allows the user to do
```
job_set = job_manager.run()
result = job_set.results()
result.get_counts(1)
```

### DIFF
--- a/qiskit/providers/ibmq/managed/managedjob.py
+++ b/qiskit/providers/ibmq/managed/managedjob.py
@@ -88,7 +88,7 @@ class ManagedJob:
            timeout: number of seconds to wait for job
 
         Returns:
-            Result object
+            Result object or ``None`` if result could not be retrieved.
 
         Raises:
             JobTimeoutError: if the job does not return results before a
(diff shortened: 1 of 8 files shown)


qiskit/providers/ibmq/managed/managedjob.py
```diff
@@ -111,14 +111,19 @@ def status(self) -> Optional[JobStatus]:
...
+    ) -> Optional[Result]:
         """Return the result of the job.
 
         Args:
            timeout: number of seconds to wait for job
+           partial: If true, attempt to retrieve partial job results.
 
         Returns:
-            Result object
+            Result object or ``None`` if result could not be retrieved.
```
diego-plan9: Not blocking, but in the same spirit as the `IBMQJob.result()` discussion in an earlier PR, I wonder if it would not be too costly (in terms of implications and changes) to avoid having this method return `None`, for preserving the parallelism between a managed job and a IBMQJob.

Issue #406, opened 2019-10-23
Reintroduce backend.jobs() and backend.retrieve_job()


diego-plan9:
### What is the expected behavior?

As revised by @ajavadia , `backend.jobs()` and `backend.retrieve_job()` are not to be deprecated during this round. We should:

* remove the `DeprecationWarning`
* when retrieving a job from a backend, reintroduce the check for ensuring that the job really belongs to that backend (probably along with the tests - see #354)
* revise the changelog

diego-plan9:
Closed via #408

PR #408, opened 2019-10-23
Reintroduce previous backend job functionality

luisg5:
### Summary

Adresses #406

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ The format is based on [Keep a Changelog].
   names (\#303).
 - A new `IBMQJobManager` class that takes a list of circuits or pulse schedules as 
   input, splits them into one or more jobs, and submits the jobs (\#389).
+- The previous functionality of `Backend.jobs()` and `Backend.retrieve_job()`
+  has been reintroduced, rather than being deprecated (\#354, #408). 
 
 ### Changed
 
@@ -59,10 +61,6 @@ The format is based on [Keep a Changelog].
   Only credentials from the new Quantum Experience can be used. (\#344)
 
 ### Deprecated
-
-- `Backend.jobs()` and `Backend.retrieve_job()` have been deprecated in favour
-  of `provider.backends().jobs()` and `provider.backends().retrieve_job()`.
-  (\#354).
 
 ## [0.3.3] - 2019-09-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,8 @@ The format is based on [Keep a Changelog].
   names (\#303).
 - A new `IBMQJobManager` class that takes a list of circuits or pulse schedules as 
   input, splits them into one or more jobs, and submits the jobs (\#389).
-- The previous functionality of `Backend.jobs()` and `Backend.retrieve_job()`
-  has been reintroduced, rather than being deprecated (\#354, #408). 
+- Added `provider.backends().jobs()` and `provider.backends().retrieve_job()`
+  (\#354).
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -60,10 +60,6 @@ The format is based on [Keep a Changelog].
 
 ### Deprecated
 
-- `Backend.jobs()` and `Backend.retrieve_job()` have been deprecated in favour
```
diego-plan9: Instead of removing the whole comment, can you reword it and move it to `Added`? Don't worry too much about the wording itself - the important bit is having it on the changelog in order to make the changelog cleanup before the release easier.

test/ibmq/test_ibmq_job.py
```diff
@@ -221,6 +221,22 @@ def test_get_jobs_from_backend(self, provider):
...
 
+        start_time = time.time()
+        job_list = backend.jobs(limit=5, skip=0)
+        self.log.info('found %s jobs on backend %s',
+                      len(job_list), backend.name())
+
+        for job in job_list:
+            self.log.info('status: %s', job.status())
+            self.assertTrue(isinstance(job.job_id(), str))
+        self.log.info('time to get job statuses: %0.3f s',
```
diego-plan9: Can you take the chance to drop the `self.log.info` (and the `start_time`, etc) from this test and from `test_get_jobs_from_backend_service`? They are a bit of a relic from older times and they don't provide too much value - we will do a more comprehensive round later.

test/ibmq/test_ibmq_job.py
```diff
@@ -233,9 +249,20 @@ def test_get_jobs_from_backend(self, provider):
         self.log.info('time to get job statuses: %0.3f s',
                       time.time() - start_time)
 
+    def test_retrieve_job_backend(self, provider):
+        """Test retrieving a single job from a backend."""
+        backend = provider.get_backend('ibmq_qasm_simulator')
+        qobj = assemble(transpile(self._qc, backend=backend), backend=backend)
+        job = backend.run(qobj)
+
+        rjob = backend.retrieve_job(job.job_id())
```
diego-plan9: Also for cleaning up - can you rename to `retrieved_job` or something a bit more descriptive?

test/ibmq/test_ibmq_job.py
```diff
@@ -250,7 +277,6 @@ def test_retrieve_job(self, provider):
     @requires_provider
     def test_retrieve_job_uses_appropriate_backend(self, provider):
         """Test that retrieved jobs come from their appropriate backend."""
-        # TODO: remove after backend.retrieve_job() deprecation
```
diego-plan9: Great catch!

test/ibmq/test_ibmq_job.py
```diff
@@ -279,16 +305,37 @@ def test_retrieve_job_uses_appropriate_backend(self, provider):
...
 
-    @requires_provider
-    def test_get_jobs_filter_job_status(self, provider):
+    def test_get_jobs_filter_job_status_backend(self, provider):
         """Test retrieving jobs from a backend filtered by status."""
         backends = provider.backends(simulator=False)
         backend = least_busy(backends)
+        with warnings.catch_warnings():
+            # Disable warnings from pre-qobj jobs.
+            warnings.filterwarnings('ignore',
```
diego-plan9: Can you check if we can remove this line (making the `with` block not needed)?

luisg5: I didn't think they would be necessary as I was reintroducing the tests. But, I left them thinking they may have been for something else. I didn't notice any warnings being caught while running the tests. And, the only reason a warning would be triggered now in `IBMQBackend` is if one tries to do `IBMQBackend.retrieve_job` with an inappropriate backend. So, I went ahead and removed it, because it doesn't seem to be in use.

diego-plan9: Sounds good!

Issue #402, opened 2019-10-22
Add job sharing on submit


diego-plan9:
### What is the expected behavior?

As requested by the API team, it will soon be possible to set the shareability of a job at `submit` time, by passing additional parameters.

diego-plan9:
Closed via #414. Note that currently it seems that _retrieval_ of shared jobs is experiencing some hiccups on the API side that will be address in a newer API release.

PR #414, opened 2019-10-25
Allow job sharing on submit

luisg5:
### Summary

Addresses #402


qiskit/providers/ibmq/api/rest/root.py
```diff
@@ -18,6 +18,8 @@
 
 from typing import Dict, List, Optional, Any
 
+from qiskit.providers.ibmq.apiconstants import ApiJobShareLevel
```
diego-plan9: Can you make this layer _not_ use `ApiJobShareLevel`, and instead just receive the strings? Is a bit of a judgement call, but in general we usually want to keep the `client/x.py` entities "smart-ish" rather than the rest layer.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -69,35 +70,74 @@ def __init__(
...
+                global level. The possible job share levels are: "global", "hub",
+                "group", "project", and "none".
+                    * global: the job is public to any user.
+                    * hub: the job is shared between the users in the same hub.
+                    * group: the job is shared between the users in the same group.
+                    * project: the job is shared between the users in the same project.
+                    * none: the job is not shared at any level.
+                If the job share level is not specified, then the job is not shared at any level.
+
+                Note:
```
diego-plan9: Can you remove the note and the example, mostly for brevity? It is useful information, but a bit more suited to the upcoming release notes or README tweaks and to the IQX documentation.

test/ibmq/test_ibmq_job_attributes.py
```diff
@@ -20,8 +20,9 @@
 from qiskit import ClassicalRegister, QuantumCircuit, QuantumRegister
 from qiskit.providers import JobStatus
 from qiskit.providers.ibmq import least_busy
-from qiskit.providers.ibmq.job.exceptions import IBMQJobFailureError, JobError
+from qiskit.providers.ibmq.job.exceptions import (IBMQJobFailureError, JobError)
```
diego-plan9: Unintended change?

luisg5: Yeah, I'm not sure why I changed that, thanks for the catch @diego-plan9! 😅

qiskit/providers/ibmq/api/clients/account.py
```diff
@@ -169,46 +173,54 @@ def job_submit(
...
 
         Returns:
             job status.
         """
         # Get the job via object storage.
-        job_info = self.client_api.submit_job_object_storage(backend_name, job_name=job_name)
+        job_info = self.client_api.submit_job_object_storage(
+            backend_name,
+            job_name=job_name,
+            job_share_level=job_share_level.value)
```
diego-plan9: If `job_share_level` is really optional (as per 206), this line might fail?

qiskit/providers/ibmq/api/clients/account.py
```diff
@@ -169,46 +173,54 @@ def job_submit(
...
             qobj_dict: the Qobj to be executed, as a dictionary.
             job_name: custom name to be assigned to the job.
+            job_share_level: level the job should be shared at.
 
         Returns:
             job status.
         """
-        return self.client_api.job_submit(backend_name, qobj_dict, job_name)
+        return self.client_api.job_submit(
+            backend_name, qobj_dict, job_name, job_share_level=job_share_level.value)
```
jyu00: The default value for `job_share_level` is `None`, which would not have `.value`

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -69,35 +70,65 @@ def __init__(
...
-                the job.
+                the job or the job share level is not valid.
         """
         # pylint: disable=arguments-differ
+        api_job_share_level = None
+        if job_share_level:
+            try:
+                api_job_share_level = ApiJobShareLevel(job_share_level)
+            except ValueError:
+                raise IBMQBackendError(
```
jyu00: Can you raise a different exception here? I'd prefer if we separate exceptions that the user cannot control (e.g. job submit failed due to server error) from the ones that the user can fix.

qiskit/providers/ibmq/api/clients/account.py
luisg5: Yeah, that would be the case. Thanks for the catch! Went ahead and updated the two places where this would be an issue.

qiskit/providers/ibmq/api/clients/account.py
luisg5: Yeah thanks for the catch! @diego-plan9 mentioned it as well, I've fixed it accordingly.

qiskit/providers/ibmq/ibmqbackend.py
luisg5: Yes, that would be a good rule to follow. I've seen `IBMQBackendValueError` used elsewhere and it appears to be more suited for this use case.

CHANGELOG.md
```diff
@@ -58,6 +58,9 @@ The format is based on [Keep a Changelog].
   specified, `IBMQJob.result()` will return partial results for jobs with 
   experiments that failed. When specified, and there are no partial results to 
   return, the function raises an `IBMQJobFailure` exception.(\#399).
+- The `IBMQBackend.run()` function now accepts an optional `job_share_level` 
+  parameter. If specified, the job could be shared with other users at the 
+  global, hub, group, project, or none level.
```
jyu00: Can you add the PR number to this?

qiskit/providers/ibmq/api/clients/account.py
```diff
@@ -169,46 +173,57 @@ def job_submit(
...
 
         Returns:
             job status.
         """
-        return self.client_api.job_submit(backend_name, qobj_dict, job_name)
+        return self.client_api.job_submit(
+            backend_name,
+            qobj_dict,
+            job_name,
+            job_share_level=getattr(job_share_level, 'value', None))
```
jyu00: I personally don't like using `getattr` with a hard coded attribute name because it's hard to catch if the variable type changes (probably not a risk here). But I don't feel too strongly about this one if you want to keep it.

luisg5: I was slightly skeptical when deciding to use `getattr` as well, but chose to go with it in the likes of keeping everything a bit more "tidy". But, I don't think it would be a big deal if we change it to access the attribute more directly.

Issue #374, opened 2019-10-07
Allow staging tests to run on more than just staging


jyu00:
### What is the expected behavior?
PR #360 moved some slow tests to run on staging system. There is no good way right now to run them on non-staging aside from updating the code. Ideally we'd want to be able to run them on either. This can be accomplished by introducing more environment variables, or maybe we can consider moving to `pytest`.

luisg5:
While working on #420 , I've also been thinking of different ways to try and address #374. I'm currently working on something, but its still a WIP. 

Since we would want to have the staging tests run on a production device as well, regardless of whether it runs on staging or not, it seems that we would have to change the functionality of the `run_on_staging` decorator. Namely, we would have to change it so that the test is not skipped if the staging credentials are not present. One possibility I've been working on is to have it return the staging provider (if the staging credentials are set) or `None` (if the staging credentials are not set). This would allow it to be used by `requires_device[s]`, which could append a `backends` argument to the tests decorated with it.

Would it be feasible to move in this direction? I know pytest has also been mentioned as an alternative.

PR #420, opened 2019-10-29
Add ability to run tests on staging locally

luisg5:
### Details and comments

This PR addresses being able to **optionally** run some of the tests on staging locally by introducing a new `requires_device` decorator. The `requires_device` decorator replaces the `requires_provider` in some tests, specifically those that meet the use the pattern:
- `backend = least_busy(provider.backends(simulator=False))` to retrieve a backend (i.e. the test is meant to be run on a real device).

To accomplish this, it requires the introduction of one new environment variable `QE_DEVICE`, which is meant to store the name of the device to test on. Therefore, when tests are to be run locally (and targeting a specific device), this environment variable could be set. If the environment variable is not set locally, then the tests use the default behavior; they use the least busy open backend.

### DIFF
--- a/test/ibmq/test_ibmq_job.py
+++ b/test/ibmq/test_ibmq_job.py
@@ -83,9 +83,10 @@ class TestIBMQJob(JobTestCase):
         self.assertGreater(contingency2[1], 0.01)
 
     @slow_test
-    @requires_device
-    def test_run_device(self, backend):
+    @requires_provider
+    def test_run_device(self, provider):
         """Test running in a real device."""
+        backend = least_busy(provider.backends(simulator=False))
         qobj = assemble(transpile(self._qc, backend=backend), backend=backend)
         shots = qobj.config.shots
         job = backend.run(qobj)
--- a/test/ibmq/test_ibmq_job.py
+++ b/test/ibmq/test_ibmq_job.py
@@ -36,7 +36,7 @@ from qiskit.result import Result
 
 from ..jobtestcase import JobTestCase
 from ..decorators import (requires_provider, requires_qe_access,
-                          run_on_staging)
+                          run_on_staging, requires_device)
 
 
 class TestIBMQJob(JobTestCase):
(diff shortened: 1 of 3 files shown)


luisg5:
While working on this PR, I've also been thinking of different ways to try and address #374. I'm currently working on something, but its still a WIP. 

Since we would want to have the staging tests run on a production device as well, regardless of whether it runs on staging or not, it seems that we would have to change the functionality of the `run_on_staging` decorator. Namely, we would have to change it so that the test is not skipped if the staging credentials are not present. One possibility I've been working on is to have it return the staging provider (if the staging credentials are set) or `None` (if the staging credentials are not set). This would allow it to be used by `requires_device[s]`, which could append a `backends` argument to the tests decorated with it.

Would it be feasible to move in this direction? I know pytest has also been mentioned as an alternative.

diego-plan9:
I have some trouble reconciling the PR description with the implementation - as a small recap:
* the ability to run the tests against staging is already present (ie. if the user has the right token and URL configured in the same way the production credentials are configured, the suite will be run against staging).
* some recent changes introduced the need of being able to select the `backend` more intelligently when running the tests manually - this is the main requirement for this PR.
* `@run_on_staging` is part of a different feature, aimed at CIs (offloading some tests). Ideally should be tackled in a different effort (can you move https://github.com/Qiskit/qiskit-ibmq-provider/pull/420#issuecomment-547643877 to #374?)

> To accomplish this, it requires the introduction of one new environment variable QE_STG_DEVICE ...

With the proposed changes, it seems `@requires_device` does not only require that environment variable - it actually requires the rest of the staging environment variables, and uses them to override the "normal" credentials, which blurs the lines and makes launching the tests manually more complex without any clear gain.

So can we re-target the PR in order to:
* keep the `@requires_device` decorator
* make it simply check if an environment variable is set: if it does, get the name of the backend from it, else use `least_busy()`
* rename the `QE_STG_DEVICE` environment variable to `QE_DEVICE` (and ping @jyu00 for updating accordingly in the CIs at the right time!)

This would allow us to have a simple way of setting the target device for the tests, which would solve the requirement - and as a bonus, provide a simple way of using other backends (which is a need that does not happen too often, but it does occur!).

test/ibmq/test_ibmq_job.py
```diff
@@ -87,7 +87,6 @@ def test_run_simulator(self, provider):
     def test_run_device(self, provider):
         """Test running in a real device."""
         backend = least_busy(provider.backends(simulator=False))
```
diego-plan9: Can you do a round of checking the instances where `least_busy()` is used? I'm not sure if this test was left out of the `@required_backend`ification on purpose (based on the PR description), and could be the case for others.

PR #427, opened 2019-11-04
Add IBMQJob todict method

jwoehr:
### Summary

Added IBMQJob.to_dict()

### Details and comments

Because reasons.


jwoehr:
I'm not sure where it got the idea I'm merging 90 commits ,,, they're all just keeping up with master.

jwoehr:
should I close this and re-fork?

diego-plan9:
Actually, `IBMQJob` should already have a `.to_dict()` due to inheritance - can you elaborate on the rationale? It is not a method that is encouraged to be used, though, as even if the Job is a model class, it has some particularities.

For the 91 commits, usually they are not too big of a problem as the get squashed during the merges. However, for manageability, perhaps recreating your `master` branch (and make it based on this repository master branch) would help in future PRs? It's a matter of personal preference - I tend to recommend keeping your fork's `master` as close as possible to the main repo's `master` (just `pull`ing every now and them from upstream), so when the time of making a new PR comes, you can just base it off your `master` branch and be confident that it contains the latest changes.

jwoehr:
> Actually, `IBMQJob` should already have a `.to_dict()` due to inheritance

I see, it's in `qiskit.validation.BaseModel`

I guess it was the limits of my Python. I wrote a piece of code:

```
_op = getattr(job_exp, 'to_dict', None)
if _op and callable(_op):
    self._pp.pprint(job_exp.to_dict())
else:
    self._pp.pprint(job_exp.__dict__)
```
and tested and it kept going through the second branch. I also tried calling `to_dict()` from the job returned by `execute` from qiskit-ibmq-provider and it just got an error. 

> can you elaborate on the rationale? It is not a method that is encouraged to be used, though, as even if the Job is a model class, it has some particularities.

In my Qis Job project I print the job dictionary if the user requests it so they can go looking for their job result later.

It doesn't really matter, I've found that `job.__dict__` works .. wonder why `job.to_dict()` didn't.

Again, the limits of my Python :)

jwoehr:
>I tend to recommend keeping your fork's `master` as close as possible to the main repo's `master` (just `pull`ing every now and them from upstream), so when the time of making a new PR comes, you can just base it off your `master` branch and be confident that it contains the latest changes.

Yes, I have a script that I run that does that several times a day. Not sure why it showed up as 90 mods, though it certainly pulled master that many times. Git, like Python, is something of a mystery ;)

jwoehr:
Moot.

jyu00:
> Yes, I have a script that I run that does that several times a day. Not sure why it showed up as 90 mods, though it certainly pulled master that many times. Git, like Python, is something of a mystery ;)

Your script probably does a `git merge`, which creates a merge commit every time. An alternative is to do `git rebase` instead. It does not create a new commit, but it could get complicated if you have a lot of commits to move.

Issue #442, opened 2019-11-12
Nicer way to filter jobs by date


ajavadia:
`backend.jobs()` currently accepts a raw `db_filter` that can be used for pretty much any type of filtering, but using it is not always easy because one needs insight into how database queries work.
I think some of the common filtering operations should be promoted to python arguments in `backend.jobs()`. This is already done for `status`. The next one I think should be `date`.

Currently if I want jobs between sunday and monday I have to do:
```
import datetime
monday = datetime.datetime(2019, 11, 11)
sunday = datetime.datetime(2019, 11, 10)
date_filter = {'creationDate': {'lt': monday.isoformat(), 'gt': sunday.isoformat()}}
job_list = backend.jobs(db_filter=date_filter)
```

can we instead do:
```
job_list = backend.jobs(datetime_begin=sunday, datetime_end=monday)
```
or something like that?

ajavadia:
BTW since we now support two types of jobs queries, their API should be consistent:
`backend.jobs()`
`provider.backends.jobs()`
except the 2nd one also allows filtering by backend.

jyu00:
To Ali's second point, it might be worth having something that asserts the signature of the two are mostly equal. It'd also apply to the methods of `Result` and `ManagedResults`.

luisg5:
Closed via #452 and #457.

PR #443, opened 2019-11-12
Add date arguments to jobs method

luisg5:
### Summary

Addresses #442


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -273,6 +275,10 @@ def jobs(
                 and `regular expressions
                 <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions>
                 `_ can be used.
+            start_datetime: filter by start date. The `start_datetime` is used to
```
jyu00: Can you clarify the docstring a little bit? Maybe something like 
```
start_datetime: filter by start date. This is used to find jobs whose creation dates are after (greater than) this date/time.
end_datetime: filter by end date. This is used to find jobs whose creation dates are before (less than) this date/time.
```

Issue #450, opened 2019-11-14
provider.backends.jobs does not work for start and end dates


nonhermitian:
### Information

- **Qiskit IBMQ Provider version**: 0.4.1
- **API URL or version**:

- **Python version**:
- **Operating system**:

### What is the current behavior?
The new `end_datetime` and `start_datetime` for retrieving jobs does not work.  I made the following queries:
```
end= 2019-11-14 15:44:11.592289 start= 2019-11-01 00:00:00
2019 11
end= 2019-11-01 00:00:00 start= 2019-10-01 00:00:00
2019 10
end= 2019-10-01 00:00:00 start= 2019-09-01 00:00:00
2019 9
end= 2019-09-01 00:00:00 start= 2019-08-01 00:00:00
2019 8
end= 2019-08-01 00:00:00 start= 2019-07-01 00:00:00
2019 7
end= 2019-07-01 00:00:00 start= 2019-06-01 00:00:00
2019 6
end= 2019-06-01 00:00:00 start= 2019-05-01 00:00:00
2019 5
end= 2019-05-01 00:00:00 start= 2019-04-01 00:00:00
2019 4
end= 2019-04-01 00:00:00 start= 2019-03-01 00:00:00
2019 3
end= 2019-03-01 00:00:00 start= 2019-02-01 00:00:00
2019 2
end= 2019-02-01 00:00:00 start= 2019-01-01 00:00:00
2019 1
end= 2019-01-01 00:00:00 start= 2018-12-01 00:00:00
2018 12
end= 2018-12-01 00:00:00 start= 2018-11-01 00:00:00
2018 11
```
where the data is stored in a `dict[year][month]` format.  Every entry in this dict is the same, jobs are being returned outside of this range.  In particular, this device was not even online a year ago.  From `jobs[2018][11]`:

```
_creation_date=datetime.datetime(2019, 11, 14, 17, 13, 24, 963000, tzinfo=datetime.timezone.utc)
```

### Steps to reproduce the problem
```python
import datetime
now = datetime.datetime.now()
now_year = now.year
now_month = now.month

jobs = {}

jobs[now_year] = {}
jobs[now_year-1] = {}

this_month = now
prev_month = datetime.datetime(now.year, now.month, 1)
counter = now_month
for kk in range(13):
    print("end=", this_month, "start=", prev_month)
    this_month_jobs = provider.backends.jobs(backend_name='ibmq_valencia',
                                             end_datetime=this_month,
[...]

PR #452, opened 2019-11-15
Fix job date filtering for between

luisg5:
### Summary

Addresses #450 
Related to #442


qiskit/providers/ibmq/ibmqbackendservice.py
```diff
@@ -182,24 +183,29 @@ def jobs(
...
-        # Create the date query according to the parameters specified.
-        if start_datetime or end_datetime:
-            date_filter = {'creationDate': {}}  # type: ignore[var-annotated]
-            if start_datetime:
-                date_filter['creationDate'].update({'gt': start_datetime.isoformat()})
-            if end_datetime:
-                date_filter['creationDate'].update({'lt': end_datetime.isoformat()})
-
-            api_filter.update(date_filter)
+        if start_datetime:
```
jyu00: Can you remove the example of using `db_filter` to filter date in the docstring?

qiskit/providers/ibmq/ibmqbackendservice.py
```diff
@@ -182,24 +183,29 @@ def jobs(
...
-            api_filter = {**db_filter, **api_filter}
+            # Argument filters take precedence, remove duplicate `db_filter` keys.
+            for filter_ in filters:
+                for key in filter_:
+                    if key in db_filter:
+                        db_filter.pop(key, None)
+
+            # Add remaining `db_filter` items.
+            for key in db_filter:
+                filters.append({key: db_filter[key]})
```
jyu00: Instead of this complicated triple `for` loop, I think you can keep `api_filter` as a simple dictionary, and use `between` or `and` for the datetime range.

test/ibmq/test_ibmq_job.py
```diff
@@ -314,39 +314,54 @@ def test_get_jobs_filter_job_status_backend_service(self, backend, provider):
...
 
         job_list = provider.backends.jobs(backend_name=backend.name(),
-                                          limit=5, skip=0, start_datetime=past_date)
-        for job in job_list:
-            self.assertTrue(job.creation_date() > str(past_date))
+                                          limit=5, skip=0, start_datetime=past_month)
+        self.assertTrue(job_list)
+        for i, job in enumerate(job_list):
+            self.assertTrue(job.creation_date() > past_month_str,
+                            '{}) job.creation_date: {}, past_month: {}'
```
jyu00: Can you make the message more descriptive? Something like "job creation date {} is not greater than past month {}" ? So it's clear what the expected condition is. Same for the other test cases.

test/ibmq/test_ibmq_job.py
```diff
@@ -400,33 +415,42 @@ def test_get_jobs_filter_counts_backend_service(self, provider):
             self.assertTrue(any(cresult.data.counts.to_dict()['0x0'] < 500
                                 for cresult in result.results))
 
-    @requires_device
-    def test_get_jobs_filter_date_backend(self, backend):
+    @requires_provider
+    def test_get_jobs_filter_date_backend(self, provider):
```
jyu00: I don't think it's all that necessary to have a separate test case for `backend.jobs()`, since it just calls `backendservice.jobs()` (but I suppose there is no harm since it's quick - your decision). It'd be more useful to have a test case that specifies both datetime as keyword and inside `db_filter`, for example `start_date=past_month, db_filter={'creationDate': {'lt': past_two_month}}`

qiskit/providers/ibmq/ibmqbackendservice.py
luisg5: The reason I decided to go with this (I do agree it is more complicated) was to ensure we kept using `lt` and `gt` for filtering when `start_datetime` and `end_datetime` are specified, since `between` uses `lte` and `gte`. But, I don't think it would be "too big a deal" to use `lte` or `gte` for filtering jobs and I'm sure the results returned would be the same, if not nearly identical. I'll go ahead and change it to use `between`, thanks @jyu00!

PR #454, opened 2019-11-19
Skip building wheels

jyu00:
### Summary
There is nothing to compile in the provider so source distribution should be sufficient.


mtreinish:
Wheels are useful because they're more efficient they're not just about binaries. Wheels are just unzipped while sdists require running the setup.py and executing the install path locally, we should still build and publish a wheel even if it's just python code.

PR #457, opened 2019-11-20
Jobs tests refactor

luisg5:
### Summary

Adresses #460 
Related to #442 

### Details and comments

This PR ensures the API remains consistent for specific functions and classes within the provider. The two cases addressed in this PR are:
- ensuring `IBMQBackend.jobs` matches the signature of `IBMQBackendService.jobs`. 
- ensuring `ManagedResults` contains the exact same public methods `Result` defines, and that the corresponding signatures of those functions match as well.

Also, the PR removes duplicate tests that are no longer needed as a direct result of the above mentioned.


jyu00:
hi @luisg5, I don't think using a decorator is the best approach to ensure signature compatibility, as the check is more related to the style and not the functionality of the code. I think a script outside of mainline that's invoked as part of the CI check is more suited. 

Also can you include making sure `ManagedResult` has the same public methods as `Result`?

test/ibmq/test_ibmq_integration.py
```diff
@@ -143,3 +148,84 @@ def test_execute_two_remote(self, provider):
         job = execute([qc, qc_extra], backend, seed_transpiler=self.seed)
         results = job.result()
         self.assertIsInstance(results, Result)
+
+    def test_ensure_backend_jobs_signature(self):
```
jyu00: `test_ibmq_integration` seems to be an odd place to put these tests, since this test file is for "integration test" (i.e. "compile and run" jobs). Maybe you can move the backend related tests to `test_ibmq_backend` (under a different class) and `test_ibmq_managed_job_signature` to `test_ibmq_jobmanager` ?

test/ibmq/test_ibmq_integration.py
```diff
@@ -143,3 +148,84 @@ def test_execute_two_remote(self, provider):
...
+        `IBMQBackendService.jobs` if its parameter list is a subset of the
+        parameter list of `IBMQBackendService.jobs`.
+        """
+        backend_cls = IBMQBackend
+        backend_service_cls = IBMQBackendService
+
+        # Retrieve parameter lists for both classes.
+        backend_jobs_args = getattr(
+            getfullargspec(backend_cls.jobs), 'args', [])
+        provider_backends_jobs_args = getattr(
```
jyu00: why not name this variable `backend_service_jobs_args`?

test/ibmq/test_ibmq_integration.py
```diff
@@ -143,3 +148,84 @@ def test_execute_two_remote(self, provider):
...
+            getfullargspec(backend_service_cls.jobs), 'args', [])
+
+        # Ensure parameter lists not empty
+        self.assertTrue(backend_jobs_args)
+        self.assertTrue(provider_backends_jobs_args)
+
+        if not all(arg in provider_backends_jobs_args for arg in backend_jobs_args):
```
jyu00: This will not catch the case wherein a developer adds a parameter to `IBMQBackendService` but forgot to add it to `IBMQBackend`. Really the signatures should be the same other than the `backend` keyword.

test/ibmq/test_ibmq_integration.py
```diff
@@ -143,3 +148,84 @@ def test_execute_two_remote(self, provider):
...
+                backend_service_cls.jobs, '__qualname__', str(backend_service_cls.jobs))
+
+            differing_args = set(backend_jobs_args) - set(provider_backends_jobs_args)
+            # pylint: disable=duplicate-string-formatting-argument
+            raise Exception("`{}` does not match the parameters of `{}`. "
+                            "`{}` has the extra parameter(s): {}"
+                            .format(backend_jobs_name, provider_backend_jobs_name,
+                                    backend_jobs_name, differing_args))
+
+    def test_ibmq_managed_job_signature(self):
```
jyu00: shouldn't the name of the test be `test_ibmq_managed_results_signature`?

test/ibmq/test_ibmq_integration.py
```diff
@@ -143,3 +148,84 @@ def test_execute_two_remote(self, provider):
...
+        result_methods = self._get_class_methods(result_cls)
+        self.assertTrue(result_methods)
+
+        # Get `ManagedResults` public methods.
+        managed_results_methods = self._get_class_methods(managed_results_cls)
+        self.assertTrue(managed_results_methods)
+
+        # Ensure `ManagedResults` has the same public methods as `Result`.
+        differing_args = set(result_methods.keys()) - set(managed_results_methods.keys())
+        self.assertEqual(len(differing_args), 0)
```
jyu00: Can you print out `differing_args` (with a nice message) if the assert fails?

test/ibmq/test_ibmq_integration.py
```diff
@@ -143,3 +148,84 @@ def test_execute_two_remote(self, provider):
...
+    def test_ibmq_managed_job_signature(self):
+        """Test `ManagedResults` and `Result` contain the same public methods.
+
+        Note:
+            Aside from ensuring that the two classes contain the same public
+            methods, it is also necessary to check that the corresponding
+            methods have the same signature.
+        """
+        result_cls = Result
+        managed_results_cls = ManagedResults
```
jyu00: Not really a big deal, but was there a reason you created new variables for these classes? Rather than doing something like `self._get_class_methods(Result)` below?

test/ibmq/test_ibmq_job_attributes.py
```diff
@@ -62,10 +62,9 @@ def test_running_job_properties(self, provider, backend):  # pylint: disable=unu
         job = backend.run(qobj)
         _ = job.properties()
 
-    @requires_provider
-    def test_job_name_backend(self, provider):
+    @requires_device
+    def test_job_name_device(self, backend):
```
jyu00: There isn't really a need to run the job name test on both real device and simulator, as job names are tracked by the api, not the backend.

test/ibmq/test_ibmq_job_attributes.py
```diff
@@ -114,10 +113,9 @@ def test_job_name_backend_service(self, provider):
         self.assertEqual(len(retrieved_jobs), 1)
         self.assertEqual(job_id, retrieved_jobs[0].job_id())
 
-    @requires_provider
-    def test_duplicate_job_name_backend(self, provider):
+    @requires_device
+    def test_duplicate_job_name_device(self, backend):
```
jyu00: Same as above.

luisg5:
Regarding the comment of moving the backend related tests to `test_ibmq_backend`: https://github.com/Qiskit/qiskit-ibmq-provider/pull/457#discussion_r350366616. It might be best to do that in a separate PR? Might help break up the changes and reviewing the PR's easier as well.

test/ibmq/test_ibmq_backend.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+        backend_service_jobs_params = set(
+            getattr(getfullargspec(IBMQBackendService.jobs), 'args', [])
+        )
+
+        # Ensure parameter lists not empty
+        self.assertTrue(backend_jobs_params)
+        self.assertTrue(backend_service_jobs_params)
+
+        # Ensure `IBMQBackend.jobs` does not have any additional parameters.
```
jyu00: Wouldn't it be easier to just combine `backend_service_jobs_params` with `acceptable_differing_params` and assert the 2 `params` are equal?

test/ibmq/test_ibmq_integration.py
```diff
@@ -143,3 +147,41 @@ def test_execute_two_remote(self, provider):
         job = execute([qc, qc_extra], backend, seed_transpiler=self.seed)
         results = job.result()
         self.assertIsInstance(results, Result)
+
+    def test_backend_jobs_signature(self):
```
jyu00: Did you forget to remove this now that the test is in `test_ibmq_backend`?

test/ibmq/test_ibmq_jobmanager.py
```diff
@@ -283,6 +287,65 @@ def test_skipped_result(self, provider):
...
+        additional_methods = set(managed_results_methods.keys()) - set(result_methods.keys())
+        self.assertTrue((len(additional_methods) == 0),
+                        "ManagedResults does not have the same public "
+                        "methods as Result. ManagedResults has the "
+                        "additional public method(s): {}"
+                        .format(additional_methods))
+
+        # Ensure `ManagedResults` is not missing any public methods.
+        missing_methods = set(result_methods.keys()) - set(managed_results_methods.keys())
+        self.assertTrue((len(missing_methods) == 0),
```
jyu00: I think you can combine this check and the one above with a single `assertEqual`, which _may_ even print out the differences if the assert fails.

test/ibmq/test_ibmq_jobmanager.py
```diff
@@ -283,6 +287,65 @@ def test_skipped_result(self, provider):
...
+                        "the public method(s): {}".format(missing_methods))
+
+        # Ensure the signature for the methods from both classes are compatible.
+        for name, method in managed_results_methods.items():
+            managed_results_params = getattr(getfullargspec(method), 'args', [])
+            result_params = getattr(getfullargspec(result_methods[name]), 'args', [])
+            self.assertTrue(managed_results_params)
+            self.assertTrue(result_params)
+            # pylint: disable=duplicate-string-formatting-argument
+            self.assertTrue(managed_results_params == result_params,
```
jyu00: `assertEqual` is probably better here

jyu00:
> Regarding the comment of moving the backend related tests to `test_ibmq_backend`: [#457 (comment)](https://github.com/Qiskit/qiskit-ibmq-provider/pull/457#discussion_r350366616). It might be best to do that in a separate PR? Might help break up the changes and reviewing the PR's easier as well.

That comment was only about moving the new test (i.e. `test_ensure_backend_jobs_signature`) you created, not the other backend related tests.

luisg5:
> That comment was only about moving the new test (i.e. test_ensure_backend_jobs_signature) you created, not the other backend tests. Also I know the comment said test_ibmq_backend, but I really meant the existing test_ibmq_backends. But that's fine since I want to get rid of test_ibmq_backends anyway.

Since you mentioned *"backend related tests"*, I thought the comment was directed to the overall tests refactor. This made sense since certain tests within `ibmqjob.py` would make more sense in a file such as `ibmqbackend.py`, such as those that test retrieving jobs.

Another reason was seeing that `ibmqbackends.py` was a candidate for removal and `test_ensure_backend_jobs_signature` did not seem to fit into any of the existing test files.

PR #462, opened 2019-11-21
Downgrade websocket error message

jyu00:
### Summary
There is really no need to issue a warning when websocket fails and the provider is retrying with http. There is nothing a user can do about the websocket failure, and the http retry usually works. It just causes confusion to users who think it's due to their code. The message, however, might be useful for developers to debug websocket issues, so it's downgraded to ~`debug`~ `info` level. `info` level is useful for getting more information without too much noise, and it's below the default level of `warning`.

### DIFF
--- a/qiskit/providers/ibmq/api/clients/account.py
+++ b/qiskit/providers/ibmq/api/clients/account.py
@@ -373,10 +373,10 @@ class AccountClient(BaseClient):
                 status_response = self._job_final_status_websocket(job_id, timeout)
             except WebsocketTimeoutError as ex:
                 logger.debug('Timeout checking job status using websocket, '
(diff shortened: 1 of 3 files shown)


luisg5:
Since a user can't really do anything about these either, would it be reasonable to downgrade any of these two?
- https://github.com/Qiskit/qiskit-ibmq-provider/blob/34ebd547933267dca16f1704d2b4a0382d2e2f18/qiskit/providers/ibmq/api/clients/websocket.py#L281
- https://github.com/Qiskit/qiskit-ibmq-provider/blob/34ebd547933267dca16f1704d2b4a0382d2e2f18/qiskit/providers/ibmq/api/clients/websocket.py#L293

Issue #466, opened 2019-11-24
`job.result()` needs a refresh option


ajavadia:
it seems like the Job class is caching its result. But it is easy to change the result during post-processing, for example marginalizing over some qubits or doing error mitigation on the result. I think this should not affect the original job result. At least, the job.result() call should have an option to fetch the original result from the API again.

jyu00:
If a user wants to change the result object but keep the original, can't they just do a `copy.deepcopy()`?

ajavadia:
Sure but it will take some time to figure out this is happening (based on my debugging experience). I don’t think it is expected that “job.result()” is *not* returning the result of the executed job every time you call it.

jyu00:
Closed via #469

PR #469, opened 2019-11-26
add refresh to ibmqjob.result

luisg5:
### Summary

Addresses #466


qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -483,7 +486,7 @@ def _wait_for_completion(
 
         return self._status in required_status
 
-    def _retrieve_result(self) -> Result:
+    def _retrieve_result(self, refresh: bool = False) -> Result:
```
jyu00: parameter not in the docstring. I'm surprise lint didn't catch this.

jyu00:
Can you also update the changelog? Since you're changing a user facing method.

qiskit/providers/ibmq/job/ibmqjob.py
luisg5: Found out docstrings are not required for private methods: https://www.python.org/dev/peps/pep-0008/#documentation-strings. As a result, pylint does not enforce them either. Must be why it wasn't caught 😅.

CHANGELOG.md
```diff
@@ -14,6 +14,14 @@ The format is based on [Keep a Changelog].
 > - **Security**: in case of vulnerabilities.
 
 
+## [UNRELEASED]
+
+### Added
+
+- `IBMQJob.result()` now accepts an optional `refresh` parameter. If 
+  specified, the function re-queries the api for the results, rather
```
jyu00: nit - can you change `If specified` to `If refresh=true is specified`? Just to be clearer.

jyu00:
One more thing. We should get into the habit of specifying the default value (e.g. `Default: foo`) in the docstring so users can get that information from the api doc (when it's available) instead of having to look at the code.

luisg5:
> One more thing. We should get into the habit of specifying the default value (e.g. Default: foo) in the docstring so users can get that information from the api doc (when it's available) instead of having to look at the code.

Went back to add the default values for this PR. I'll make sure to do so in future ones as well 👍

PR #471, opened 2019-12-04
Modify exception message

luisg5:
### Details and comments

Replace the message for chained exceptions within `request`.


qiskit/providers/ibmq/api/session.py
```diff
@@ -199,9 +199,26 @@ def request(  # type: ignore[override]
 
             if self.access_token:
                 message = message.replace(self.access_token, '...')
-                # Replace the original message on the `RequestException` as well.
+                # Replace the original message on the `RequestException`.
                 ex.args = (message,)
```
jyu00: You don't need this line anymore, right? Since `_modify_chained_exception_messages` will replace the message?

luisg5: Yeah, you're right! We wouldn't need that line after all. Thanks @jyu00!

qiskit/providers/ibmq/api/session.py
```diff
@@ -198,10 +198,7 @@ def request(  # type: ignore[override]
                     pass
 
             if self.access_token:
-                message = message.replace(self.access_token, '...')
-                # Replace the original message on the `RequestException`.
-                ex.args = (message,)
-                # Replace the original message on the chained exceptions.
+                # Modify the original message on the chained exceptions.
```
jyu00: I think you still need `message = message.replace(self.access_token, '...')` ? Since it's used in line 204.

test/ibmq/test_account_client.py
```diff
@@ -248,6 +250,42 @@ def test_job_cancel(self):
...
+        api = backend._api
+        job = backend.run(qobj)
+        job_id = job.job_id()
+
+        session_ = api.client_api.session
+        access_token_ = session_.access_token
+
+        exception_message = 'The access token in this exception ' \
+                            'message should be replaced: {}'.format(access_token_)
+        self.assertIn(access_token_, exception_message)
```
jyu00: Is this assert really necessary? You did just format the message with the token right above...

test/ibmq/test_account_client.py
```diff
@@ -248,6 +250,42 @@ def test_job_cancel(self):
...
+
+        with self.assertRaises(RequestsApiError) as exception_cm:
+            with mock.patch.object(
+                    HTTPConnectionPool,
+                    'urlopen',
+                    side_effect=MaxRetryError(
+                        HTTPConnectionPool('host'), 'url', reason=exception_message)):
+                _ = api.job_get(job_id)
```
jyu00: Is it possible to do on the `backend.run()` (or `api.job_submit()`) path instead of `api.job_get()`? That'd make the test faster

test/ibmq/test_account_client.py
```diff
@@ -248,6 +250,42 @@ def test_job_cancel(self):
...
+
+        def _assert_token_not_in_exc_chain(test_class, access_token, exc):
+            """Helper function to assert access token not in exception chain."""
+            if exc.__cause__:
+                _assert_token_not_in_exc_chain(test_class, access_token, exc.__cause__)
+            elif exc.__context__:
```
jyu00: Can you make your verification closer to what a user would see (e.g. stderr)? If the verification uses the same code as mainline, then it doesn't really help much. For example, if `exc` has another attribute `__foo__` that also contains the access_token, both mainline and this test would fail to catch that.

luisg5: Yeah, that would make sense. Instead of relying on the same mechanism to check whether the access token is removed from the exception chain, I decided to check the entire traceback instead, since this it is what the user would see.

test/ibmq/test_account_client.py
```diff
@@ -248,6 +251,30 @@ def test_job_cancel(self):
...
+            with mock.patch.object(
+                    HTTPConnectionPool,
+                    'urlopen',
+                    side_effect=MaxRetryError(
+                        HTTPConnectionPool('host'), 'url', reason=exception_message)):
+                _ = api.job_submit(backend.name(), qobj.to_dict(), use_object_storage=True)
+        except RequestsApiError:
+            exception_chain_as_str = traceback.format_exc()
+            self.assertTrue(exception_chain_as_str)
+            self.assertNotIn(self.access_token, exception_chain_as_str)
```
jyu00: Can you move lines 275-276 out of the `except`? So if the assert fails, it doesn't print the previous exception (i.e. `RequestApiError`) that may contain the api token. Also instead of using `assertNotIn`, can you just check the token is not in the str and do a `self.fail()`? The `assertNotIn` would also print the access token if it fails.

test/ibmq/test_account_client.py
```diff
@@ -248,6 +251,30 @@ def test_job_cancel(self):
...
 
+    def test_access_token_not_in_exception_traceback(self):
+        """Check that access token is replaced within chained request exceptions."""
+        backend = self.provider.backends.ibmq_qasm_simulator
+        circuit = transpile(self.qc1, backend, seed_transpiler=self.seed)
+        qobj = assemble(circuit, backend, shots=1)
+        api = backend._api
+
+        session_ = api.client_api.session
+        access_token_ = session_.access_token
```
jyu00: Any reason you don't just use `self.access_token`?

luisg5: I must have overseen that... Thanks for the catch!

PR #475, opened 2019-12-06
Allow retrying on specific POST requests

diego-plan9:
### Summary

Allow retrying on specific responses to `POST` queries (ie. the ones on `STATUS_FORCELIST`). By default, retrying on `POST` is not recommended due to the requests not being idempotent (risking causing side effects), but for those specific errors the API team confirmed it was safe to retry. The implementation is via customizing `Retry` instead of adding `POST` directly to `method_whitelist`, for leaning on the side of caution and not retry all POST requests.

Additionally, tweaks a bit the message logged when there is an object storage failure during job submit, in order to help troubleshooting.

### DIFF
--- a/qiskit/providers/ibmq/api/clients/account.py
+++ b/qiskit/providers/ibmq/api/clients/account.py
@@ -165,11 +165,12 @@ class AccountClient(BaseClient):
                     qobj_dict=qobj_dict,
                     job_name=job_name,
                     job_share_level=job_share_level)
-            except Exception:  # pylint: disable=broad-except
+            except Exception as ex:  # pylint: disable=broad-except
                 # Fall back to submitting the Qobj via POST if object storage
                 # failed.
                 logger.info('Submitting the job via object storage failed: '
-                            'retrying via regular POST upload.')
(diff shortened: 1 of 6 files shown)


qiskit/providers/ibmq/api/clients/account.py
```diff
@@ -165,11 +165,12 @@ def job_submit(
...
                     job_name=job_name,
                     job_share_level=job_share_level)
-            except Exception:  # pylint: disable=broad-except
+            except Exception as ex:  # pylint: disable=broad-except
                 # Fall back to submitting the Qobj via POST if object storage
                 # failed.
                 logger.info('Submitting the job via object storage failed: '
-                            'retrying via regular POST upload.')
+                            'retrying via regular POST upload: %s',
+                            str(ex))
```
jyu00: Given object storage involves multiple steps, is this sufficient for debugging? Maybe we can log the traceback at debug level?

diego-plan9: Yes! Added in 19b5f76

PR #476, opened 2019-12-06
Allow retrieving QASM jobs

### LINKED ISSUE #473: Support retrieving failed composer job
A user should be able to get an error message for a composer job that failed validation.

Composer jobs do not have the qobject until the validator service creates one. If the validation fails, then the job will not have a qobject and therefore no `kind` field. This would cause `IBMQJob` model validation to fail.

jyu00:
### Summary
Fixes #473. Note that a QASM job can be retrieved, but very few functions can be called with the job because the format is no longer supported.


test/ibmq/test_ibmq_jobmanager.py
```diff
@@ -282,7 +282,9 @@ def test_skipped_result(self, provider):
 
         result_manager = job_set.results()
         if cancelled:
-            with self.assertRaises(IBMQManagedResultDataNotAvailable):
+            with self.assertRaises(IBMQManagedResultDataNotAvailable,
+                                   msg="IBMQManagedResultDataNotAvailable not "
+                                       "raised for job {}".format(cjob.job_id())):
```
luisg5: Were you looking to log or print the `msg` if `IBMQManagedResultDataNotAvailable` was not actually raised?

If so, I don't believe it would be. It mentions on the docs that `msg` is only used by the context manager of `self.assertRaises`. To actually out the message, I think we would need to wrap `self.assertRaises` in a try and except to catch the exception we don't expect, then print out the message. That'd be one solution at least.

jyu00: Yes the `msg` is intended to be printed if the exception is not raised. I _am_ using `assertRaises()` as a context manager though, and the message does get printed:
```
AssertionError: IBMQManagedResultDataNotAvailable not raised : IBMQManagedResultDataNotAvailable not raised for job 5deec46b03329400119bf1c0
```

luisg5: Yeah, you're right! I was thinking of the case where a different exception is raised, sorry.

PR #483, opened 2019-12-12
Fix job filter test

jyu00:
### Summary
Fix `test_retrieve_jobs_filter_counts` and `requires_device`.

### Details and comments
- Now that we've switched to object storage, we can no longer filter jobs by result counts (since results are no longer in job data).
- Update `requires_device` decorator to ensure the backend has at least 5 qubits.


test/ibmq/test_ibmq_job.py
```diff
@@ -377,20 +377,34 @@ def test_retrieve_jobs_between_datetimes_not_overriden(self, provider):
...
+
+        my_filter = {'backend.name': backend.name(),
+                     'summaryData.summary.qobj_config.n_qubits': 3,
+                     'status': 'COMPLETED'}
 
         job_list = provider.backends.jobs(backend_name=backend.name(),
-                                          limit=5, skip=0, db_filter=my_filter)
+                                          limit=2, skip=0, db_filter=my_filter)
+        self.assertEqual({o_job.job_id() for o_job in jobs},
+                         {r_job.job_id() for r_job in jobs})
```
luisg5: Is `jobs` on line `400` supposed to be `job_list`?

jyu00: Yes

Issue #491, opened 2019-12-17
Job class missing methods in Qiskit spec


nonhermitian:
### What is the expected behavior?
The qiskit spec has the following for the jobs class:

>The job also has additional calls Job.done(), Job.running() and Job.cancelled() which returns a boolean (true/false) to indicate whether the job is done, running, or cancelled.

Do we want these?  In my experience this would be super nice because checking if jobs are done is currently confusing / ugly.

ewinston:
This is not a comment about whether we want these but if we did it would probably be better to put in terra basejob for consistency. I liked having them but I believe they placed a burden on the servers since they are often used in loops. Not sure if recent updates to the api changes things?

nonhermitian:
Not sure how this is different than people just calling `job.status()` repeatedly in loops.  I think the websockets take care of this anyway.  The nice thing with these is that it is simply to do `if job.done()`.  Currently one has to know that `if job.status() == job.status().DONE` is the way to do it, but this is really not intuitive.  Similarly with using the`JobStatus` itself, which is not imported by default as well.

jyu00:
> I liked having them but I believe they placed a burden on the servers since they are often used in loops.

That is indeed an issue. The api's solution is to use websocket, but the provider's implementation today doesn't update the user with intermediate status. There is an [issue opened in Aqua](https://github.com/Qiskit/qiskit-aqua/issues/752) on this, where they call `job.status()` (which doesn't use websocket) in a loop. The provider will hopefully address this in the near future.

nonhermitian:
So the changes presented here do not change the status quo, that is to say an end user can query the status as much as they like and get blocked by the IBM cloud.  So I do not understand why having convenience functions that make checking the status easier are an issue? The Aqua issue is just developers not understanding the limitations of this call.  The monitoring tools in Terra know to adaptively change the status check interval based on job position.

jyu00:
I agree these convenience functions are nice to have. I was just pointing out that calling `job.status()` in a loop is a known issue that will be addressed separately.

ewinston:
A local simulator doesn't mind putting this in a loop so it could be in basejob but overridden by providers as they see fit.

jyu00:
Since it's in the spec it seems reasonable to have them in `BaseJob`.

jyu00:
[...]

PR #494, opened 2019-12-19
add job status methods.

luisg5:
### Summary

Addresses: #491


CHANGELOG.md
```diff
@@ -14,6 +14,17 @@ The format is based on [Keep a Changelog].
 > - **Security**: in case of vulnerabilities.
 
 
+## [UNRELEASED]
+
+### Added
+
+- `IBMQJob` now has three new methods: `done()`, `running()`, and
+  `cancelled()`. The methods are used to confirm the current status
```
jyu00: The wording "used to confirm" sounds really weird. How about "The methods are used to indicate the job status". Also I don't think it's necessary to list the methods individually below as they're pretty self explanatory. These methods are introduced so users don't need to know about `JobStatus`, and saying "job status is foo" kind of defeats the purpose.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -290,6 +290,45 @@ def status(self) -> JobStatus:
 
         return self._status
 
+    def done(self) -> bool:
+        """Return whether the job is done.
```
jyu00: Can you use the same wording as `JobStatus`? Like "Return whether the job has successfully run".

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -290,6 +290,45 @@ def status(self) -> JobStatus:
...
+            job_status: the job status to check against.
+
+        Returns:
+            True if the current job status matches the desired one, else false.
+        """
+        if self._status in JOB_FINAL_STATES:
+            return self._status == job_status
+
+        self.status()  # Get the latest job status.
+        return self._status == job_status
```
jyu00: You don't actually need to check for `JOB_FINAL_STATES` since `job.status()` already does that. So this method can just be `return self.status() == job_status`

test/ibmq/test_ibmq_job_states.py
```diff
@@ -125,6 +125,36 @@ def test_unrecognized_status(self):
         with self.assertRaises(IBMQJobApiError):
             self.wait_for_initialization(job)
 
+    def test_done_status(self):
```
jyu00: I know most of the tests in this class don't have docstrings, but two wrongs don't make a right ;)

luisg5: Yes, I agree completely! 😂

PR #497, opened 2019-12-19
Use internal provider

### LINKED ISSUE #495: Allow tests to run under non-default provider
New IQX update introduces a 5-job per backend limit, including `ibmq_qasm_simulator`. This causes some Travis tests to fail, if they happen to have 5 jobs running in parallel. This should be fixed.

jyu00:
### Summary
Fixes #495 . Introduces new environment variable `QE_HGP` to select which provider to use.


test/ibmq/test_ibmq_job.py
```diff
@@ -308,9 +310,11 @@ def test_retrieve_jobs_start_datetime(self, provider):
                             'greater than or equal to past month: {}'
                             .format(i, job.creation_date(), past_month_str))
 
-    @requires_provider
-    def test_retrieve_jobs_end_datetime(self, provider):
+    @requires_qe_access
+    def test_retrieve_jobs_end_datetime(self, qe_token, qe_url):
```
luisg5: > marker

luisg5: Is there a specific reason to use `requires_qe_access` to activate a provider in the `test_retrieve_jobs_*` test cases, rather than just using `requires_provider`?

jyu00: Yes. Because this sets of tests (`test_retrieve_jobs_end_datetime` etc) require there to be jobs submitted in the previous 2 months. `requires_provider` uses the provider specified via `QE_HGP`, which is new and doesn't have jobs from last 2 months.

PR #500, opened 2019-12-24
Update requirements.txt

chlorophyll-zz:
requests-ntlm doesnt exist
new name: requests_ntlm


jyu00:
Thanks @chlorophyll-zz!

PR #501, opened 2020-01-02
Update test_filter_backends

diego-plan9:
### Summary

Coming from the API team, fix the `test_filter_config_properties` for avoid it failing in staging: the test was hard-coded for a number of qubits and in some (common/default) `staging` configuration it was unable to find a backend with the specified requisites. This PR makes the tests depend on the backend chosen for other tests, hopefully making it more robust.

In the process, extended the asserts of other the tests in the file to provide stronger guarantees that the filtering is being done correctly.

### DIFF
--- a/test/ibmq/test_filter_backends.py
+++ b/test/ibmq/test_filter_backends.py
@@ -17,19 +17,26 @@
 from qiskit.providers.ibmq import least_busy
 
 from ..ibmqtestcase import IBMQTestCase
-from ..decorators import requires_provider
+from ..decorators import requires_provider, requires_device
 
 
 class TestBackendFilters(IBMQTestCase):
     """Qiskit Backend Filtering Tests."""
 
-    @requires_provider
-    def test_filter_config_properties(self, provider):
+    @requires_device
+    def test_filter_config_properties(self, backend):
         """Test filtering by configuration properties"""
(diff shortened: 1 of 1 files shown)


jyu00:
Thanks @diego-plan9!

test/ibmq/test_filter_backends.py
```diff
@@ -17,19 +17,26 @@
...
+        provider = backend._provider
+        n_qubits = backend.configuration().n_qubits
 
         filtered_backends = provider.backends(n_qubits=n_qubits, local=False)
+
         self.assertTrue(filtered_backends)
+        for filtered_backend in filtered_backends:
+            with self.subTest(backend=filtered_backend):
+                self.assertEqual(n_qubits, backend.configuration().n_qubits)
+                self.assertFalse(backend.configuration().local)
```
luisg5: I believe it might be better to use `filtered_backend` for the asserts instead of `backend`. After running the test, it seems that `backend` actually refers to the backend passed by the decorator and not the one used by `subTest`. As a result, it seems that the test continually asserts the same backend as of now.

diego-plan9: Yes! Good catch, thanks :+1:

PR #506, opened 2020-01-08
Add job share level to managed jobs

jyu00:
### Summary
Related to #402 and #333. Add shareability to managed jobs.


jyu00:
Yeah I guess it's worth adding a test case for that since it does its own checking.

PR #508, opened 2020-01-09
Make share level and status enums case insensitive

luisg5:
### Details and comments

Converts the strings passed to `ApiJobShareLevel` and `ApiJobStatus` to lowercase and uppercase, respectively.

### DIFF
--- a/qiskit/providers/ibmq/job/ibmqjob.py
+++ b/qiskit/providers/ibmq/job/ibmqjob.py
@@ -281,7 +281,7 @@ class IBMQJob(BaseModel, BaseJob):
 
         with api_to_job_error():
             api_response = self._api.job_status(self.job_id())
-            self._update_status_position(ApiJobStatus(api_response['status']),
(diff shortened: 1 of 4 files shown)


jyu00:
Can you also do this for `status` in `IBMQBackendService.jobs()` ?

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -555,7 +555,7 @@ def _wait_for_completion(
             except UserTimeoutExceededError:
                 raise IBMQJobTimeoutError(
                     'Timeout while waiting for job {}'.format(self._job_id))
-        self._update_status_position(ApiJobStatus(status_response['status']),
+        self._update_status_position(ApiJobStatus(status_response['status'].upper()),
```
jyu00: These two are actually unnecessary since `StatusResponseSchema` already mandates `status` value must be one of the `ApiJobStatus` values.

PR #511, opened 2020-01-10
Add job tags

### LINKED ISSUE #507: Allow tagging jobs
The API now supports tagging jobs. It was originally implemented to support #481, but it'd be nice to allow end users the ability.

jyu00:
### Summary
Allow the user to assign tags to a job and subsequently use them for filtering. 
Closes #507

### DIFF
--- a/qiskit/providers/ibmq/managed/managedjobset.py
+++ b/qiskit/providers/ibmq/managed/managedjobset.py
@@ -81,7 +81,8 @@ class ManagedJobSet:
             raise IBMQJobManagerInvalidStateError("Jobs were already submitted.")
 
         self._backend = backend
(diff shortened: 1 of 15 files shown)


qiskit/providers/ibmq/managed/managedjobset.py
```diff
@@ -78,14 +81,16 @@ def run(
             raise IBMQJobManagerInvalidStateError("Jobs were already submitted.")
 
         self._backend = backend
+        self._tags = job_tags.copy()
```
luisg5: If `job_tags` is not specified, it would default to `None` and this would fail. We should do a check before copying `job_tags` to prevent this

Issue #502, opened 2020-01-06
Report job limits


jyu00:
### What is the expected behavior?
Report the total number of jobs a user can run in the given provider and backend, as well as the current number of jobs the user is running.

PR #513, opened 2020-01-10
Add Backend Jobs Limit

luisg5:
### Summary

Addresses #502


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -41,6 +43,22 @@
 logger = logging.getLogger(__name__)
 
 
+@bind_schema(BackendJobsLimitResponseSchema)
+class BackendJobsLimit(BaseModel):
```
luisg5: Since `IBMQBackend.jobs_limit()` is a user facing method, I thought it best to create a model so users could access the attributes directly, instead of receiving a dictionary with the fields. However, I'm not sure on the best place (*file*) to place this model in?

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ The format is based on [Keep a Changelog].
 
 - `IBMQJob` now has three new methods: `done()`, `running()`, and
   `cancelled()`. The methods are used to indicate the job status. (\#494)
+- `IBMQBackend` now has a new `jobs_limit()` method that returns the 
+  jobs limit for a backend, such as the current number of running jobs 
+  on the backend and also the maximum number of jobs that could be 
+  submitted to the backend. (\#513)
```
jyu00: Can you check with the api team to see if the limit is per provider? And clarify that here and the method's docstring.

qiskit/providers/ibmq/api/rest/validation.py
```diff
@@ -63,3 +63,20 @@ class StatusResponseSchema(BaseSchema):
...
     # Required properties
     status = String(required=True, validate=OneOf([status.value for status in ApiJobStatus]))
+
+
+class BackendJobsLimitResponseSchema(BaseSchema):
+    """Schema for BackendJobsLimit"""
+
+    # required properties
+    maximum_jobs = Integer(required=True)
+    running_jobs = Integer(required=True)
```
jyu00: I think we should avoid making fields required if possible, so older versions of the provider won't blow up if the api changes the naming. You can instead of `missing=None` and add some wording saying `None` is returned if the information is not available.

qiskit/providers/ibmq/ibmqbackend.py
jyu00: Can you move this class to its own module?

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -41,6 +43,22 @@
...
+    """Jobs limit for a backend."""
+    def __init__(self, maximum_jobs: int, running_jobs: int, **kwargs: Any) -> None:
+        """Creates a new BackendJobsLimit instance.
+
+        Args:
+            maximum_jobs: maximum number of jobs for the backend.
+            running_jobs: current number of jobs running on the backend.
```
jyu00: Can you 1) check with Kevin (since he's working on the same thing for iqx) to ensure the field names match and 2) clarify a bit what `maximum_jobs` mean (e.g. `maximum number of jobs this account is allowed to run on this backend with this provider at a time`).

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -254,6 +272,29 @@ def defaults(self, refresh: bool = False) -> Optional[PulseDefaults]:
 
         return self._defaults
 
+    def jobs_limit(self) -> BackendJobsLimit:
```
jyu00: `jobs_limit` sounds a bit weird? Is `job_limit` better?

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -254,6 +272,29 @@ def defaults(self, refresh: bool = False) -> Optional[PulseDefaults]:
...
+        """Return the jobs limit for the backend.
```
jyu00: Can you update the docstring to make it more informative on how to use this?

test/ibmq/test_account_client.py
```diff
@@ -163,6 +163,14 @@ def test_backend_properties(self, backend):
         properties = api.backend_properties(backend.name())
         self.assertIsNotNone(properties)
 
+    @requires_device
+    def test_backend_jobs_limit(self, backend):
+        """Check the backend job limits of a real backend."""
+        api = self._get_client()
+
+        jobs_limit = api.backend_jobs_limit(backend.name())
+        self.assertIsNotNone(jobs_limit)
```
jyu00: Can you also check that the attributes of `jobs_limit` are not `None` (and maybe not 0)?

jyu00:
It might also be nice to have a helper method that tells the user how many more jobs they can submit to the backend before the limit is reached, so they don't have to decipher `BackendJobsLimit`.

CHANGELOG.md
luisg5: From what I have seen, the limits for a user are for a given provider/device combination. But, I'll double check with them 👍

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -272,29 +255,72 @@ def defaults(self, refresh: bool = False) -> Optional[PulseDefaults]:
...
+        jobs running on the backend at this time and the maximum number
+        of concurrent jobs a user could submit to the backend at a time.
 
         Note:
-            If `BackendJobsLimit.maximum_jobs = -1`, then there
-            are no limits to the maximum number of jobs that could
-            be submitted on the backend.
+            The job limit information for the backend is given for
+            this account (i.e. a specific backend with a specific
+            provider).
```
jyu00: How about
`The job limit information for the backend is provider specific. For example, if you have access to the same backend via different providers, the job limit information might be different for each provider.
`

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -272,29 +255,72 @@ def defaults(self, refresh: bool = False) -> Optional[PulseDefaults]:
...
+
+            If the method call was successful, you can inspect the job
+            limit for the backend by accessing the ``maximum_jobs``
+            and ``running_jobs`` attributes of the ``BackendJobLimit``
+            class returned.
```
jyu00: nit - but it's really a `BackendJobLimit` _instance_ (not class) that's returned

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -272,29 +255,72 @@ def defaults(self, refresh: bool = False) -> Optional[PulseDefaults]:
...
+        api_job_limit = self._api.backend_job_limit(self.name())
 
         try:
-            return BackendJobsLimit.from_dict(api_jobs_limit)
+            return BackendJobLimit.from_dict(api_job_limit)
         except ValidationError as ex:
             raise LookupError(
                 "Couldn't get backend jobs limit: {0}".format(ex))
 
+    def remaining_jobs(self) -> int:
```
jyu00: Can you change the method name to something like `remaining_jobs_count` ? Just to be clearer that a "count" is returned, not jobs.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -272,29 +255,72 @@ def defaults(self, refresh: bool = False) -> Optional[PulseDefaults]:
...
+        """Return remaining jobs that could be submitted to the backend.
+
+        Return the remaining number of jobs that could be submitted to the
+        backend, for this account, before the job limit is reached. See
+        ``IBMQBackend.job_limit()`` for the job limit of a given backend.
```
jyu00: I'd avoid using the word `account`, since an account can have multiple providers. Using `provider` is more accurate. Also, I'd add the word `concurrent` to the description since the current wording makes it sound like the limit is for all jobs (running or finished). Maybe something like

`Return the number of jobs that can be submitted to this backend with this provider before the limit on concurrent jobs is reached.`

This applies to the rest of the docstring.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -272,29 +255,72 @@ def defaults(self, refresh: bool = False) -> Optional[PulseDefaults]:
...
+            this account (i.e. a specific backend with a specific
+            provider).
+
+            If the method call was successful, you can inspect the job
+            limit for the backend by accessing the ``maximum_jobs``
+            and ``running_jobs`` attributes of the ``BackendJobLimit``
+            class returned.
+
+            For example:
+                backend_job_limit = IBMQBackend.job_limit()
```
jyu00: It's probably better to do `backend.job_limit()` so users don't confuse it for a class method.

qiskit/providers/ibmq/ibmqbackend.py
luisg5: Very true! 😅

qiskit/providers/ibmq/backend/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Utilities related to IBMQBackend."""
```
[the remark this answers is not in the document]
luisg5: Would it okay to move `ibmqbackend.py` to the `backend` directory in a separate PR to keep the changes slightly isolated from one another. Or should I just go ahead and do it in this one?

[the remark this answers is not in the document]
jyu00: What I meant was to keep both in the base directory (`qiskit.providers.ibmq`) for now. We can decide if we want to move them based on the new services coming.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -254,6 +255,72 @@ def defaults(self, refresh: bool = False) -> Optional[PulseDefaults]:
 
         return self._defaults
 
+    def job_limit(self) -> BackendJobLimit:
+        """Return job limit for the backend.
+
+        The job limit information may include the current number of
+        jobs running on the backend at this time and the maximum number
+        of concurrent jobs a user could submit to the backend at a time.
```
jyu00: I think the first part of the sentence needs to indicate it's the user's jobs, since it may be misinterpreted as all jobs. Maybe something like 
`The job limit information may include, for this backend, the current number of unfinished jobs you have and the maximum number of unfinished jobs you can have.`

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -254,6 +255,72 @@ def defaults(self, refresh: bool = False) -> Optional[PulseDefaults]:
...
+
+        Returns:
+            Remaining number of jobs a user could submit to the backend
+            with this provider before the limit on concurrent jobs is reached.
+
+        Raises:
+            LookupError: If jobs limit for the backend can't be found.
+        """
+        backend_job_limit = self.job_limit()
+        return backend_job_limit.maximum_jobs - backend_job_limit.running_jobs
```
jyu00: `maximum_jobs` can be `-1` for unlimited or, less likely, `None`.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -254,6 +255,72 @@ def defaults(self, refresh: bool = False) -> Optional[PulseDefaults]:
...
+        api_job_limit = self._api.backend_job_limit(self.name())
+
+        try:
+            return BackendJobLimit.from_dict(api_job_limit)
+        except ValidationError as ex:
+            raise LookupError(
+                "Couldn't get backend jobs limit: {0}".format(ex))
+
+    def remaining_jobs_count(self) -> int:
+        """Return remaining jobs that could be submitted to the backend.
```
jyu00: nit: can you update the first sentence to include `the number`? since only the first line in the docstring is displayed in the summary table.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -432,6 +499,10 @@ def status(self) -> BackendStatus:
         """Return the online backend status."""
         return self._status
 
+    def job_limit(self) -> None:
+        """Return the job limits for the backend."""
+        return None
+
```
jyu00: probably want to override `remaining_jobs_count` as well

PR #516, opened 2020-01-16
Refactor job_limit

luisg5:

### DIFF
--- a/qiskit/providers/ibmq/ibmqbackend.py
+++ b/qiskit/providers/ibmq/ibmqbackend.py
@@ -35,8 +35,7 @@ from .api.exceptions import ApiError
 from .backendjoblimit import BackendJobLimit
 from .credentials import Credentials
 from .exceptions import (IBMQBackendError, IBMQBackendValueError,
-                         IBMQBackendApiError, IBMQBackendApiProtocolError,
-                         IBMQBackendJobLimitDataNotAvailable)
+                         IBMQBackendApiError, IBMQBackendApiProtocolError)
 from .job import IBMQJob
 from .utils import update_qobj_config, validate_job_tags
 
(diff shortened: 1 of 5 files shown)


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -287,29 +288,36 @@ def job_limit(self) -> BackendJobLimit:
...
-            raise LookupError(
-                "Couldn't get backend jobs limit: {0}".format(ex))
+            raise IBMQBackendApiProtocolError(
+                'Unexpected return value from the server when '
+                'querying job limit data for the backend: {}.'.format(ex))
 
     def remaining_jobs_count(self) -> Optional[int]:
         """Return the number of remaining jobs that could be submitted to the backend.
-
+F
```
jyu00: I don't think you meant to add this lol

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -287,29 +288,36 @@ def job_limit(self) -> BackendJobLimit:
...
+            IBMQBackendApiProtocolError: If an unexpected value received from the server.
+            IBMQBackendJobLimitDataNotAvailable: If the job limit information is not available.
         """
         api_job_limit = self._api.backend_job_limit(self.name())
 
         try:
-            return BackendJobLimit.from_dict(api_job_limit)
+            job_limit = BackendJobLimit.from_dict(api_job_limit)
+            if job_limit.maximum_jobs is None or job_limit.running_jobs is None:
+                raise IBMQBackendJobLimitDataNotAvailable(
```
jyu00: while we should avoid making fields in schemas as required if possible, in case of future api changes, in this case it's simpler to just make them required. Since if the fields do change in the future, this code will raise an error anyway.

qiskit/providers/ibmq/ibmqbackend.py
luisg5: I can't remember but I'm sure there is a perfectly good explanation for it... 😅. Thank goodness for Git!

qiskit/providers/ibmq/ibmqbackend.py
luisg5: Yeah, that would definitely simplify things! I'll go ahead and add that to the PR.

PR #518, opened 2020-01-20
Change BackendJobLimit attribute name

luisg5:
### Details and comments

Change the attribute `running_jobs` in `BackendJobLimit` to `active_jobs`.


jyu00:
nit - can you change `unfinished jobs` to `active jobs` in the docstrings? Just to be more consistent.

Issue #517, opened 2020-01-20
Support backend.active_jobs()


jyu00:
### What is the expected behavior?
Support querying a backend for the jobs currently in-route:
```
backend.active_jobs() -> [Job0, Job1, ...]
```
The method returns a list of job objects corresponding to the current, unfinished jobs from that provider.

PR #521, opened 2020-01-21
Add backend active jobs

luisg5:
### Summary

Addresses #517 

### Details and comments

TODO:
- Might be worth adding a test to check if the db filter is concatenated when using jobs tags.
- Also, review one more time to see if everything is good.
- Consider whether the `status` in db_filter should be removed if both status argument and status in db_filter are specified.


qiskit/providers/ibmq/ibmqbackendservice.py
```diff
@@ -275,6 +255,69 @@ def jobs(
...
+            The status db filter used to query the api in order to search
+            for jobs based on multiple statuses.
+
+        Raises:
+            IBMQBackendError: If any of the status values is not recognized.
+        """
+        statuses_filter = {'or': []}  # type: Dict[str, List[Dict[str, Any]]]
+        for _status in statuses:
+            _status = self._get_status_filter(_status)
+            statuses_filter['or'].append({'and': [_status]})
```
jyu00: Not sure if it's actually easier, but instead of concatenating a bunch of `and`, you can just do `"inq": [status1, status2...]`. I don't think it works with regular express though (hence might not be easier).

luisg5: Attempted to see if `inq` would support regular expressions and it doesn't seem to be the case. Wasn't able to find much with documentation on whether it is supported either. The main reason I chose to concatenate the statuses together was to maintain the relationship between certain filters. For example:
```
{'status': ApiJobStatus.RUNNING.value, 'infoQueue': {'exists': False}}
```
requires both to be true.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -368,9 +370,14 @@ def jobs(
         Args:
             limit: number of jobs to retrieve.
             skip: starting index for the job retrieval.
-            status: only get jobs
-                with this status, where status is e.g. `JobStatus.RUNNING` or
-                `'RUNNING'`
+            status: only get jobs with a specific status or jobs with a status that matches
+                any of the given list of statuses.
```
jyu00: This sentence is a bit hard to read. How about
```
status: only get jobs with this status or one of the statuses. 
For example, you can specify `status=JobStatus.RUNNING` or `status="RUNNING"` or `status=["RUNNING", "ERROR"]
```

qiskit/providers/ibmq/ibmqbackendservice.py
```diff
@@ -98,7 +98,7 @@ def jobs(
             limit: int = 10,
             skip: int = 0,
             backend_name: Optional[str] = None,
-            status: Optional[Union[JobStatus, str]] = None,
+            status: Optional[Union[JobStatus, str, List[Union[JobStatus, str]]]] = None,
```
jyu00: You should update the docstring here too

test/ibmq/test_ibmq_job.py
```diff
@@ -296,13 +297,55 @@ def test_retrieve_job_error(self, provider):
...
-        for job in job_list:
+        # Get the most recent jobs that are done.
+        backend_jobs = backend.jobs(limit=10, skip=0, status=JobStatus.DONE)
+        backend_jobs_status_filtered = backend.jobs(limit=10, skip=0, status=[JobStatus.DONE])
+        backend_jobs_str_filtered = backend.jobs(limit=10, skip=0, status=['DONE'])
+        self.assertTrue(backend_jobs)
+        self.assertTrue(backend_jobs_status_filtered)
+        self.assertTrue(backend_jobs_str_filtered)
+        self.assertEqual(backend_jobs, backend_jobs_status_filtered)
+        self.assertEqual(backend_jobs, backend_jobs_str_filtered)
```
jyu00: Can you convert these into sub tests? Like
```
status_args = [JobStatus.DONE, 'DONE', [JobStatus.DONE], ['DONE']]
for arg in status_args:
    with self.subTest(arg=arg):
```

test/ibmq/test_ibmq_job.py
```diff
@@ -296,13 +297,55 @@ def test_retrieve_job_error(self, provider):
...
+        """Test retrieving jobs filtered by all possible job statuses."""
+        backend = provider.get_backend('ibmq_qasm_simulator')
+        # pylint: disable=unnecessary-comprehension
+        all_job_statuses = [status for status in JobStatus]
+        # Get the most recent jobs with any job status.
+        backend_jobs_filtered = backend.jobs(limit=10, status=all_job_statuses)
+        backend_jobs = backend.jobs(limit=10)
+        self.assertTrue(backend_jobs)
+        self.assertTrue(backend_jobs_filtered)
+        self.assertEqual(backend_jobs, backend_jobs_filtered)
```
jyu00: This assert will fail if another test instance submitted another job in between the two `jobs()` calls.

test/ibmq/test_ibmq_job.py
```diff
@@ -296,13 +297,55 @@ def test_retrieve_job_error(self, provider):
...
+    def test_retrieve_multiple_job_statuses(self, provider):
+        """Test retrieving jobs filtered by multiple job statuses."""
+        backend = provider.get_backend('ibmq_qasm_simulator')
+        # Get the most recent jobs that failed or were cancelled.
+        job_statuses_to_filter = [JobStatus.ERROR, JobStatus.CANCELLED]
+        backend_jobs_filtered = backend.jobs(limit=10, status=job_statuses_to_filter)
+        self.assertTrue(backend_jobs_filtered)
+
+        for job in backend_jobs_filtered:
+            self.assertTrue(job.status() in job_statuses_to_filter)
```
jyu00: Can you print the job id if the assert fails? It'd help debugging

test/ibmq/test_ibmq_job.py
```diff
@@ -296,13 +297,55 @@ def test_retrieve_job_error(self, provider):
...
+
+    @requires_provider
+    def test_retrieve_active_jobs(self, provider):
+        """Test retrieving jobs that are currently unfinished."""
+        backend = provider.get_backend('ibmq_qasm_simulator')
+        active_job_states = [status for status in ApiJobStatus
+                             if status not in API_JOB_FINAL_STATES]
+        active_jobs = backend.active_jobs()
```
jyu00: It might be worth adding another test/subTest that ensures a `QUEUED` job is returned in `active_jobs()` (by submitting it to the busiest device) since with the simulator the jobs are more likely to be in `RUNNING` state.

qiskit/providers/ibmq/ibmqbackendservice.py
```diff
@@ -275,6 +255,76 @@ def jobs(
...
+        of the given list of statuses.
+
+        Returns:
+            The status db filter used to query the api when searching for jobs that
+            match any of the given list of statuses.
+
+        Raises:
+            IBMQBackendError: If any of the status values is not recognized.
+        """
+        statuses_filter = {'or': []}  # type: Dict[str, List[Dict[str, Any]]]
```
jyu00: does this mean if the user has `or` specified in `db_filter` it'd essentially be discarded? I don't think we want that.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -411,6 +413,23 @@ def jobs(
             limit, skip, self.name(), status,
             job_name, start_datetime, end_datetime, job_tags, job_tags_operator, db_filter)
 
+    def active_jobs(self, limit: int = 10) -> List[IBMQJob]:
+        """Return the current, unfinished jobs submitted to this backend.
+
+        Return the jobs submitted to this backend with this provider that are
+        currently in an unfinished status, including: "INITIALIZING", "VALIDATING",
+        "QUEUED", and "RUNNING".
+
```
jyu00: Can you add docstring for `limit`?

qiskit/providers/ibmq/ibmqbackendservice.py
```diff
@@ -121,8 +121,9 @@ def jobs(
             limit: number of jobs to retrieve. Default: 10.
             skip: starting index for the job retrieval. Default: 0.
             backend_name: name of the backend. Default: None.
-            status: only get jobs with this status, where status is e.g.
-                `JobStatus.RUNNING` or `'RUNNING'`. Default: None.
+            status: only get jobs with this status or one of the statuses.
+                For example, you can specify `status=JobStatus.RUNNING` or `status="RUNNING"`
+                    or `status=["RUNNING", "ERROR"]
```
jyu00: You took out the `Default: None.` part...

test/ibmq/test_ibmq_job.py
```diff
@@ -289,12 +292,87 @@ def test_retrieve_job_error(self, provider):
...
+        # pylint: disable=unnecessary-comprehension
+        all_job_statuses = [status for status in JobStatus]
+        # Get the most recent jobs with any job status.
+        backend_jobs_filtered = backend.jobs(limit=10, status=all_job_statuses)
+        backend_jobs = backend.jobs(limit=10)
+        self.assertTrue(backend_jobs)
+        self.assertTrue(backend_jobs_filtered)
+
+    @requires_provider
+    def test_retrieve_multiple_job_statuses(self, provider):
```
jyu00: I think the multi-status filtering (this and the next tests) would be much more valuable if we can _confirm_ they return jobs in all of the specified statuses. For example, if only `CANCELLED` jobs are returned, it's hard to tell if no `ERROR` jobs were found or if the query was constructed incorrectly. To accomplish this, you would need to actually create a cancelled and a failed jobs. It might be easier to combine these two tests (using `subTest`) so you only need to create those jobs once.

test/ibmq/test_ibmq_job.py
```diff
@@ -289,12 +292,87 @@ def test_retrieve_job_error(self, provider):
...
+                    self.assertTrue(job.status() is JobStatus.DONE,
+                                    "job status for job {} was '{}' but "
+                                    "it should be: '{}'"
+                                    .format(job.job_id(), job.status(), JobStatus.DONE))
 
-        self.assertTrue(job_list)
-        for job in job_list:
-            self.assertTrue(job.status() is JobStatus.DONE)
+    @requires_provider
+    def test_retrieve_all_job_statuses(self, provider):
```
jyu00: Honestly I don't see much value in this test case. All it shows is that _some_ jobs were returned from `.jobs()` (which is covered by other test cases). We can't really verify the content of the returned list.

luisg5: Yeah, I could see that. It seemed more relevant when we were comparing the two job lists retrieved, but then again you mentioned the chance of them being different if a job was submitted between the two `jobs` calls. I'll just remove it, since it would be very difficult to verify it as well.

luisg5:
Pending: Will be updating the changelog and seeing if docstrings could be improved

test/ibmq/test_ibmq_job.py
jyu00: The existing code says the argument filters win:
```
        if db_filter:
            # Argument filters takes precedence over db_filter for same keys
            api_filter = {**db_filter, **api_filter}
```

test/ibmq/test_ibmq_job.py
```diff
@@ -289,12 +291,109 @@ def test_retrieve_job_error(self, provider):
...
+        """Test retrieving jobs that are currently unfinished."""
+        backend = max([b for b in provider.backends(simulator=False) if b.status().operational],
+                      key=lambda b: b.status().pending_jobs)
+        active_job_statuses = {api_status_to_job_status(status) for status in ApiJobStatus
+                               if status not in API_JOB_FINAL_STATES}
+
+        qobj = assemble(transpile(self._qc, backend=backend), backend=backend)
+        job = backend.run(qobj)
+        # Wait for the job to queue, run, or finish.
+        while job.status() not in list(JOB_FINAL_STATES) + [JobStatus.QUEUED, JobStatus.RUNNING]:
```
jyu00: Why do we need to wait for job to be in one of these states? It's "active" as long as it hasn't finished.

test/ibmq/test_ibmq_job.py
```diff
@@ -289,12 +291,109 @@ def test_retrieve_job_error(self, provider):
...
+            time.sleep(0.5)
+
+        before_status = job.status()
+        active_jobs = backend.active_jobs()
+        if before_status is JobStatus.QUEUED and job.status() is JobStatus.QUEUED:
```
jyu00: Similar to above, the job doesn't have to be QUEUED, it just needs not be in final state. There's also no need to check for before status.

PR #525, opened 2020-01-24
Filter queued status

luisg5:
### Summary

Adresses #465


test/ibmq/test_ibmq_job.py
```diff
@@ -296,6 +296,34 @@ def test_retrieve_jobs_status(self, provider):
         for job in job_list:
             self.assertTrue(job.status() is JobStatus.DONE)
 
+    @skip('Enable when api fixes job statuses')
+    @requires_provider
+    def test_retrieve_jobs_queued(self, provider):
+        """Test retrieving jobs that are queued."""
+        backend = max([b for b in provider.backends(simulator=False) if b.status().operational],
+                      key=lambda b: b.status().pending_jobs)
```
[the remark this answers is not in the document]
luisg5: Since the main purpose of adding this function is for testing purposes, would placing it into `ibmq.utils.utils` be best? Another possible alternative was to place it in `ibmq.__init__` along with `least_busy`, but I did not add it there since the function was added for testing and not users per se...

test/ibmq/test_ibmq_job.py
```diff
@@ -296,6 +296,34 @@ def test_retrieve_jobs_status(self, provider):
...
+            time.sleep(0.5)
+
+        # Get the five most recent queued jobs.
+        job_list_queued = backend.jobs(status=JobStatus.QUEUED, limit=5)
+        for queued_job in job_list_queued:
+            self.assertTrue(queued_job._status == JobStatus.QUEUED,
+                            "status for job {} is '{}' but it should be {}"
+                            .format(queued_job.job_id(), queued_job._status,
+                                    JobStatus.QUEUED))
+
```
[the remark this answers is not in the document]
luisg5: Yeah, that would definitely be a sane check! I went ahead and added the in. Thanks for the suggestion! 👍

test/ibmq/test_ibmq_job.py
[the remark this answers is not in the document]
jyu00: Oh I meant in a new `test/utils.py`, not in the source code package.

qiskit/providers/ibmq/utils/utils.py
```diff
@@ -56,3 +58,24 @@ def validate_job_tags(job_tags: Optional[List[str]], exception: Type[Exception])
     if job_tags and (not isinstance(job_tags, list) or
                      not all(isinstance(tag, str) for tag in job_tags)):
         raise exception("job_tags needs to be a list or strings.")
+
+
+def most_busy(backends: List['IBMQBackend']) -> 'IBMQBackend':  # type: ignore[name-defined]
```
jyu00: Per my other comment, this should be under the test directory, and I'd prefer a more explicit name like `most_busy_backend`, with `provider` being the argument.

test/ibmq/test_filter_backends.py
```diff
@@ -71,3 +72,10 @@ def test_filter_least_busy(self, provider):
         backends = provider.backends()
         filtered_backends = least_busy(backends)
         self.assertTrue(filtered_backends)
+
+    @requires_provider
+    def test_filter_most_busy(self, provider):
```
jyu00: No need to test `most_busy` once you move it :)

test/ibmq/test_ibmq_job.py
```diff
@@ -296,6 +295,41 @@ def test_retrieve_jobs_status(self, provider):
...
+        backend = most_busy(provider.backends(simulator=False))
+        qobj = assemble(transpile(self._qc, backend=backend), backend=backend)
+        job = backend.run(qobj)
+
+        # Wait for the job to queue, run, or reach a final state.
+        leave_states = list(JOB_FINAL_STATES) + [JobStatus.QUEUED, JobStatus.RUNNING]
+        while job.status() not in leave_states:
+            time.sleep(0.5)
+
+        before_status = job.status()
```
jyu00: nit - you can use `job._status` here since you just called `job.status()` above. While it's not much, it saves 1 trip to the api and therefore reduces the chance the job may transition to RUNNING.

qiskit/providers/ibmq/utils/utils.py
luisg5: Ah! That'd be more consistent with its use case, being that we only use it in test. Went ahead and moved that.

Initially, I passed in the list of backends, since that would give us more flexibility with the specific backend we want to get (has more than `n` qubits, is/not is a simulator, etc). But, now that I think of it, we won't really need to specific backends for our use case. I'll make those changes as well!

test/utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Raises:
+        QiskitError: if passing a list of backend names that is
+            either empty or none have attribute ``pending_jobs``
+    """
+    try:
+        return max([b for b in provider.backends(simulator=False) if b.status().operational],
+                   key=lambda b: b.status().pending_jobs)
+    except (ValueError, TypeError):
+        raise QiskitError("Can only find most busy backend from a non-empty list.")
```
jyu00: nit - the error message `Can only find most busy backend from a non-empty list.` is not very useful and likely untrue. Since this is for test only probably don't even need the `except`.

PR #535, opened 2020-02-04
Start of visualization migration / new stuff

nonhermitian:
### Summary
Adds the interactive backend error map, gate_map, and new html repr for backends, as related functions and docs.

The `plotly` wrappers are no longer necessary when they become part of terra: https://github.com/Qiskit/qiskit-terra/pull/3786

### Details and comments

Screen Shot 2020-02-05 at 1 53 27 PM


nonhermitian:
Note that we can take advantage of the docs building to test these routines.

PR #538, opened 2020-02-05
Normalize proxy urls

### LINKED ISSUE #528: Handling of malformed proxy configuration

On a run with a newer version of Python, four of the proxy tests were failing with:
```
urllib3.exceptions.ProxySchemeUnknown: Not supported proxy scheme
```

```
test/ibmq/test_proxies.py::TestProxies::test_invalid_proxy_address_authclient FAILED
test/ibmq/test_proxies.py::TestProxies::test_invalid_proxy_address_versionclient FAILED
test/ibmq/test_proxies.py::TestProxies::test_invalid_proxy_port_authclient FAILED
test/ibmq/test_proxies.py::TestProxies::test_invalid_proxy_port_versionclient FAILED
````

It seems related to the way `INVALID_PORT_PROXIES` and `INVALID_ADDRESS_PROXIES` are defined (ie. lacking a `http://` prefix in the value), and due to recent changes In Python (more details at https://github.com/psf/requests/issues/5297).

It might be just a matter of tweaking the tests and keeping track of the upstream issue - however, it might be possible that users who had previously a working-but-now-invalid proxy configuration encounter the same issue, which is a bit obscure to track. Perhaps adding an extra check on our side or a similar measure might be in order?

### PR DESCRIPTION
**Pull request by jyu00**
### Summary
Closes #528. 
It adds `http://` to proxy urls without one and issues a deprecation warning that it's no longer supported.

### DIFF
--- a/qiskit/providers/ibmq/credentials/credentials.py
+++ b/qiskit/providers/ibmq/credentials/credentials.py
@@ -15,6 +15,7 @@
 """Model for representing IBM Q credentials."""
 
 import re
+import warnings
 
 from typing import Dict, Tuple, Optional, Any
 from requests_ntlm import HttpNtlmAuth
@@ -78,6 +79,9 @@ class Credentials:
         self.proxies = proxies or {}
         self.verify = verify
 
+        # Normalize proxy urls.
+        self._prepend_protocol_if_needed()
+
     def is_ibmq(self) -> bool:
         """Return whether the credentials represent a IBMQ account."""
         return all([self.hub, self.group, self.project])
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: qiskit/providers/ibmq/credentials/credentials.py [hunk 1]
```diff
@@ -119,6 +123,26 @@ def connection_parameters(self) -> Dict[str, Any]:
...
         return request_kwargs
 
+    def _prepend_protocol_if_needed(self) -> None:
+        """Prepend the proxy URLs with protocol if needed."""
+        if 'urls' not in self.proxies:
+            return
+
+        warning_issued = False
+        for key, url in self.proxies['urls'].items():
+            colon_pos = url.find(':')
```
diego-plan9: Non blocking, as I did not dig deep enough into the block - but would some of the utils in https://docs.python.org/3/library/urllib.parse.html help making the logic a bit more robust? Even if it seems they are part of the upstream issue, on 3.7.5 it seems possible to provide a default `scheme` gracefully-ish:

```python
In [20]: urlparse('1.2.3.4')                                                                                                        
Out[20]: ParseResult(scheme='', netloc='', path='1.2.3.4', params='', query='', fragment='')

In [21]: urlparse('1.2.3.4', scheme='http')                                                                                         
Out[21]: ParseResult(scheme='http', netloc='', path='1.2.3.4', params='', query='', fragment='')

In [22]: urlparse('lol://1.2.3.4', scheme='http')                                                                                   
Out[22]: ParseResult(scheme='lol', netloc='1.2.3.4', path='', params='', query='', fragment='')
```

jyu00: `urlparse()` on the problematic versions might return the IP address as scheme:
```
>>> urlparse('1.2.3.4:80', 'http')
ParseResult(scheme='1.2.3.4', netloc='', path='80', params='', query='', fragment='')
```
I suppose I can check if the returned scheme is one of `http`, `https`, or `socks5`, since they're the only supported protocols, even though digits and dot are technically valid scheme characters.

jyu00: On a second thought, I'll just stick to the original solution of finding a colon. While extremely unlikely, someone could in theory have `socks5` as their proxy hostname.

PR #541, opened 2020-02-11
Add IQX Dashboard

nonhermitian:
### Summary
Adds the IQX dashboard.  Usage is:
```python
from qiskit import IBMQ
import qiskit.providers.ibmq.jupyter

IBMQ.load_account()
%iqx_dashboard
```

fixes https://github.com/Qiskit/qiskit-terra/issues/3770


qiskit/providers/ibmq/jupyter/dashboard/dashboard.py
```diff
@@ -0,0 +1,312 @@
...
+        do_pop = False
+        ind = None
+        for idx, job in enumerate(self.jobs):
+            if job.job_id == job_id:
+                do_pop = True
+                ind = idx
+                break
+        if not do_pop:
+            raise Exception('job_id not found')
+        if 'CANCELLED' not in self.jobs[ind].children[3].value:
```
jyu00: Shouldn't this check for DONE and ERROR in addition to CANCELLED? Not that it matters much since the `cancel()` will just fail.

qiskit/providers/ibmq/jupyter/dashboard/dashboard.py
```diff
@@ -0,0 +1,312 @@
...
+            self.dashboard._thread.do_run = False
+            self.dashboard._thread.join()
+            self.dashboard.close()
+        self.dashboard = None
+
+    def update_single_job(self, update_info: Tuple):
+        """Update a single job instance
+
+        Args:
+            update_info (tuple): Updated job info.
```
jyu00: Can you document what the expected elements are in this tuple? Or better yet, use a helper class or individual parameters so one doesn't need to guess what's in each index location.

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
```diff
@@ -0,0 +1,82 @@
...
+        status (Enum): Job status.
+        watcher (JobWatcher): Job watcher instance
+
+    """
+    prev_status_name = None
+    prev_queue_pos = None
+    interval = 2
+    exception_count = 0
+
+    while status.name not in ['DONE', 'CANCELLED', 'ERROR']:
```
jyu00: There is a `JOB_FINAL_STATES` in `qiskit.providers.jobstatus` that consists of these final states.

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
```diff
@@ -0,0 +1,82 @@
...
+        time.sleep(interval)
+        try:
+            status = job.status()
```
jyu00: Instead of doing `job.status()` in a loop, you can now use `wait_for_final_state()`, which uses websockets and thus generates far less traffic to the API. This can be in a separate PR though.

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
```diff
@@ -0,0 +1,82 @@
...
+            exception_count = 0
+
+            if status.name == 'QUEUED':
+                queue_pos = job.queue_position()
+                if queue_pos != prev_queue_pos:
+                    est_time = utc_to_local(job.queue_info().estimated_start_time)
```
jyu00: It'd be nice to use the same humanized format like IQX does (e.g. "in 2 hours")

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
```diff
@@ -0,0 +1,82 @@
...
+
+                    update_info = (job.job_id(), status.name+' ({})'.format(queue_pos),
+                                   est_time.strftime("%H:%M %Z (%m/%d)"), status.value)
```
jyu00: Doesn't `update_single_job` expects the 3rd element to be the queue position?

qiskit/providers/ibmq/jupyter/dashboard/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+==================================
+Dashboard (:mod:`theia.dashboard`)
```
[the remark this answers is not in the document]
nonhermitian: Was left over from testing package.  Fixed

qiskit/providers/ibmq/jupyter/dashboard/backend_update.py
```diff
@@ -0,0 +1,59 @@
...
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""A module for the async backend widget updates"""
+import time
+import threading
+
+
+def update_backend_info(device_list, interval=30):
+    """Updates the device list from another thread
```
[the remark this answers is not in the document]
nonhermitian: This is already done.

qiskit/providers/ibmq/jupyter/dashboard/backend_update.py
```diff
@@ -0,0 +1,61 @@
...
+                    stat_msg = status.status_msg
+                    pending = str(status.pending_jobs)
+
+                    color = '#000000'
+                    if stat_msg == 'active':
+                        color = '#34bc6e'
+                    if stat_msg in ['maintenance', 'internal']:
+                        color = '#FFB000'
+
+                    status_wid = backend.children[0].children[1].children[0].children[1].children[1]
```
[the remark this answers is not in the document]
nonhermitian: Added a brief comment.  However, this type of statement is quite common when dealing with the widgets where you need to grab a widget inside a container of multiple nested widgets.  Due to the usage of lists when adding children to a parent widget.

qiskit/providers/ibmq/jupyter/dashboard/dashboard.py
```diff
@@ -0,0 +1,312 @@
...
+        self._init_subscriber()
+        self.dashboard: Accordion_with_thread = None
+        self.backend_dict: Dict = None
+        self.job_viewer: wid.VBox = None
+        self._clear_jobs_button: wid.GridBox = make_clear_button(self)
+        self._jobs_labels: wid.HBox = make_labels()
+        self.refresh_jobs_board()
+
+    def _get_backends(self):
+        from qiskit import IBMQ  # pylint: disable=import-outside-toplevel
```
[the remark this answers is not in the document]
nonhermitian: yep, done

qiskit/providers/ibmq/jupyter/dashboard/dashboard.py
nonhermitian: This is just trying to cancel a job after you clicked the cancel button.

qiskit/providers/ibmq/jupyter/dashboard/dashboard.py
nonhermitian: done.

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
nonhermitian: Yes, but this uses the enum which is ugly in my opinion, and we said we would move to a string anyway.

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
nonhermitian: I am not waiting for the final state here, but rather am looking for changes in state so as to update the job monitor accordingly.

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
nonhermitian: This is humanized in the sense that it gives up the time the job is supposed to start in local time, eg 14:45 EST.  It is also of fixed width verses something that displays durations that may or may not include things like days, hours, minutes.

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
nonhermitian: Nope, but that should be clarified in the docs now.  The queue position is just part of the string in arg[1].

qiskit/providers/ibmq/jupyter/dashboard/backend_update.py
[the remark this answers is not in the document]
jyu00: Not entirely. There are still `Parameters` instead of `Args`. One-sentence comment that spans 2 lines, etc.

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
jyu00: The `while` on line 48 says it is checking for job status until a final state is reached. You can supply a callback function to `wait_for_final_state()` that updates the job monitor as the status changes.

qiskit/providers/ibmq/jupyter/dashboard/backend_update.py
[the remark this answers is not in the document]
nonhermitian: Args is shorthand for parameters and gets converted as such by sphinx.

qiskit/providers/ibmq/jupyter/dashboard/dashboard.py
jyu00: But presumably you're doing `if 'CANCELLED` because canceling is moot if the job has already been cancelled? That applies to `DONE` and `ERROR` as well.

nonhermitian: Ok yes, I see what your saying now.

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
nonhermitian: Wait for the final state does not do an adaptive time interval like the current checker does. Also, why does job.status not use websockets?

jyu00: True, I should add that feature (not for this PR but later). 
`job.status()` doesn't use websockets because it returns immediately, which defeats the purpose of using a long running websocket.

nonhermitian: `job.status` and `job.wait_for_final_state` both call `job._get_status_position` with the same arguments to get the running status.  How is it that the latter uses websockets, while the former does not?

jyu00: `job._get_status_position()` is used to interpret the API response, not to communicate with the API. `job.status()` calls `._api.job_status()`, which does a 1 time call to the API that returns immediately. `job.wait_for_final_state()` calls `._api.job_final_status()`, which keeps the connection open (via websockets) until the job is done.

nonhermitian: So another issue with the callback is that it does not keep state from one call to the next.  So things like changing HTML status fields only if the status has changed cannot be supported because there is no way to keep track of the status from one callback to the next.

jyu00: Yeah as mentioned before, invoking the callback only if something changed (status or queue position etc) would be a nice feature to add to `wait_for_final_state`. But that'd be a separate PR. The job watcher will just have to continue polling (unless you can extract the previous state from the widget somehow, but updating `wait_for_final_state` seems to be the better solution anyway).

nonhermitian: Yes, but as it stands now I could be doing a ton of widget refreshing for no reason because the callback does not support it.  I prefer to wait for the callback to be more generalized before moving all these routines which have been tested for many months now, over to it.  But I agree it should be done.

PR #542, opened 2020-02-12
Allow autodoc with sphinx

jyu00:

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,10 +1,9 @@
-##################################
-Qiskit IBM Q Account documentation
-##################################
+####################################
+Qiskit IBM Q Provider documentation
+####################################
 
 .. toctree::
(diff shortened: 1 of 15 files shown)


docs/index.rst
```diff
@@ -1,10 +1,9 @@
-##################################
-Qiskit IBM Q Account documentation
-##################################
+####################################
+Qiskit IBM Q Provider documentation
```
abbycross: I think it is now IBM Quantum Provider -- @nonhermitian do you know?

jyu00: I debated about this. I thought since it's `ibmq-provider`, it should be `IBM Q Provider`. But I'm open to suggestions.

nonhermitian: I would like to call it IQX provider, but lets leave it as IBM Quantum for now.

PR #543, opened 2020-02-12
Update branding

abbycross:
### Summary

Update IBM Q to IBM Quantum where appropriate 

### Details and comments

I did not change instances of IBMQ (with no space between M and Q).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Qiskit IBMQ Provider
+# Qiskit IBM Quantum Provider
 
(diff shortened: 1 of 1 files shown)


jyu00:
hi @abbycross , thanks for making the updates! A few places that say `IBM Quantum ...` should be `IBM Quantum Experience`, since we're moving towards calling the services as "IQX", whereas "IBM Quantum" is just the group.

abbycross:
Hi @jyu00  -- should I change all IBM Quantum to IBM Quantum Experience?  E.g., IBM Quantum Experience Provider?  The last I heard about using IQX is that it's for internal use only, but it would be great if we were allowed to use it in these cases.

jyu00:
> Hi @jyu00 -- should I change all IBM Quantum to IBM Quantum Experience? E.g., IBM Quantum Experience Provider? The last I heard about using IQX is that it's for internal use only, but it would be great if we were allowed to use it in these cases.

@abbycross  Looking closer to the docs, I think we can use "IBM Quantum Provider" when referencing `qiskit-ibmq-provider`, and "IBM Quantum Experience" otherwise. I don't know if we can use IQX yet, but "IBM Quantum Experience" is fair game.

PR #567, opened 2020-02-26
Move Job monitor from Terra and add est wait time

nonhermitian:
### Summary
Moves the `job_monitor` from Terra into the provider and adds an estimated wait time when the device is queued.


qiskit/providers/ibmq/job/job_monitor.py
```diff
@@ -0,0 +1,124 @@
...
+                  _interval_set: bool = False,
+                  quiet: bool = False,
+                  output: TextIO = sys.stdout) -> None:
+    """A text-based job status checker.
+
+    Args:
+        job: The job to check.
+        interval: The interval at which to check.
+        _interval_set (bool): Was interval time set by user?
+        quiet (bool): If True, do not print status messages.
```
jyu00: Why is this parameter necessary? If `quiet=True`, then this function is essentially the same as `wait_for_final_state` as it just waits for the job to finish without printing anything.

qiskit/providers/ibmq/job/job_monitor.py
```diff
@@ -0,0 +1,124 @@
...
+        time.sleep(interval)
+        status = job.status()
+        msg = status.value
+
+        if status.name == 'QUEUED':
+            queue_info = job.queue_info()
+
+            time_str = ''
+            if queue_info:
+                est_time = queue_info.estimated_start_time
```
jyu00: Unfortunately `estimated_start_time` can be `None` even if `queue_info` itself isn't, since the API might return only _some_ of the queue information.

qiskit/providers/ibmq/job/job_monitor.py
```diff
@@ -0,0 +1,124 @@
...
+                prev_time_str = time_str
+
+            else:
+                time_str = prev_time_str
+
+            msg += ' ({queue}) [Est. wait time: {time}]'.format(queue=job.queue_position(),
+                                                                time=time_str)
+
+            if job.queue_position() is None:
+                interval = 2
```
jyu00: user-specified interval is not used if `job.queue_position()` is None?

qiskit/providers/ibmq/job/job_monitor.py
nonhermitian: Quiet was added here: https://github.com/Qiskit/qiskit-terra/pull/2037.  I agree it is not needed anymore.

qiskit/providers/ibmq/job/job_monitor.py
nonhermitian: Fixed by doing a check.

qiskit/providers/ibmq/job/job_monitor.py
nonhermitian: Good find.  Fixed.

nonhermitian:
Also added error message at output if the job errored.

PR #572, opened 2020-03-02
Cleanup Exceptions

luisg5:
### Summary

Addresses #106

### Details and comments

TODO: Update changelog

The manner in which exceptions are used throughout the provider appear to be done so consistently, and the practices mentioned in the original issue also appear to be in place. However, there are parts of the provider that could be cleaned up. This PR aims to cleanup some of the exceptions by:

- Improving the exception messages on those that lacked detail.
- Adding the Job ID to the messages that would benefit from it.
- Adding a new `VisualizationError` exception for the visualization modules (for consistency).
- Changing `IBM Q Experience` -> `IBM Quantum Experience`.


qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -749,7 +760,8 @@ def _get_status_position(
                 if queue_info._status == ApiJobStatus.PENDING_IN_QUEUE.value:
                     status = JobStatus.QUEUED
         except (KeyError, ValidationError) as ex:
-            raise IBMQJobApiError("Unexpected return value received from the server.") from ex
+            raise IBMQJobApiError('Unexpected return value received from the server when '
+                                  'getting status for job {}: {}'.format(self.job_id(), str(ex)))
```
jyu00: why get rid of `from ex`? it seems more helpful to say exception A was caused by B.

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -194,11 +195,12 @@ def save_account(
         """
         if url != QX_AUTH_URL:
             raise IBMQAccountCredentialsInvalidUrl(
-                'Invalid IBM Q Experience credentials found. ' + UPDATE_ACCOUNT_TEXT)
+                'The URL specified ({}) is not an IBM Quantum Experience authentication '
+                'URL. Valid authentication URL: {}.'.format(url, QX_AUTH_URL))
```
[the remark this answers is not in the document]
luisg5: In this case, the user trying to save an account to disk, with a specific URL. I thought it would be best to point them to the correct URL to use if they are using the incorrect one. But, there is the possibility of using an old url, in which case it's best to point them towards how to update their account. So, I'll switch it back! 👍

qiskit/providers/ibmq/api/clients/websocket.py
```diff
@@ -112,8 +112,9 @@ def from_bytes(cls, json_string: bytes) -> 'WebsocketResponseMethod':
         """Instantiate a message from a bytes response."""
         try:
             parsed_dict = json.loads(json_string.decode('utf8'))
-        except (ValueError, AttributeError) as ex:
-            raise WebsocketIBMQProtocolError('Unable to parse message') from ex
+        except (ValueError, AttributeError):
+            raise WebsocketIBMQProtocolError('Unable to parse the message received from '
+                                             'the server: {}'.format(json_string))
```
jyu00: Another place that has `from ex` removed.

qiskit/providers/ibmq/ibmqbackendservice.py
```diff
@@ -369,7 +369,7 @@ def retrieve_job(self, job_id: str) -> IBMQJob:
         try:
             job_info = self._provider._api.job_get(job_id)
         except ApiError as ex:
-            raise IBMQBackendApiError('Failed to get job "{}": {}'
+            raise IBMQBackendApiError('Failed to get job {}: {}'
```
jyu00: This should also have `from ex`. Can you check if there are others that should be chained or not chained (`raise from None`)?

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -198,7 +199,7 @@ def save_account(
 
         if not token or not isinstance(token, str):
             raise IBMQAccountCredentialsInvalidToken(
-                'Invalid token found: "%s" %s' % (token, type(token)))
+                'Invalid IBM Quantum Experience token found: "{}" {}.'.format(token, type(token)))
```
jyu00: nit - can you change the message to `Invalid IBM Quantum Experience token found: "{}" of type {}`?

qiskit/providers/ibmq/managed/managedjobset.py
```diff
@@ -159,8 +160,8 @@ def retrieve_jobs(self, provider: AccountProvider, refresh: bool = False) -> Non
         matched = pattern.match(first_job.name())
         if not matched:
             raise IBMQJobManagerInvalidStateError(
-                "Job {} is tagged for the job set {} but does not have a proper job name.".format(
-                    first_job.job_id(), self.job_set_id()))
+                'Job {} is tagged for the job set {} but it does not '
```
jyu00: you'll need a comma before the `but` since you're connecting two sentences. Alternatively you can just use the old sentence (i.e. leaving out the `it` after the `but`, making it a dependent clause). Same for line 177.

PR #574, opened 2020-03-02
Optional package doc fix

jyu00:
### Summary
Only the visualization extra packages are required, not everything in `requirements-dev.txt`.

### DIFF
--- a/qiskit/providers/ibmq/jupyter/__init__.py
+++ b/qiskit/providers/ibmq/jupyter/__init__.py
@@ -23,8 +23,10 @@ that extend the functionality of Qiskit for the IBM
 Quantum devices.
 
 Note:
-    To use the tools locally, you'll need to install packages
-    listed in ``requirements-dev.txt``.
+    To use these tools locally, you'll need to install the
+    ``visualization`` extra packages::
+
+        pip install qiskit-ibmq-provider[visualization]
 
 Detailed information on a single backend
 ========================================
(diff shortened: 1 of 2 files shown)


luisg5:
Would it be worth having a separate requirements file, such as `requirements-visualizations.txt`? Other than making it easier for users to install the requirements, it might also make sense since not all the libraries for visualization are used for development.

jyu00:
> Would it be worth having a separate requirements file, such as `requirements-visualization.txt`? 

I gave this some thought, and right now I'm leaning towards no, for the reasons below.

>On top of making it easier for users to install the requirements, 

I think it's actually easier for a user to install from a single requirements file rather than multiple, to setup a development environment. 

> it might also make sense since not all the libraries for visualization are used for development.

Depends on what you mean by "development". Yes the libraries are only needed if you're writing the visualization modules, but that's true for the other libraries. For example the Sphinx libraries are not needed if you're not updating docs, or pproxy if you're not doing the networking modules.

luisg5:
> Depends on what you mean by "development". Yes the libraries are only needed if you're writing > the visualization modules, but that's true for the other libraries. For example the Sphinx libraries > are not needed if you're not updating docs, or pproxy if you're not doing the networking modules.

The initial thought was that the modules were optional and making it easier for the users to install them, since they would be all in one place. But, considering this case, I do agree it would be best to keep them under one single requirements file.

Issue #499, opened 2019-12-20
job.cancel() returns true but job not cancelled


jyu00:
### Information

- **Qiskit IBMQ Provider version**:
- **API URL or version**:

- **Python version**:
- **Operating system**:

### What is the current behavior?
`job.cancel()` can return `True` even if the cancel had failed.

### Steps to reproduce the problem
It cannot be reproduced at will, but show up in several Travis runs (e.g. [1425](https://travis-ci.com/Qiskit/qiskit-ibmq-provider/jobs/269353015) and [1414](https://travis-ci.com/Qiskit/qiskit-ibmq-provider/jobs/269108014))

### What is the expected behavior?
Job is cancelled if `job.cancel()` returns true.

### Suggested solutions
The provider currently checks for `error` in the response data to determine if the cancel was successful. After talking to the api team, it should also be checking for `'cancelled': True` in the response data.

jyu00:
In some specific window `cancel()` may still fail even with `cancelled: True`.

jyu00:
API rolled out a new release that had a job cancel related fix. This should be re-tested and see if we can close.

luisg5:
When a job was cancelled while in the `RUNNING` state, `cancel()` would succeed. However, after the job finished running, the api would overwrite the `CANCELLED` status to `COMPLETE`. This issue was fixed by their new release, as they do not allow overwriting the job status if it is in a final state (i.e. `CANCELLED`).

jyu00:
As mentioned, `test_skipped_result` failed with 
```
======================================================================
FAIL: test_skipped_result (test.ibmq.test_ibmq_jobmanager.TestResultManager)
Test one of jobs has no result.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/jyu00/qiskit-ibmq-provider/test/decorators.py", line 72, in _wrapper
    return func(obj, *args, **kwargs)
  File "/home/travis/build/jyu00/qiskit-ibmq-provider/test/decorators.py", line 100, in _wrapper
    return func(*args, **kwargs)
  File "/home/travis/build/jyu00/qiskit-ibmq-provider/test/ibmq/test_ibmq_jobmanager.py", line 391, in test_skipped_result
    result_manager.get_counts(max_circs)
AssertionError: IBMQManagedResultDataNotAvailable not raised : IBMQManagedResultDataNotAvailable not raised for job 5e6bdf77d9955f0019621576
```

[...]

PR #576, opened 2020-03-03
Remove safety cancel

luisg5:
### Summary

Addresses #499

### Details and comments

If a job was successfully cancelled while running, its status may become `CANCELLED` and then transition to `COMPLETED` after it finished running. This is no longer the case, since the api does not overwrite the job if it is in a final state (e.g. `CANCELLED`). So, the "todos" could be addressed by removing the safeties that were set in place.


jyu00:
You missed 2 - one in `test_ibmq_job` and one in `test_account_client`.

PR #577, opened 2020-03-04
Disable http retry on job submit

### LINKED ISSUE #486: Disable retry with HTTP
Job submit currently retries with HTTP if object storage fails. The API will soon disable job submit through HTTP, so that retry path needs to be disabled. We should instead verify the object storage path has sufficient retries.

jyu00:
### Summary
Closes #486 .


luisg5:
I don't believe [Api.job_submit](https://github.com/jyu00/qiskit-ibmq-provider/blob/607366b01a84b50c0c72b2f18f2f989baf0fcfe0/qiskit/providers/ibmq/api/rest/root.py#L112) is used anywhere else after removing http, so it could be removed?

Issue #524, opened 2020-01-24
Add more debugging messages


jyu00:
### What is the expected behavior?
When running tests in CI it is often difficult to determine why a test had failed. Some of the failures are difficult to recreate when running locally. It'd be nice to have a lot more debugging messages, such as major job transition (e.g. `Job 123 submitted to backend foo`), and endpoint used, and provide a way to toggle these messages in the CI.

PR #579, opened 2020-03-05
Improve logging

luisg5:
### Summary

Addresses #524

### Details and comments

This PR sets up a logger for the package, which allows loggers within the modules to inherit the correct level when the `QISKIT_IBMQ_PROVIDER_LOG_LEVEL` environment variable is set. Also, it allows users to specify a filename to place logs into, via the `QISKIT_IBMQ_PROVIDER_LOG_FILE` environment variable.

It also adds more debugging messages, using `Session.request` as the centralized logging placement.

An example of the log format is:
```
websocket.get_job_status:DEBUG:2020-03-17 16:21:45,632: Received message from websocket: {'hubInfo': '...', 'id': '5e70f8f18a2f2f0018e12452', 'endDate': '2020-03-17T16:21:43.196Z', 'backend': {'id': '5ae875670f020500393162ad', 'name': '...'}, 'kind': 'q-object-external-storage', 'status': 'COMPLETED', 'creationDate': '2020-03-17T16:21:05.284Z'}
```

The first part: 
`websocket.get_job_status:DEBUG:2020-03-17 16:21:45,632:`
is formatted automatically.

The second part:
`Received message from websocket: {'hubInfo': '...', 'id': '5e70f8f18a2f2f0018e12452', 'endDate': '2020-03-17T16:21:43.196Z', 'backend': {'id': '5ae875670f020500393162ad', 'name': '...'}, 'kind': 'q-object-external-storage', 'status': 'COMPLETED', 'creationDate': '2020-03-17T16:21:05.284Z'}` 
is the specified message to include in the log.

### DIFF
--- a/qiskit/providers/ibmq/__init__.py
+++ b/qiskit/providers/ibmq/__init__.py
@@ -121,8 +121,13 @@ def least_busy(backends: List[BaseBackend]) -> BaseBackend:
             does not have the ``pending_jobs`` attribute in its status.
     """
     try:
-        return min([b for b in backends if b.status().operational],
-                   key=lambda b: b.status().pending_jobs)
+        backends_to_filter = {}
+        for backend in backends:
+            backend_status = backend.status()
+            if backend_status.operational:
+                backends_to_filter[backend] = backend_status
+
+        return min(backends_to_filter.keys(), key=lambda b: backends_to_filter[b].pending_jobs)
     except (ValueError, TypeError):
         raise IBMQError('Unable to find the least_busy '
                         'backend from an empty list.') from None
--- a/qiskit/providers/ibmq/__init__.py
+++ b/qiskit/providers/ibmq/__init__.py
@@ -21,17 +21,30 @@ IBM Quantum Provider (:mod:`qiskit.providers.ibmq`)
 
 Modules representing the IBM Quantum Provider.
 
-Environment Variables
+Logging
 =====================
-The following environment variables are used by the IBM Quantum Provider:
+
+The IBM Quantum Provider uses the ``qiskit.providers.ibmq`` logger.
+
+You are able to change its level manually or via the environment variable
+``QISKIT_IBMQ_PROVIDER_LOG_LEVEL``. Likewise, you are able to specify a file
+to log the messages to with the environment variable ``QISKIT_IBMQ_PROVIDER_LOG_FILE``.
+
+To change the level manually, simply grab the logger and set its level::
+
+    import logging
+    logging.getLogger('qiskit.providers.ibmq').setLevel(logging.WARNING)
+
+Details pertaining to the environment variables used by the logger:
 
     * ``QISKIT_IBMQ_PROVIDER_LOG_LEVEL``: Specifies the log level to use, for the
-      provider modules, when logging to the screen. If an invalid level is set, the
-      log level defaults to ``WARNING``. The valid log levels are ``DEBUG``, ``INFO``,
-      ``WARNING``, ``ERROR``, and ``CRITICAL`` (case-insensitive).
-    * ``QISKIT_IBMQ_PROVIDER_LOG_FILE``: Specifies the name of the logfile to create,
-      which will contain the logs. If it is not specified, a logfile will not be created,
-      but the message will still be logged to the screen.
+      provider modules, when logging. If an invalid level is set, the log level defaults
+      to ``WARNING``. The valid log levels are ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``,
+      and ``CRITICAL`` (case-insensitive). If the environment variable is not set, then
+      the parent logger's level is used, which also defaults to `WARNING`.
+    * ``QISKIT_IBMQ_PROVIDER_LOG_FILE``: Specifies the name of the logfile to use when logging.
+      If specified, the log messages will be logged to the file but not to the screen. If it
+      is not specified, the log messages will only be logged to the screen.
 
 Functions
 =========
--- a/qiskit/providers/ibmq/__init__.py
+++ b/qiskit/providers/ibmq/__init__.py
@@ -27,10 +27,11 @@ The following environment variables are used by the IBM Quantum Provider:
 
     * ``QISKIT_IBMQ_PROVIDER_LOG_LEVEL``: Specifies the log level to use, for the
       provider modules, when logging to the screen. If an invalid level is set, the
-      log level defaults to ``INFO``. The valid log levels are ``DEBUG``, ``INFO``,
+      log level defaults to ``WARNING``. The valid log levels are ``DEBUG``, ``INFO``,
       ``WARNING``, ``ERROR``, and ``CRITICAL`` (case-insensitive).
     * ``QISKIT_IBMQ_PROVIDER_LOG_FILE``: Specifies the name of the logfile to create,
-      which will contain the logs. If it is not specified, a logfile will not be created.
+      which will contain the logs. If it is not specified, a logfile will not be created,
+      but the message will still be logged to the screen.
 
 Functions
 =========
--- a/qiskit/providers/ibmq/__init__.py
+++ b/qiskit/providers/ibmq/__init__.py
@@ -42,9 +42,9 @@ Details pertaining to the environment variables used by the logger:
       to ``WARNING``. The valid log levels are ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``,
       and ``CRITICAL`` (case-insensitive). If the environment variable is not set, then
       the parent logger's level is used, which also defaults to `WARNING`.
-    * ``QISKIT_IBMQ_PROVIDER_LOG_FILE``: Specifies the name of the logfile to use when logging.
-      If specified, the log messages will be logged to the file but not to the screen. If it
-      is not specified, the log messages will only be logged to the screen.
+    * ``QISKIT_IBMQ_PROVIDER_LOG_FILE``: Specifies the name of the logfile to use when logging
+      messages. If specified, the log messages will be logged to the file but not to the screen.
+      If it is not specified, the log messages will only be logged to the screen.
 
 Functions
 =========
--- a/qiskit/providers/ibmq/__init__.py
+++ b/qiskit/providers/ibmq/__init__.py
@@ -58,6 +58,7 @@ Exceptions
     IBMQProviderError
 """
 
+import logging
 from typing import List
 
 from .ibmqfactory import IBMQFactory
@@ -68,8 +69,14 @@ from .accountprovider import AccountProvider
 from .backendjoblimit import BackendJobLimit
 from .exceptions import *
 from .ibmqbackendservice import IBMQBackendService
+from .utils.utils import setup_logger
 
 from .version import __version__
+
+# Setup the logger for the IBM Quantum Provider package.
+logger = logging.getLogger(__name__)
+setup_logger(logger)
+
 
 IBMQ = IBMQFactory()
 """A global instance of an account manager that is used as the entry point for convenience."""
--- a/qiskit/providers/ibmq/__init__.py
+++ b/qiskit/providers/ibmq/__init__.py
@@ -102,6 +102,14 @@ setup_logger(logger)
 IBMQ = IBMQFactory()
 """A global instance of an account manager that is used as the entry point for convenience."""
 
+# Constants used by the IBM Quantum logger.
+IBMQ_PROVIDER_LOGGER_NAME = 'qiskit.providers.ibmq'
+"""The name of the IBM Quantum logger."""
+QISKIT_IBMQ_PROVIDER_LOG_LEVEL = 'QISKIT_IBMQ_PROVIDER_LOG_LEVEL'
+"""The environment variable name that is used to set the level for the IBM Quantum logger."""
+QISKIT_IBMQ_PROVIDER_LOG_FILE = 'QISKIT_IBMQ_PROVIDER_LOG_FILE'
+"""The environment variable name that is used to set the file for the IBM Quantum logger."""
+
 
 def least_busy(backends: List[BaseBackend]) -> BaseBackend:
     """Return the least busy backend from a list.
@@ -121,13 +129,8 @@ def least_busy(backends: List[BaseBackend]) -> BaseBackend:
             does not have the ``pending_jobs`` attribute in its status.
     """
     try:
-        backends_to_filter = {}
-        for backend in backends:
-            backend_status = backend.status()
-            if backend_status.operational:
-                backends_to_filter[backend] = backend_status
-
-        return min(backends_to_filter.keys(), key=lambda b: backends_to_filter[b].pending_jobs)
+        return min([b for b in backends if b.status().operational],
+                   key=lambda b: b.status().pending_jobs)
     except (ValueError, TypeError):
         raise IBMQError('Unable to find the least_busy '
                         'backend from an empty list.') from None
--- a/qiskit/providers/ibmq/__init__.py
+++ b/qiskit/providers/ibmq/__init__.py
@@ -21,6 +21,17 @@ IBM Quantum Provider (:mod:`qiskit.providers.ibmq`)
 
 Modules representing the IBM Quantum Provider.
 
+Environment Variables
+=====================
+The following environment variables are used by the IBM Quantum Provider:
+
+    * ``QISKIT_IBMQ_PROVIDER_LOG_LEVEL``: Specifies the log level to use, for the
+      provider modules, when logging to the screen. If an invalid level is set, the
+      log level defaults to ``INFO``. The valid log levels are ``DEBUG``, ``INFO``,
+      ``WARNING``, ``ERROR``, and ``CRITICAL`` (case-insensitive).
+    * ``QISKIT_IBMQ_PROVIDER_LOG_FILE``: Specifies the name of the logfile to create,
+      which will contain the logs. If it is not specified, a logfile will not be created.
+
 Functions
 =========
 .. autosummary::
--- a/qiskit/providers/ibmq/__init__.py
+++ b/qiskit/providers/ibmq/__init__.py
@@ -26,25 +26,22 @@ Logging
 
 The IBM Quantum Provider uses the ``qiskit.providers.ibmq`` logger.
 
-You are able to change its level manually or via the environment variable
-``QISKIT_IBMQ_PROVIDER_LOG_LEVEL``. Likewise, you are able to specify a file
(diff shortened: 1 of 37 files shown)


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -213,11 +213,14 @@ def _submit_job(
                 job_share_level=job_share_level,
                 job_tags=job_tags)
         except ApiError as ex:
+            logger.debug('Error: There was an error submitting the job: %s.', str(ex))
```
jyu00: I don't really see the value of logging the error when it's already being raised...

qiskit/providers/ibmq/utils/utils.py
```diff
@@ -56,3 +58,31 @@ def validate_job_tags(job_tags: Optional[List[str]], exception: Type[Exception])
...
+    It involves:
+        * Use the `LOG_LEVEL` environment variable to determine the log level
+            for the provider. If it is not set, or is set to an invalid level,
+            the log level is defaulted to `INFO` within `setup_test_logging`.
+        * Use the `STREAM_LOG` environment variable to determine whether the logs
+            should be logged to the console. If it is not set, or set to an invalid
+            level, it defaults to `TRUE`. The valid options are `TRUE` and `FALSE`,
+            case insensitive.
+    """
+    from qiskit.test.utils import setup_test_logging
```
jyu00: Can we not use setup_**test**_logging for production logging?

qiskit/providers/ibmq/utils/utils.py
```diff
@@ -56,3 +58,31 @@ def validate_job_tags(job_tags: Optional[List[str]], exception: Type[Exception])
...
+
+    log_level = os.getenv('LOG_LEVEL', '')
+    stream_log = os.getenv('STREAM_LOG', '')
```
jyu00: I know they're the env variables used by qiskit test, but they're terrible names. Maybe something more unique like `QISKIT_IBMQ_PROVIDER_LOG_LEVEL`. Also can we have separate options for logging to the screen vs file? These variables should also be documented in a `ibmq/__init__`.

qiskit/providers/ibmq/api/rest/job.py
```diff
@@ -63,6 +66,7 @@ def get(self) -> Dict[str, Any]:
         url = self.get_url('self')
 
         response = self.session.get(url).json()
+        logger.debug('Retrieved job %s. Endpoint: %s', self.job_id, url)
```
jyu00: These log messages are nice to have, but if you don't want to add logging to every method, you could just log in `session` (which should be done regardless) since a lot can be inferred from the url + request data.

qiskit/providers/ibmq/utils/utils.py
jyu00: I strongly dislike `LOG_LEVEL`, and I don't think having separate control over test vs production logging is a bad thing.

qiskit/providers/ibmq/api/rest/job.py
jyu00: I think `session.request()` is a nice, centralized place for logging. It also allows you to log extra information easily.  If you think there would be too much noise, you can always choose not to log for some endpoints (such as `auth`).

qiskit/providers/ibmq/utils/utils.py
```diff
@@ -56,3 +59,47 @@ def validate_job_tags(job_tags: Optional[List[str]], exception: Type[Exception])
...
+    stream_handler = logging.StreamHandler()
+    stream_handler.setFormatter(formatter)
+    logger.addHandler(stream_handler)
+
+    # Set the logging file, if specified.
+    if log_file:
+        # Setup the file handler.
+        file_handler = logging.FileHandler(log_file)
+        file_handler.setFormatter(formatter)
+        logger.addHandler(file_handler)
```
jyu00: I'm thinking about not logging to the screen if a log file is specified. This is useful for people who only want to look at the log if there's an error and don't want their screen cluttered.

qiskit/providers/ibmq/api/clients/websocket.py
```diff
@@ -259,8 +259,6 @@ def get_job_status(
                                 timeout = original_timeout - (time.time() - start_time)
                             else:
                                 response_raw = yield from websocket.recv()
-                        logger.debug('Received message from websocket: %s',
-                                     response_raw)
```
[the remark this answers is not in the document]
luisg5: I didn't know the API only sent data when there is a change. The reason I got rid of it was because some of the information contained in the message might be "sensitive", such as the `hub`, `group` and `project`. The idea was to prevent these from showing up in the travis runs. Is the information sensitive enough to censor them before logging the response?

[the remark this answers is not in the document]
luisg5: Similarly, some endpoints would show the device name as well. An example of one is:
```
qiskit.providers.ibmq.session.request:DEBUG:2020-03-11 10:07:07,690: Endpoint: /devices/ibmq_16_melbourne/queue/status. Method: GET. Request Data: {'allow_redirects': True, 'timeout': (5.0, None)}.
```

luisg5:
**TODO**: Fix `test_log_file` since tests are run in parallel.

qiskit/providers/ibmq/api/clients/websocket.py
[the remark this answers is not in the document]
jyu00: I don't think they're super sensitive, but it doesn't hurt to filter them out.

test/ibmq/test_ibmq_logger.py
```diff
@@ -0,0 +1,177 @@
...
+                self.assertEqual(self.logger.level, level_to_set,
+                                 'The logger level was set to {}, but it should '
+                                 'be {}.'.format(self.logger.level, logging.INFO))
+
+    def test_valid_log_levels(self):
+        """Test setting up a logger with all valid levels, case insensitive."""
+        all_valid_log_levels = {
+            'debug': logging.DEBUG, 'iNFo': logging.INFO,
+            'WARNING': logging.WARNING, 'error': logging.ERROR,
+            'CRITICAL': logging.CRITICAL}
```
jyu00: you can probably combine this test with the one above since you're using mixed casing here.

test/ibmq/test_ibmq_logger.py
```diff
@@ -0,0 +1,177 @@
...
+            self.logger.debug('This is a debug message that should not be logged in the file.')
+            self.logger.warning('This is a warning message that should be logged in the file.')
+            self.logger.error('This is a error message that should be logged in the file.')
+
+            # Retrieve the file handler.
+            file_handler = self.logger.handlers[0]
+            file_path = file_handler.baseFilename
+
+            # Assert the file exists.
+            self.assertTrue(os.path.exists(file_path),
```
jyu00: the `file_path` used for verification should be whatever you passed into setup_logger (i.e. self.log_file) and not from the handler.

test/ibmq/test_ibmq_logger.py
```diff
@@ -0,0 +1,177 @@
...
+                        self.assertIn(substring, content_as_str,
+                                      'The substring "{}" was not found in the file {}({}).'
+                                      .format(substring, self.log_file, file_path))
+                    else:
+                        self.assertNotIn(substring, content_as_str,
+                                         'The substring "{}" was found in the file {}({}).'
+                                         .format('debug message', self.log_file, file_path))
+
+            # Delete the file.
+            os.remove(file_path)
```
jyu00: you can use a temporary file instead so you don't have to worry about deleting/conflicting

test/ibmq/test_ibmq_logger.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Note:
+            The log level should default to `NOTSET` when no level is specified.
+        """
+        default_level_not_set = logging.NOTSET
+
+        with patch.dict('os.environ'):
+            if self.log_level_env in os.environ:
+                del os.environ[self.log_level_env]
+            setup_logger(self.logger)
```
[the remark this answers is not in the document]
luisg5: That was my first thought, but I thought it would be too repetitive. I guess both solutions have their pros and cons. Although it's not an intensive task, it does seem worthwhile not having to do a `tearDown` after each test case if possible. Thanks!

test/ibmq/test_ibmq_logger.py
```diff
@@ -0,0 +1,177 @@
...
+
+from qiskit.providers.ibmq.utils.utils import setup_logger
+
+from ..ibmqtestcase import IBMQTestCase
+
+
+class TestLogger(IBMQTestCase):
+    """Tests related to logger setup via ``setup_logger()``."""
+    log_level_env = 'QISKIT_IBMQ_PROVIDER_LOG_LEVEL'
+    log_file_env = 'QISKIT_IBMQ_PROVIDER_LOG_FILE'
```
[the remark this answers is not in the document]
luisg5: Always a bit skeptical to use global variables... But yes, this does seem like an appropriate place, thank you!

luisg5:
**TODO**: I'm gonna take another look at the PR and also run it on travis, using the environment variables and see if everything is working as expected. But, it should be ready for review.

qiskit/providers/ibmq/api/clients/websocket.py
```diff
@@ -261,11 +290,11 @@ def get_job_status(
...
 
                         response = WebsocketResponseMethod.from_bytes(response_raw)
                         last_status = response.data
+                        logger.debug('Received message from websocket: %s',
+                                     response.get_data_filtered())
```
jyu00: `get_data_filtered()` does a deepcopy(), which is not really an issue since the data is tiny. However, you might still want to check the log level before calling it in case data gets bigger in the future.

test/ibmq/test_ibmq_logger.py
```diff
@@ -0,0 +1,152 @@
...
+                    self.assertEqual(logger.level, level_value,
+                                     'The logger level was set to {}, but it should '
+                                     'be {}.'.format(logger.level, level_value))
+
+    # TODO: NamedTemporaryFiles do not support name in Windows
+    @skipIf(os.name == 'nt', 'Test not supported in Windows')
+    def test_log_file(self):
+        """Test setting up a logger by specifying a file and log level."""
+        logger = logging.getLogger(self.id())
+        log_level_warning = ('ERROR', logging.ERROR)
```
jyu00: Is this supposed to be `log_level_error`?

jyu00:
One more comment. You mentioned that you wanted to reuse `LOG_LEVEL` for both test and mainline so the user doesn't have to specify two different variables. You can still accomplish this by manually setting the logging level of  `qiskit.providers.ibmq` to `LOG_LEVEL` in `IBMQTestCase`.

luisg5:
> One more comment. You mentioned that you wanted to reuse LOG_LEVEL for both test and mainline so the user doesn't have to specify two different variables. You can still accomplish this by manually setting the logging level of qiskit.providers.ibmq to LOG_LEVEL in IBMQTestCase.

Are you suggesting changing the level of the `qiskit.providers.ibmq` logger within the tests, to the level specified by `LOG_LEVEL`, when `LOG_LEVEL` is set as an environment variable?

jyu00:
Some comments based on the Travis output:

- [ ] Drop `%(name)s` from the log format to shorten the log. `%(module)s.%(funcName)s` should be enough to tell us where it's from.
- [ ] Why is log level displayed as `[secure]`?
- [ ] Drop `/devices/.../queue/status` and `/devices/v/1`
- [ ] Log that we're uploading/downloading to/from object storage but don't log the actual url
- [ ] Log request data only for certain endpoints

jyu00:
> Are you suggesting changing the level of the `qiskit.providers.ibmq` logger within the tests, to the level specified by `LOG_LEVEL`, when `LOG_LEVEL` is set as an environment variable?

Yes, if the `qiskit.providers.ibmq` logger level is UNSET.

luisg5:
> Why is log level displayed as [secure]?

This was the case because I had the setting in my Travis CI to hide the value set for `QISKIT_IBMQ_PROVIDER_LOG_LEVEL`. I unchecked the box, so it's not like that in future runs.

Also, I added this into `IBMQTestCase`:
> One more comment. You mentioned that you wanted to reuse LOG_LEVEL for both test and mainline so the user doesn't have to specify two different variables. You can still accomplish this by manually setting the logging level of qiskit.providers.ibmq to LOG_LEVEL in IBMQTestCase.

qiskit/providers/ibmq/api/session.py
```diff
@@ -259,3 +268,50 @@ def _modify_chained_exception_messages(self, exc: BaseException) -> None:
...
+                        # Replace backend name in job submission request.
+                        request_data['json']['backend']['name'] = '...'
+                    logger.debug('Endpoint: %s. Method: %s. Request Data: %s.',
+                                 filtered_url, method, request_data)
+                else:
+                    if bare:
+                        if method == 'PUT':
+                            logger.debug('Uploading Qobj to object storage.')
+                        elif method == 'GET':
+                            logger.debug('Downloading Qobj from object storage.')
```
jyu00: Can you move these loggings to the actual job/result upload/download methods? I think it's better not to rely on PUT/GET with bare url to be object storage upload/download.

qiskit/providers/ibmq/api/session.py
```diff
@@ -259,3 +268,50 @@ def _modify_chained_exception_messages(self, exc: BaseException) -> None:
...
+            request_data: Dict[str, Any]
+    ) -> None:
+        """Log the request data, filtering out specific information.
+
+        Note:
+            The string ``...`` is used to denote information that has been filtered out
+            from the request, within the url and request data.
+
+            The following endpoint URLs are not logged in order to reduce noise: ``/users``
+            and ``/version``. Likewise, the request data is only logged for the following
```
jyu00: The docstring is now outdated with additional endpoints not being logged. It might be easier to either just say "not all endpoints are logged" or define a "constant" for excluded endpoints and just say those in the constant are not logged.

test/ibmq/test_ibmq_logger.py
```diff
@@ -0,0 +1,152 @@
...
+from tempfile import NamedTemporaryFile
+from unittest import skipIf, mock
+
+from qiskit.providers.ibmq.utils.utils import setup_logger
+
+from ..ibmqtestcase import IBMQTestCase
+
+# Constants used by `TestLogger`.
+QISKIT_IBMQ_PROVIDER_LOG_LEVEL = 'QISKIT_IBMQ_PROVIDER_LOG_LEVEL'
+QISKIT_IBMQ_PROVIDER_LOG_FILE = 'QISKIT_IBMQ_PROVIDER_LOG_FILE'
```
jyu00: They're probably better to be defined in `qiskit.providers.ibmq.__init__` since they're mainline constants.

test/ibmqtestcase.py
```diff
@@ -14,12 +14,31 @@
 
 """Custom TestCase for IBMQProvider."""
 
+import os
+import logging
+
 from qiskit.test import QiskitTestCase
 
+# Name of the ibmq provider modules logger.
+IBMQ_PROVIDER_LOGGER_NAME = 'qiskit.providers.ibmq'
```
jyu00: Same as above, this should be defined in mainline.

qiskit/providers/ibmq/__init__.py
```diff
@@ -93,8 +121,13 @@ def least_busy(backends: List[BaseBackend]) -> BaseBackend:
...
     try:
-        return min([b for b in backends if b.status().operational],
-                   key=lambda b: b.status().pending_jobs)
+        backends_to_filter = {}
+        for backend in backends:
+            backend_status = backend.status()
+            if backend_status.operational:
+                backends_to_filter[backend] = backend_status
+
+        return min(backends_to_filter.keys(), key=lambda b: backends_to_filter[b].pending_jobs)
```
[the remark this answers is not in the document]
luisg5: I changed it because `status()` was being called twice for every backend and this was cluttering the log. But, we're not going to log `/devices/<device_name>/status` so I'll change it back.

qiskit/providers/ibmq/api/session.py
luisg5: I separated out filtering out url logs by introducing another private method [_is_worth_logging(...)](https://github.com/Qiskit/qiskit-ibmq-provider/pull/579/files#diff-fb7e1b8f723ef3e2f360841969f49284R315), in order to make filtering more readable and maintainable (in case we add other endpoints in the future). Would you rather have a "constant" be defined still or would the function be an okay addition instead?

test/ibmq/test_ibmq_logger.py
luisg5: Would it be a good idea to include these constants under the html docs for `qiskit.providers.ibmq` so users are more aware of them? Or, is their intention just for having a more easier/better access with in the provider modules?

qiskit/providers/ibmq/api/session.py
```diff
@@ -259,3 +268,74 @@ def _modify_chained_exception_messages(self, exc: BaseException) -> None:
...
+
+        Returns:
+            Whether the endpoint URL should be logged.
+        """
+        if endpoint_url in ('/devices/.../queue/status', '/devices/v/1'):
+            return False
+        if endpoint_url.startswith(('/users', '/version')):
+            return False
+        if 'objectstorage' in endpoint_url:
+            return False
```
luisg5: It was mentioned that the object storage endpoint url should not be logged. It's a bit more difficult to filter since the `url` is not an endpoint but instead an absolute url. I did notice that `objectstorage` is contained in the base url and filtered for it, since I couldn't think of another way to filter for it...

qiskit/providers/ibmq/api/session.py
jyu00: I personally like a "constant" since it's obvious and easy to update, but it's not a big deal to just use the method.

test/ibmq/test_ibmq_logger.py
jyu00: Attributes don't show up very well in the docs now, so I wouldn't bother.

qiskit/providers/ibmq/api/rest/job.py
```diff
@@ -170,6 +173,7 @@ def put_object_storage(self, url: str, qobj_dict: Dict[str, Any]) -> str:
         """
         data = json.dumps(qobj_dict, cls=json_encoder.IQXJsonEconder)
         response = self.session.put(url, data=data, bare=True)
+        logger.debug('Uploading Qobj to object storage.')
```
jyu00: Can you move this to before line 175? We had cases wherein job data was never uploaded for unknown reason, and this would help to see if upload was initiated. Same for result download.

PR #582, opened 2020-03-05
Use old job status if new one is not available

### LINKED ISSUE #564: Support callback only on state change in wait_for_final_state()
`wait_for_final_state()` currently supports a callback function. Often this callback function only cares when the job state changes (job status, queue position, etc). It'd be if `wait_for_final_state()` provides the option to invoke callback only on state changes, so the caller doesn't have to keep track of that.

### LINKED ISSUE #581: job.wait_for_final_state() doesn't invoke callback within wait time
### Information

- **API URL or version**:

- **Python version**:
- **Operating system**:

`job.wait_for_final_state()` doesn't invoke callback function within wait time if wait time is small. This is due to a recent API change that only provides job status through websocket when there's a status change.

jyu00:
### Summary
Fixes #581 . `job.wait_for_final_state()` will now invoke the callback function with old information if no new information is available.

Also closes #564 since it's the same block of code.


qiskit/providers/ibmq/utils/utils.py
```diff
@@ -56,3 +58,65 @@ def validate_job_tags(job_tags: Optional[List[str]], exception: Type[Exception])
     if job_tags and (not isinstance(job_tags, list) or
                      not all(isinstance(tag, str) for tag in job_tags)):
         raise exception("job_tags needs to be a list or strings.")
+
+
+class RefreshQueue(Queue):
+    """A queue that replaces the oldest item with the new when full.
```
luisg5: Swapping `new` for `newest item` might make more sense in this sentence

qiskit/providers/ibmq/utils/utils.py
```diff
@@ -56,3 +58,65 @@ def validate_job_tags(job_tags: Optional[List[str]], exception: Type[Exception])
...
+    """
+
+    def __init__(self, maxsize: int):
+        """RefreshQueue constructor.
+
+        Args:
+            maxsize: Maximum size of the queue.
+        """
+        self.condition = Condition()
+        # self.queue = Queue(maxsize)
```
luisg5: Is there a reason why this was left commented out?

qiskit/providers/ibmq/utils/utils.py
jyu00: I didn't want to use `newest item` because that implies the item is already on the queue and not being added to the queue. Reworded it to hopefully make it clearer.

Issue #549, opened 2020-02-18
Remove shots from job submit


jyu00:
### What is the expected behavior?
`shots` currently is present both inside and outside the qobj. The one outside is not used but required by the API. The API recently update the code to no longer require it, and thus it should be removed from the provider code.

jyu00:
Fixed via #583.

PR #583, opened 2020-03-11
Remove shots from job submit

luisg5:
### Summary

Addresses #549 

### Details and comments

Removes shots from `create_remote_jobs` and other accompanying places.

Rather than writing another test, an extra assert was added to `test_job_submit_object_storage`. It asserts that the `shots` attribute is not returned by the server in the job submit info.

### DIFF
--- a/qiskit/providers/ibmq/job/schema.py
+++ b/qiskit/providers/ibmq/job/schema.py
@@ -36,7 +36,6 @@ FIELDS_MAP = {
     'error': '_error',
     'name': '_name',
     'timePerStep': '_time_per_step',
-    'shots': '_api_shots',
     'tags': '_tags'
 }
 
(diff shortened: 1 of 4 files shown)


qiskit/providers/ibmq/job/schema.py
```diff
@@ -36,7 +36,6 @@
     'error': '_error',
     'name': '_name',
     'timePerStep': '_time_per_step',
-    'shots': '_api_shots',
```
jyu00: Can you verify that when retrieving older jobs, they won't have the `shots` attribute?

luisg5: According to the API, `shots` would still be present in the root level attributes when retrieving older jobs. But, it would not be present in the root level for newer jobs that only submitted `shots` within the obj. As a result, I re-added the `'shots': '_api_shots'` field back into `FIELD_MAPS`.

Issue #580, opened 2020-03-05
Fix flaky tests


jyu00:
### What is the expected behavior?
There are a bunch of tests that fail occasionally due to timing. It'd be nice to make them more stable.

PR #587, opened 2020-03-13
Fix flaky tests

jyu00:
### Summary
Addresses #580 . This probably doesn't fix all tests, but it's a start.

### DIFF
--- a/test/ibmq/websocket/test_websocket_integration.py
+++ b/test/ibmq/websocket/test_websocket_integration.py
@@ -29,6 +29,7 @@ from qiskit.providers.jobstatus import JobStatus
 
 from ...ibmqtestcase import IBMQTestCase
 from ...decorators import requires_provider, requires_device
+from ...utils import most_busy_backend, bell_in_qobj
 
 
 class TestWebsocketIntegration(IBMQTestCase):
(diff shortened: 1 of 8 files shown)


test/ibmq/websocket/test_websocket_integration.py
```diff
@@ -142,16 +147,21 @@ def _job_status_side_effect(*args, **kwargs):
...
+        qobj = bell_in_qobj(backend, shots=backend.configuration().max_shots)
+        job = backend.run(qobj, validate_qobj=True)
 
-        with self.assertRaises(JobTimeoutError):
-            job.result(timeout=0.1)
+        try:
+            with self.assertRaises(JobTimeoutError):
+                job.result(timeout=0.1)
+        finally:
+            job.cancel()
```
luisg5: Might also be worth wrapping this in a try except, as mentioned above, so the test would not fail if there is a problem cancelling the job

test/utils.py
```diff
@@ -53,3 +57,17 @@ def get_large_circuit(backend: IBMQBackend) -> QuantumCircuit:
     circuit.measure(list(range(n_qubits)), list(range(n_qubits)))
 
     return circuit
+
+
+def bell_in_qobj(backend: IBMQBackend, shots: int = 1024) -> Qobj:
```
luisg5: Nice addition!

luisg5: Nice addition!

test/utils.py
```diff
@@ -71,3 +74,30 @@ def bell_in_qobj(backend: IBMQBackend, shots: int = 1024) -> Qobj:
...
+        verify: Verify job status.
+
+    Returns:
+        Whether job has been cancelled.
+    """
+    cancelled = False
+    for _ in range(2):
+        # Try twice in case job is not in a cancellable state
+        try:
+            if job.cancel():
```
luisg5: You forgot to set `cancelled = True` within `if job.cancel()` 😅

jyu00: Oops

Issue #526, opened 2020-01-29
Evaluate supporting editing some job attributes


diego-plan9:
### What is the expected behavior?

In a recent update, API has introduced an endpoint for editing a small subset of attributes of a job. It would be nice to check if it can be used by the provider.

jyu00:
If we do allow the user to modify job tags, we should ensure they cannot modify those used by Job Manager.

PR #590, opened 2020-03-16
Edititable job attributes

luisg5:
### Summary

Addresses #526


qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -279,6 +285,173 @@ def cancel(self) -> bool:
             raise IBMQJobApiError('Unexpected error when cancelling job {}: {}'
                                   .format(self.job_id(), str(error))) from error
 
+    def change_name(self, name: str) -> bool:
```
luisg5: Another alternative was `update_name`, in case it sounds better than `change_name` one?

The same would go for `modify_tags` -> `update_tags`.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -279,6 +285,173 @@ def cancel(self) -> bool:
...
+            These tags are necessary to retrieve a job within a
+            :class:`ManagedJobSet<qiskit.providers.ibmq.managed.ManagedJobSet>`.
+            If an attempt is made to remove one of these tags, the removal for that
+            tag will be ignored.
+
+        This method plays it safe when attempting to remove tags that are not found in the
+        job's existing tags by ignoring them. As a result, this method will return ``True``
+        when no tags are removed, or not all tags are removed, if they are not actually
+        associated with this job. The following examples give a better indication of these
+        cases.
```
luisg5: Decided to go with this logic since it made the most sense. Essentially, the idea is that attempting to remove a tag that is not present is successful, since the desired tag is not part of the job's tags anyway. Also, to make the user aware of the case, a warning is issued if a tag to remove is not found in the job's tags.

Although, it could be changed so that if none of the tags to remove are found in the job's tags, the method would return `False`. This is essentially the case for the first example below (But, instead of returning `True`, returning `False`)

However, this would open up the question:  If some tags are removed and others are not (the second example), what should be returned `True` or `False`? This is the reason why I went with the logic mentioned above.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -279,6 +285,173 @@ def cancel(self) -> bool:
...
+        updated_tags = response.get('tags', None)
+        updated_tags_set = set(updated_tags) if updated_tags else None
+        update_successful = (tags_to_update == updated_tags_set)
+
+        if update_successful:
+            self._tags = updated_tags  # Cache updated tags if successful.
+
+        return update_successful
+
+    def remove_tags(self, tags: List[str]) -> bool:
```
jyu00: I think you can consolidate this method with `modify_tags`, something like 
```
def update_tags(self, replacement_tags=None, additional_tags=None, remove_tags=None)
```

It's one fewer method for the users to remember, also easier to update a specific tag (e.g. changing "priority-high" to "priority-low").

test/ibmq/test_ibmq_job_attributes.py
```diff
@@ -106,6 +106,50 @@ def test_job_name(self, provider):
...
+                retrieved_job_ids = {job.job_id() for job in retrieved_jobs}
+                self.assertIn(job_id, retrieved_job_ids)
+
+                if new_name:
+                    # Check retrieving the job by its new name using regular expressions.
+                    job_name_regex = '^{}$'.format(new_name)
+                    retrieved_jobs = provider.backends.jobs(backend_name=backend.name(),
+                                                            job_name=job_name_regex)
+                    self.assertEqual(len(retrieved_jobs), 1)
+                    self.assertEqual(job_id, retrieved_jobs[0].job_id())
```
jyu00: I don't think testing retrieving the new name using partial/regex is necessary as they're covered by other tests. You should, however, do a `job.refresh()` (which will re-fetch job attributes) and verify the job name is the new one.

test/ibmq/test_ibmq_jobmanager.py
```diff
@@ -303,6 +303,160 @@ def test_job_tags(self, provider):
...
+
+                # Ensure the job set long id is in the new tags.
+                self.assertIn(job_set._id_long, job.tags(),
+                              'The job {} with tags {} does not contain the job set '
+                              'long id "{}".'.format(job_id, job.tags(), job_set._id_long))
+
+                # Ensure the job is retrieved with the new tags it is associated with.
+                retrieved_jobs = provider.backends.jobs(job_tags=tags_after_appending)
+                retrieved_job_ids = {job.job_id() for job in retrieved_jobs}
+                self.assertIn(job_id, retrieved_job_ids)
```
jyu00: Same as above, retrieving is unnecessary

test/ibmq/test_ibmq_jobmanager.py
```diff
@@ -303,6 +303,160 @@ def test_job_tags(self, provider):
...
+        job_id = job.job_id()
+
+        timestamp = str(time.time()).replace('.', '')
+        new_tags_to_overwrite = [
+            [],  # empty tags.
+            ['{}_new_tag_{}'.format(timestamp, i) for i in range(2)]  # non-empty unique tags.
+        ]
+        # Overwrite the tags for a job in the job set.
+        for new_tags in new_tags_to_overwrite:
+            new_tags_with_id_long = new_tags + [job_set._id_long]
```
jyu00: It'd be more interesting to test replacement tags _without_ the job set ID

test/ibmq/test_ibmq_jobmanager.py
```diff
@@ -303,6 +303,160 @@ def test_job_tags(self, provider):
...
+                                'The tags are {}, but they should be {}.'
+                                .format(job_id, job.tags(), new_tags_with_id_long))
+
+                # Ensure the job's tags match the new tags and contain the job set long id.
+                self.assertEqual(set(job.tags()), set(new_tags_with_id_long),
+                                 'The tags for job {} are {}, but they should be {}'
+                                 .format(job_id, job.tags(), new_tags_with_id_long))
+
+                # Ensure the job set long id is in the new tags.
+                self.assertIn(job_set._id_long, job.tags(),
```
jyu00: This seems to be a duplicate check as above. If `job.tags()` equals to `new_tags_with_id_long` (which has `_id_long`), then `_id_long` must be in `job.tags()`.

qiskit/providers/ibmq/api/rest/job.py
```diff
@@ -72,6 +72,30 @@ def get(self) -> Dict[str, Any]:
...
+            JSON response containing the name of the updated attribute and its
+            corresponding value.
+        """
+        url = self.get_url('self')
+
+        data = {}
+        if 'name' in job_attribute_info:
+            data['name'] = job_attribute_info['name']
+        elif 'tags' in job_attribute_info:
+            data['tags'] = job_attribute_info['tags']
```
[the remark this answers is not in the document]
luisg5: Silly mistake! 😅

test/ibmq/test_ibmq_job_attributes.py
luisg5: I could see how they could be redundant. The idea was more so to test that it was possible to retrieve the job after updating the name (ensuring the new name was changed successfully in the server). I'll go ahead and replace that with your suggested test, however, since its essentially the same idea.

luisg5:
Some of the tests have been shaky, I'm gonna look into it before we would merge.

qiskit/providers/ibmq/managed/managedjobset.py
```diff
@@ -418,6 +419,42 @@ def name(self) -> str:
...
+            if job:
+                try:
+                    _ = job.update_name("{}_{}_".format(name, i))
+                except IBMQJobUpdateError as ex:
+                    raise IBMQJobManagerUpdateError(
+                        "An error occurred when updating the name for this job set: {}"
+                        "As a result, some of the jobs in this set may not have been "
+                        "updated. Please try updating the job set's name again "
+                        "to update the name for all jobs in this set."
+                        .format(str(ex))) from ex
```
luisg5, line 451: Decided to "fail fast" if there is an error with updating the name of a job in the job set, and letting the user know of it. 

Or is it more reasonable to continue updating the name of the rest of the jobs, while logging the exception for the jobs that failed to update. Then, raise an exception in the end, specifying which jobs were not updated?

Note: The case is similar for `update_tags`.

qiskit/providers/ibmq/managed/managedjobset.py
```diff
@@ -418,6 +419,42 @@ def name(self) -> str:
...
+        """
+        if not isinstance(name, str):
+            raise IBMQJobManagerInvalidStateError(
+                '"{}" of type "{}" is not a valid job name. '
+                'The job name needs to be a string.'.format(name, type(str)))
+
+        for i, job in enumerate(self.jobs()):
+            if job:
+                try:
+                    _ = job.update_name("{}_{}_".format(name, i))
```
jyu00, line 444: Since the job name format (`{}_{}_`) is now used at multiple places, it's better to either declare it as a constant or put it in a method.

qiskit/providers/ibmq/managed/managedjobset.py
```diff
@@ -418,6 +419,42 @@ def name(self) -> str:
...
+                except IBMQJobUpdateError as ex:
+                    raise IBMQJobManagerUpdateError(
+                        "An error occurred when updating the name for this job set: {}"
+                        "As a result, some of the jobs in this set may not have been "
+                        "updated. Please try updating the job set's name again "
```
jyu00, line 449: Don't tell the user to "try again" unless you know it's a temporary networking error (which everyone knows happens from time to time) because it makes the code looks buggy and the user would wonder why you don't just retry in the code.

qiskit/providers/ibmq/managed/managedjobset.py
jyu00: Job manager usually issues a warning instead of raising an error if _some_ of the job operations failed. It'd also be nice to denote a job name update failed in the job manager report output (e.g. with an asterisk next to the job name or something)

test/ibmq/test_ibmq_job_attributes.py
```diff
@@ -107,6 +107,37 @@ def test_job_name(self, provider):
...
+    def test_job_name_update(self, provider):
+        """Test changing the name associated with a job."""
+        backend = provider.get_backend('ibmq_qasm_simulator')
+        qobj = assemble(transpile(self._qc, backend=backend), backend=backend)
+
+        # Use a unique job name
+        initial_job_name = str(time.time()).replace('.', '')
+        job = backend.run(qobj, job_name=initial_job_name, validate_qobj=True)
+        job_id = job.job_id()
+        # TODO No need to wait for job to run once api is fixed
```
jyu00, line 120: I keep seeing this in tests, but is this really true? I thought it only applies to `backend.jobs()` with status as the filter?

test/ibmq/test_ibmq_job_attributes.py
```diff
@@ -107,6 +107,37 @@ def test_job_name(self, provider):
...
+        while job.status() not in JOB_FINAL_STATES + (JobStatus.RUNNING,):
+            time.sleep(0.5)
+        rjob = provider.backends.retrieve_job(job_id)
+        self.assertEqual(rjob.name(), initial_job_name)
```
jyu00, line 124: Setting a job name with `backend.run()` is already covered in another test case; no need to verify this here unless you want to combine the 2 test cases.

qiskit/providers/ibmq/managed/managedjobset.py
```diff
@@ -418,6 +419,42 @@ def name(self) -> str:
         """
         return self._name
 
+    def update_name(self, name: str) -> str:
```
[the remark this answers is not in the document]
jyu00: Also, `ManagedJobSet.retrieve_jobs()` verifies each job name is proper (i.e. all have the same prefix). Since individual job can now have its name updated at random, it may be easier to just not do that verification.

test/ibmq/test_ibmq_job_attributes.py
```diff
@@ -349,6 +380,100 @@ def test_job_tags_and(self, provider):
...
+                _ = job.update_tags(removal_tags=tags_to_remove)  # Update the job tags.
+                # Cached results may be returned if updating the same job too quickly.
+                # Wait before updating again.
+                time.sleep(2)
+                job.refresh()
+                self.assertEqual(set(job.tags()), tags_after_removal_set,
+                                 'Updating the tags for job {} was unsuccessful.'
+                                 'The tags are {}, but they should be {}.'
+                                 .format(job_id, job.tags(), list(tags_after_removal_set)))
+
```
jyu00, line 476: You should also have test cases that covers when some combinations of `replace/add/remove` are specified.

test/ibmq/test_ibmq_job_attributes.py
luisg5: Yeah, that seems to be the case. So they are not needed in the tests, I'll remove them.

qiskit/providers/ibmq/job/exceptions.py
```diff
@@ -42,3 +42,7 @@ class IBMQJobInvalidStateError(IBMQJobError):
 class IBMQJobTimeoutError(JobTimeoutError, IBMQJobError):
     """Errors raised when a job operation times out."""
     pass
+
+
+class IBMQJobUpdateError(IBMQJobError):
```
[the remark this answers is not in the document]
luisg5: I believed adding the exception was a good idea, in order to specify the type of error that occurred. But, it is likely better to use `IBMQJobApiError` since it's relevant and doesn't require a new exception class 👍

qiskit/providers/ibmq/managed/utils.py
```diff
@@ -110,7 +127,19 @@ def format_job_details(
...
         report.append("    job ID: {}".format(job.job_id()))
-        report.append("    name: {}".format(job.name()))
+
+        job_name_info = "    name: {}".format(job.name())
+        # If a job name does not match the managed job set name, mark it with an asterisk.
+        matched = JOB_SET_NAME_RE.match(job.name()) if job.name() else None
+        if not matched:
+            raise IBMQJobManagerInvalidStateError(
+                'Job {} is tagged for the job set {} but does not appear '
+                'to belong to the set.'.format(job.job_id(), managed_jobs_id))
```
jyu00, line 137: Unmatched job names should be handled elsewhere as this function is only used to format a report.

test/ibmq/test_ibmq_job_attributes.py
```diff
@@ -349,6 +375,165 @@ def test_job_tags_and(self, provider):
...
+                            additional_tags=additional_tags,
+                            removal_tags=removal_tags)
+        # Cached results may be returned if updating the same job too quickly.
+        # Wait before updating again.
+        time.sleep(2)
+        job.refresh()
+        self.assertEqual(set(job.tags()), set(tags_after_update),
+                         'Updating the tags for job {} was unsuccessful.'
+                         'The tags are {}, but they should be {}.'
+                         .format(job_id, job.tags(), tags_after_update))
```
jyu00, line 535: Since updating & verifying job tags is repeated for 7 test cases, you can probably put it in a function to make the code more readable.

qiskit/providers/ibmq/managed/managedjobset.py
```diff
@@ -173,12 +172,18 @@ def retrieve_jobs(self, provider: AccountProvider, refresh: bool = False) -> Non
...
+            if not matched or job.backend().name() != self._backend.name():
                 raise IBMQJobManagerInvalidStateError(
                     'Job {} is tagged for the job set {} but does not appear '
                     'to belong to the set.'.format(job.job_id(), self.job_set_id()))
+            if matched.group(1) != self._name:
+                logger.warning('The job %s, belonging to job set %s, does not appear to have '
+                               'a proper job name. The job name should include the job set name '
+                               '"%s", but instead includes the job set name "%s". Use the '
+                               'update_name() method to update the name of the jobs in the job set',
+                               job.job_id(), self.job_set_id(), self.name(), matched.group(1))
```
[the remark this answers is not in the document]
luisg5: Yeah, it's a bit difficult to know whether a job belongs to a job set since certain attributes could be updated at any time. Would it be more reasonable to check whether the job contains the job set tag, in order to determine whether the job belongs to the job set?

qiskit/providers/ibmq/managed/managedjobset.py
```diff
@@ -173,12 +172,18 @@ def retrieve_jobs(self, provider: AccountProvider, refresh: bool = False) -> Non
         jobs_dict = {}
         for job in jobs:
             # Verify the job is proper.
-            matched = pattern.match(job.name()) if job.name() else None
-            if not matched or matched.group(1) != self._name or \
-                    job.backend().name() != self._backend.name():
+            matched = JOB_SET_NAME_RE.match(job.name()) if job.name() else None
+            if not matched or job.backend().name() != self._backend.name():
```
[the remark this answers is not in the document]
luisg5: The idea was that the job name should at least follow the job set name format. But, I could see that it's not really necessary, since the job name could be changed at any time.

qiskit/providers/ibmq/managed/managedjobset.py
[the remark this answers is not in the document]
jyu00: The job set tag is already used to determine whether a job belongs to a set. Checking for job name and backend name is just extra precaution. One possible strategy is to collect all the job names and see if they all have the same prefix. If so, deem that as the name of the set. If not, set the job set name to None and issue a warning.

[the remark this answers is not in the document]
luisg5: As mentioned offline, I kept the original behavior of enforcing that all jobs have a proper name and raising an error if a job does not.

PR #596, opened 2020-03-20
Change from IBMQ -> IQX

nonhermitian:
### Summary
Change labels from IBMQ -> IQX 

Requires: https://github.com/Qiskit/qiskit-terra/pull/3997

### Details and comments
Changes names of things to IQX for better alignment.

### DIFF
--- a/qiskit/providers/ibmq/__init__.py
+++ b/qiskit/providers/ibmq/__init__.py
@@ -57,40 +57,40 @@ Classes
 
     AccountProvider
     BackendJobLimit
-    IBMQBackend
-    IBMQBackendService
-    IBMQFactory
+    IQXBackend
+    IQXBackendService
+    IQXFactory
 
 Exceptions
 ==========
 .. autosummary::
     :toctree: ../stubs/
 
-    IBMQError
-    IBMQAccountError
-    IBMQAccountCredentialsNotFound
-    IBMQAccountCredentialsInvalidFormat
-    IBMQAccountCredentialsInvalidToken
-    IBMQAccountCredentialsInvalidUrl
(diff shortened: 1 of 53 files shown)


jyu00:
We are going to need some migration plan. Otherwise this is going to break everyone who uses IBMQ. This will also require users to change a lot of code since variables like `IBMQ` and `IBMQError` are used a lot.

nonhermitian:
The Terra PR has backward compatibility

jyu00:
The terra PR only covers `IBMQ` (admittedly that's the most frequently used one). It's likely people also use the exception classes (e.g. `IBMQError`), `IBMQJobManager`, etc. 

If `qiskit-ibmq-provider` is to be moved out of Qiskit meta, perhaps that should happen before this change. So users that update Qiskit will not drag in this IQX change, which can be installed separately when they're ready to modify their applications.

nonhermitian:
If there is usage of things like `IBMQError` it is not in our elements.  There is some additional stuff in Aqua that needs changing, but it is minimal compared to the changes here and in Qiskit/qiskit-terra#3997.  I am not sure the usage of `IBMQJobManager` as the functionality is relatively new and was not publicized widely.  One could do a `0.6` release with the `IQX` part and backport to the `0.5` that uses the older code.

jyu00:
How about we keep `IBMQError`, `IBMQJobManager`, and maybe `IBMQFactory` as empty classes that inherit from their IQX counterparts, with deprecation warnings? I can do this in a separate PR. This would give people that do reference these classes directly some warning. We can then remove them in a later release.

nonhermitian:
That also works.

nonhermitian:
Closing due to change in plans

PR #598, opened 2020-03-23
Fix asyncio threading

### LINKED ISSUE #595: nest_asyncio.apply() fails when not running under main thread
### Information

- **API URL or version**:

- **Python version**:
- **Operating system**:

`nest_asyncio.apply()` fails with `no current event loop in thread` when `websocket.py` is imported on a child thread. 

Run a script that imports qiskit on a child thread:

```
import threading

def import_fun():
    import qiskit

t = threading.Thread(target=import_fun)
t.start()
t.join()
```

One should be able to import `qiskit-ibmq-provider` on a child thread. 

### Suggested solutions
Manually get the event loop before calling `nest_asyncio.apply()`. If one is not available, create a new loop.

czachow:
Fixes #595 

### Summary
- fixed the problem arising from using nest_aynsio with multithreading
- if no asyncio event loop exists a new one is created and passed to nest_asyncio for patching

### Details and comments
- unsure how to implement tests for my change, any suggestions?


qiskit/providers/ibmq/api/clients/websocket.py
```diff
@@ -44,7 +44,14 @@
...
 # tornado) has its own event loop already so we need to patch it.
 # Patch asyncio to allow nested use of `loop.run_until_complete()`.
-nest_asyncio.apply()
+# Before applying the patch, check if an event loop is available otherwise
+# create one and set it active
+try:
+    LOOP = asyncio.get_event_loop()
+except RuntimeError:
+    LOOP = asyncio.new_event_loop()
+asyncio.set_event_loop(LOOP)
```
jyu00: You don't actually need to set the event loop here since it's not used (yet). `nest_asyncio.apply()` will use the loop you pass in and not the active loop.

jyu00:
For testing, you can reload the `qiskit.providers.ibmq.api.clients.websocket` module in a separate thread, similar to the steps described in the issue (you need to reload instead of import since it's already imported). 

Can you also verify that this works in a Jupyter notebook? Since it also uses the event loop.

youens:
Thanks @czachow, this has been a problem for me as well. 👍 🍪

czachow:
> Thanks @czachow, this has been a problem for me as well. 👍 🍪

Always happy if I can help and contribute to qiskit.

czachow:
I implemented a test but there is still one open problem. 

If 'qiskit' is imported inside a thread the new asyncio loop is created outside any method. It is left behind (not stopped/closed but force closed) when the thread finishes and thus a 'ResourceWarning' is issued. In my opinion this might be okay but is not nice.

test/ibmq/websocket/test_websocket.py
```diff
@@ -49,6 +48,22 @@ def test_invalid_url(self):
             asyncio.get_event_loop().run_until_complete(
                 client.get_job_status('job_id'))
 
+    def test_asyncio_threading(self):
+        """Test asyncio when importing webserver in new thread"""
+        import sys
+        import importlib
+        import threading
```
jyu00: Can you move these imports to the top of the module?

jyu00:
> If 'qiskit' is imported inside a thread the new asyncio loop is created outside any method. It is left behind (not stopped/closed but force closed) when the thread finishes and thus a 'ResourceWarning' is issued. In my opinion this might be okay but is not nice.

You can always close the loop in your thread, since it's not used anymore, like
```
from qiskit.providers.ibmq.api.clients.websocket import LOOP
LOOP.close()
```

Issue #545, opened 2020-02-13
Default Provider saved in account


ryan-woo:
### What is the expected behavior?

It would be wonderful to be able to set your default provider in the IBMQ account that gets saved to disk. Then `IBMQ.load_account()` returns your default provider.

nonhermitian:
I concur. This is simply adding a new key to the `[ibmq]` section, however the logic around it will require some work, e.g. what to do if the set default is no longer there.

PR #611, opened 2020-03-27
Allow saving default provider in account

luisg5:
### Summary

Addresses #545 

### Details and comments

Allows saving a default provider with ``save_account()`` and enabling a default provider using ``enable_account()`` by specifying a provider as a string in the format ``<hub_name>/<group_name>/<project_name>``.


qiskit/providers/ibmq/credentials/configrc.py
```diff
@@ -144,7 +172,9 @@ def store_credentials(
                        'Set overwrite=True to overwrite.')
         return
 
-    # Append and write the credentials to file.
+    # Clear `stored_credentials` to disallow saving multiple credentials to disk,
+    # and write the new `credentials` to file.
+    stored_credentials.clear()
```
luisg5: When saving different providers to disk, multiple ``[ibmq*]`` sections could be written, since the ``unique_id`` would be different for each one. 

To prevent that, I `clear` the `stored_credentials` before writing the new credentials, given that having multiple accounts stored on disk is discourage and causes the account methods such as ``stored_accounts``, ``load_account``, and ``delete_account`` to raise an exception. 

Should we keep this as the default behavior or revert to the previous, which would allow users to save different providers to disk (having the account methods still raise an exception of course)?

qiskit/providers/ibmq/credentials/configrc.py
```diff
@@ -89,11 +98,30 @@ def write_qiskit_rc(
...
+        }
+
+        if all(hgp_dict.values()):
+            provider_to_save_as_str = get_provider_as_str(hgp_dict)
+            # Add the provider that will be saved to disk.
+            credentials_dict['default_provider'] = provider_to_save_as_str
+
+        return credentials_dict
 
     def _section_name(credentials_: Credentials) -> str:
```
luisg5: When specifying a default provider to save to disk, the ``base_name`` is given by the `{}_{}_{}_{}` formatter. 

It seems like this is set, in the case that multiple ``[ibmq*]`` sections are present. If multiple sections are allowed, it makes sense to keep this. Otherwise, we might be able to change it so that ``[ibmq]`` is written as the section name, for the account saved in disk?

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -202,6 +216,11 @@ def save_account(
                 'Invalid IBM Quantum Experience token '
                 'found: "{}" of type {}.'.format(token, type(token)))
 
+        if hgp:
+            # Set the hub/group/project info for the `credentials`.
+            hgp_dict = get_provider_as_dict(hgp)
+            kwargs.update(hgp_dict)
```
jyu00: Setting h/g/p in `Credential` has a lot of implication in terms of saving it, due to how things were done before. It's probably easier to just pass in a new `default_provider` parameter.

qiskit/providers/ibmq/credentials/configrc.py
jyu00: You should not overwrite the existing content if `overwrite=True` is not specified. Also I'd prefer there remains a single `[ibmq]` section, which can be done if you pass in `default_provider` instead of saving it in `Credential`.

qiskit/providers/ibmq/api/clients/auth.py
```diff
@@ -136,8 +149,50 @@ def user_hubs(self) -> List[Dict[str, str]]:
...
+                by ``hub``, ``group``, and ``project``, respectively.
+
+        Returns:
+            The index of the hub/group/project, specified by `credentials`, in `user_hubs`,
+            else ``None``.
+        """
+        # The provider, as a dictionary, specified within `credentials`.
+        specified_hgp_entry = {'hub': credentials.hub,
+                               'group': credentials.group,
+                               'project': credentials.project}
```
jyu00: this h/g/p can now be in several format - as a string, dictionary, or embedded in a `Credential`. There's a existing `HubGroupProject` named tuple, maybe it's better to convert that into a class that has methods like `to_stored_format()`.

luisg5:
Gonna do another pass over the documentation, making sure it's nice and clear. Also, a general pass through, to make sure I didn't miss anything and everything is good.

qiskit/providers/ibmq/credentials/hubgroupproject.py
```diff
@@ -16,5 +16,94 @@
 
 from collections import namedtuple
 
-HubGroupProject = namedtuple('HubGroupProject',
-                             ['hub', 'group', 'project'])
+from .exceptions import HubGroupProjectInvalidStateError
+
+HubGroupProjectTuple = namedtuple('HubGroupProjectTuple',
+                                  ['hub', 'group', 'project'])
```
luisg5: Decided to keep an internal nametuple that is to be used only by this class. I thought it would make the code more legible, since it would allow us to spec types such as `HubGroupProjectTuple` instead of `Tuple[str, str, str]`. Also, it would allow us to index the individual tuple fields as well.

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -38,7 +40,25 @@
 
 
 class IBMQFactory:
-    """Factory and account manager for IBM Quantum Experience."""
+    """Factory and account manager for IBM Quantum Experience.
+
+    When saving an account, via :meth:`save_account`, you should specify a default
```
jyu00, line 45: I don't know if one _should_ specify a default. Users that only have access to the open h/g/p certainly don't need to do this.

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -38,7 +40,25 @@
 
 
 class IBMQFactory:
-    """Factory and account manager for IBM Quantum Experience."""
+    """Factory and account manager for IBM Quantum Experience.
+
```
jyu00, line 44: I think this block comment can use more words before jumping into `save_account()`.

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -38,7 +40,25 @@
...
+    When saving an account, via :meth:`save_account`, you should specify a default
+    provider to use for the account. This provider will be used when loading your
```
jyu00, line 46: I think you meant "This provider will be _returned_"

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -51,6 +71,9 @@ def enable_account(
             self,
             token: str,
             url: str = QX_AUTH_URL,
+            hub: str = None,
+            group: str = None,
+            project: str = None,
```
jyu00, line 76: These should be `Optional[str]` :)

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -63,13 +86,17 @@ def enable_account(
         Args:
             token: IBM Quantum Experience token.
             url: URL for the IBM Quantum Experience authentication server.
+            hub: Name of the hub to filter for.
```
jyu00, line 89: I think `Name of the hub to use` sounds better?

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -63,13 +86,17 @@ def enable_account(
...
+            project: Name of the project to filter for.
             **kwargs: Additional settings for the connection:
 
                 * proxies (dict): proxy configuration.
                 * verify (bool): verify the server's TLS certificate.
 
         Returns:
-            The provider for the default open access project.
+            The provider for the default open access project, or a default provider,
+            if specified by `hub`, `group`, `project`.
```
jyu00, line 99: How about
```
If `hub`, `group`, and `project` are specified, the corresponding provider is returned. Otherwise the provider for the open access project is returned.
```

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -102,7 +129,13 @@ def enable_account(
                            'account.')
             return None
 
-        return providers[0]
+        default_provider = providers[0]
+
+        # Only filter if at least one is set.
+        if hub or group or project:
+            default_provider = self.get_provider(hub=hub, group=group, project=project)
```
jyu00, line 136: You should check if the provider for the h/g/p can be found and raise an error if not, so the user doesn't accidentally use the wrong provider. For example, a user makes a typo when specifying h/g/p and doesn't find out until the job finally finishes 7 hours later lol.

qiskit/providers/ibmq/credentials/hubgroupproject.py
```diff
@@ -14,7 +14,97 @@
...
+            group: str = None,
+            project: str = None
+    ) -> None:
+        """HubGroupProject constructor."""
+        self.hub = hub
+        self.group = group
+        self.project = project
+
+    @classmethod
+    def from_str(cls, hgp: str) -> 'HubGroupProject':
```
jyu00, line 37: `from_stored_format()` would be a better name since it's more accurate and mirrors `to_stored_format()`.

test/ibmq/test_ibmq_factory.py
```diff
@@ -114,6 +118,19 @@ def test_pass_unreachable_proxy(self, qe_token, qe_url):
             ibmq.enable_account(qe_token, qe_url, proxies=proxies)
         self.assertIn('ProxyError', str(context_manager.exception))
 
+    @requires_qe_access
+    def test_enable_specified_provider(self, qe_token, qe_url):
+        """Test enabling an account with a specified provider."""
+        ibmq = IBMQFactory()
+        non_default_provider = get_provider(ibmq, qe_token, qe_url, default=False)
+        with no_file('Qconfig.py'), custom_qiskitrc(), no_envs(CREDENTIAL_ENV_VARS):
+            hgp = HubGroupProject.from_credentials(non_default_provider.credentials)
```
jyu00, line 128: Any reason you need to convert it to a `HubGroupProject` instead of just using `non_default_provider.credentials.hub`?

test/ibmq/test_ibmq_factory.py
```diff
@@ -132,19 +149,61 @@ def setUp(self):
...
-    def test_save_account(self):
+    @requires_qe_access
+    def test_save_account(self, qe_token, qe_url):
         """Test saving an account."""
         with custom_qiskitrc():
-            self.factory.save_account(self.token, url=AUTH_URL)
+            default_provider = get_provider(self.factory, qe_token=qe_token, qe_url=qe_url)
+            hgp = HubGroupProject.from_credentials(default_provider.credentials)
+            self.factory.save_account(token=self.token, url=AUTH_URL,
+                                      hub=hgp.hub, group=hgp.group, project=hgp.project)
```
jyu00, line 160: why is it necessary to specify hub/group/project for this test? I'd think a test of `save_account()` without h/g/p is necessary.

test/ibmq/test_ibmq_factory.py
```diff
@@ -132,19 +149,61 @@ def setUp(self):
...
+
+    @requires_qe_access
+    def test_delete_account(self, qe_token, qe_url):
         """Test deleting an account."""
         with custom_qiskitrc():
-            self.factory.save_account(self.token, url=AUTH_URL)
+            default_provider = get_provider(self.factory, qe_token=qe_token, qe_url=qe_url)
+            hgp = HubGroupProject.from_credentials(default_provider.credentials)
+            self.factory.save_account(token=self.token, url=AUTH_URL,
+                                      hub=hgp.hub, group=hgp.group, project=hgp.project)
```
jyu00, line 207: same as before, why does this need a h/g/p?

test/ibmq/test_ibmq_factory.py
luisg5: I think this comment, along with two others, was due to a misunderstanding on my part. It was mentioned that ``save_account`` could "force" the users to specify all three parameters ``hub``, ``group``, and ``project`` since it's not called often. I understood this as making all three parameters mandatory. But, I understand that the user should either specify all three or none of them. I'll make the necessary changes to address this!

test/ibmq/test_ibmq_factory.py
luisg5: I thought it was more convenient, but definitely unnecessary overhead 😅

qiskit/providers/ibmq/credentials/configrc.py
```diff
@@ -33,17 +34,18 @@
...
             location is used (``$HOME/.qiskit/qiskitrc``).
 
     Returns:
-        A dictionary with the contents of the configuration file, in the
-        ``{credential_unique_id: Credentials}`` format. The dictionary is
-        empty if the input file does not exist.
+        A tuple containing the found credentials, if any, and the default
+        provider stored, if specified in the configuration file. The format
+        for the found credentials is ``{credentials_unique_id: Credentials}``,
+        whereas the format for the default provider is ``<hub_name>/<group_name>/<project_name>``.
```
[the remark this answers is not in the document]
luisg5: One reason was to be able to use the string if an exception occurs, to show the user the string they saved in the configuration file. But, I could refactor the exceptions to take this into account, it might make things cleaner in the end as well. Thanks!

docs/tutorials/1_the_ibmq_account.ipynb
```diff
@@ -102,15 +102,15 @@
...
+      "\u001b[0;31mImportError\u001b[0m: cannot import name 'IBMQ' from 'qiskit' (/Users/luisgarcia/PycharmProjects/test-providers/qiskit-ibmq-provider/bin/activate/lib/python3.7/site-packages/qiskit/__init__.py)"
+     ]
```
luisg5, line 113: I'll fix the errors similar to this one, didn't know they should up on GitHub.

qiskit/providers/ibmq/credentials/configrc.py
```diff
@@ -122,20 +148,23 @@ def _section_name(credentials_: Credentials) -> str:
 
 def store_credentials(
         credentials: Credentials,
+        default_provider: Optional[str] = None,
```
jyu00, line 151: This should take a `HubGroupProject` instead of a `str`, as it should be this module/method's responsibility to know what the stored format is, not the caller's.

qiskit/providers/ibmq/ibmqfactory.py
jyu00: What I meant was that since this block comment is meant to describe the _the entire IBMQFactory class_, it should contain more information than just how to save a default provider. Like it should explain what this class is used for and the terminologies (token, provider, configuration file, etc). Perhaps this can go into a different doc PR.
<br>
Note that my original comment regarding the docstring was
>Perhaps a nice, concise description, of how to use IBMQFactory to manage accounts can go into IBMQFactory's docstring. 
<br>
Also while it doesn't hurt to give users more information, I don't think it's necessary to describe in details how to specify a default provider.

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -60,20 +85,29 @@ def enable_account(
             the legacy Quantum Experience and Qconsole (also known as the
             IBM Quantum Experience v1) credentials is no longer supported.
 
+            If using a default provider, all three parameters `hub`, `group`, `project`
+            must be specified.
```
jyu00, line 89: Why do all three parameters need to be specified for `enable_account()`? Since this method loads all the `AccountProvider` instances, it should be able to just return the provider found by `get_provider()` (which doesn't require all three parameters).

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -102,7 +136,27 @@ def enable_account(
...
+
+        # If `hub`, `group`, `project` are all specified, attempt to get the corresponding provider.
+        if all([hub, group, project]):
+            try:
+                default_provider = self.get_provider(hub=hub, group=group, project=project)
+            except IBMQProviderError as ex:
+                raise IBMQAccountError('The default provider (hub/group/project) '
+                                       'to use could not be found:'
+                                       'hub = "{}", group = "{}", project = "{}"'
+                                       .format(hub, group, project)) from ex
```
jyu00, line 157: Why convert `IBMQProviderError`, which is already a subclass of `IBMQAccountError`, to a new `IBMQAccountError`?

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -121,7 +175,8 @@ def load_account(self) -> Optional[AccountProvider]:
         """Authenticate against IBM Quantum Experience from stored credentials.
 
         Returns:
-            The provider for the default open access project.
+            If the configuration file specifies a provider, it is returned.
```
jyu00, line 178: Can you change it to `default provider`, just to be clear.

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -130,9 +185,13 @@ def load_account(self) -> Optional[AccountProvider]:
                 credentials are found.
             IBMQAccountCredentialsInvalidUrl: If invalid IBM Quantum Experience
                 credentials are found.
+            IBMQAccountError: If the default provider stored on disk could not be found.
+            HubGroupProjectInvalidStateError: If the default provider stored on disk could
```
jyu00, line 189: `HubGroupProjectInvalidStateError` should probably be converted to an `IBMQAccountError`, if that's the high level exception class used in this module.

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -168,30 +227,56 @@ def load_account(self) -> Optional[AccountProvider]:
...
+
+        # If specified, attempt to get the provider stored for the account.
+        if stored_provider_hgp:
+            hub, group, project = stored_provider_hgp.to_tuple()
+            try:
+                default_provider = self.get_provider(hub=hub, group=group, project=project)
+            except IBMQProviderError as ex:
+                raise IBMQAccountError('The default provider (hub/group/project) stored on '
+                                       'disk could not be found: '
+                                       'hub = "{}", group = "{}", project = "{}"'
```
jyu00, line 241: It's probably helpful to tell the users that they can overwrite the default provider stored on the disk with `save_account(overwrite=True)` in the error message.

docs/tutorials/1_the_ibmq_account.ipynb
```diff
@@ -76,7 +76,7 @@
     "The `IBMQ` account has functions for handling administrative tasks. The credentials can be saved to disk, or used in a session and never saved.\n",
     "\n",
     "- `enable_account(TOKEN)`: Enable your account in the current session.\n",
-    "- `save_account(TOKEN)`: Save your account to disk for future use.\n",
+    "- `save_account(TOKEN, HUB, GROUP, PROJECT)`: Save your account to disk for future use and optionally specify a default provider to use when loading your account.\n",
```
jyu00, line 79: Why are HUB/GROUP/PROJECT mentioned for `save_account()` but not `enable_account()`?

docs/tutorials/1_the_ibmq_account.ipynb
```diff
@@ -125,7 +125,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "where we have assumed that the user has stored their IQX account information locally ahead of time using `IBMQ.save_account(TOKEN)`. `TOKEN` here is the API token you obtain from your IQX account.\n",
+    "where we have assumed that the user has stored their IQX account information locally ahead of time using `IBMQ.save_account(TOKEN, HUB, GROUP, PROJECT)`. `TOKEN` here is the API token you obtain from your IQX account. `HUB`, `GROUP`, and `PROJECT` are optional parameters that allow you to specify a default provider you would like to save, along with your IQX account. This provider will be returned when loading your account.\n",
```
jyu00, line 128: Given the sample code at line 119 uses `IBMQ.save_account(TOKEN)`, it's probably better to not mention HUB/GROUP/PROJECT here to avoid confusion, especially since h/g/p was already explained at line 79.

qiskit/providers/ibmq/ibmqfactory.py
luisg5: I was re-raising in order to add a bit more context to the exception message. But, you're right, it might not be necessary, since the one provided by `IBMQProviderError` should be straightforward enough when enabling an account.

qiskit/providers/ibmq/ibmqfactory.py
luisg5: The idea was to make sure the user does not accidentally receive a provider they didn't exactly specify. But, it does seem a bit limiting, especially since not all the arguments are required by `get_provider`. I'll go ahead and change that.

luisg5:
Also, as a note. Per your suggestion, I'll create another PR to handle the documentation for `IBMQFactory` 👍

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -164,34 +192,61 @@ def load_account(self) -> Optional[AccountProvider]:
...
+        # The provider for the default open access project.
+        default_provider = providers[0]
+
+        # If specified, attempt to get the provider stored for the account.
+        if stored_provider_hgp:
+            hub, group, project = stored_provider_hgp.to_tuple()
+            try:
+                default_provider = self.get_provider(hub=hub, group=group, project=project)
+            except IBMQProviderError as ex:
+                raise IBMQAccountError('The default provider (hub/group/project) stored on '
```
jyu00, line 207: If you're going to re-raise to add more context, can you keep it as the more specific `IBMQProviderError` (and maybe ditch the `from` as it doesn't add any new information).

Issue #591, opened 2020-03-16
Use newer mypy


jyu00:
### What is the expected behavior?
We currently pin the `mypy` version to 0.720. The reason being we didn't want to introduce too many uncertainties when we first started implementing type hints. Now that we have it for a while, we should use the latest version as it can find additional issues.

PR #613, opened 2020-03-27
Upgrade mypy

luisg5:
### Summary

Addresses #591 

### Details and comments

Unpins `mypy` in `requirements-dev.txt` and fixes issues that arose from upgrading. The errors reported after upgrading were related to:
- using mypy error codes that have been changed/updated.
- Not specifying type comments for unknown types.
- New checks added to the newer mypy.


qiskit/providers/ibmq/accountprovider.py
```diff
@@ -141,7 +141,7 @@ def _discover_remote_backends(self, timeout: Optional[float] = None) -> Dict[str
 
         return ret
 
-    def __eq__(  # type: ignore[overide]
+    def __eq__(  # type: ignore[override]
```
luisg5: Funny how this worked in the previous version and wasn't caught... lol.

qiskit/providers/ibmq/api/clients/websocket.py
```diff
@@ -114,7 +114,7 @@ def from_bytes(cls, json_string: bytes) -> 'WebsocketResponseMethod':
             parsed_dict = json.loads(json_string.decode('utf8'))
         except (ValueError, AttributeError) as ex:
             exception_to_raise = WebsocketIBMQProtocolError(
-                'Unable to parse the message received from the server: {}'.format(json_string))
+                'Unable to parse the message received from the server: {!r}'.format(json_string))
```
luisg5: Mypy reports the following error when formatting bytes:
```
On Python 3 '{}'.format(b'abc') produces "b'abc'"; use !r if this is a desired behavior
```

Specifying `!r` seemed like the appropriate action to take as a result.

qiskit/providers/ibmq/credentials/__init__.py
```diff
@@ -81,12 +81,12 @@ def discover_credentials(
         ('environment variables', (read_credentials_from_environ, {})),
         ('qiskitrc', (read_credentials_from_qiskitrc,
                       {'filename': qiskitrc_filename}))
-    ])
+    ])  # type: MutableMapping[str, Any]
```
jyu00: Why is this a `MutableMapping` and not `Dict`?

luisg5: At first, I was trying to be more specific with the return type and `mypy` would complain about `Dict`, because it expected tuples as the entries. `MutableMapping` was a suggested solution. After changing the type comment to something simpler (the actual type would be way too long and it is also mentioned in line 77), I left it as `MutableMapping`. It looks like using `Dict` is no issue if the type is simple.

Overlooked it, but we could just use `OrderedDict` as the type, since it is imported...

jyu00: In that case line 75 should also be `OrderedDict`?

Issue #584, opened 2020-03-11
Add tests for different providers


jyu00:
### What is the expected behavior?
Currently all tests run under the same provider, as specified by `QE_HGP`. We should have tests that go through the basic paths using both open and premium providers.

PR #618, opened 2020-04-06
Add tests different providers

luisg5:
### Summary

Addresses #584


test/ibmq/test_basic_server_paths.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def test_retrieving_jobs(self):
+        """Test retrieving jobs with filters."""
+        backend_name = 'ibmq_qasm_simulator'
+        past_month = datetime.now() - timedelta(days=30)
+        past_month_str = past_month.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
+        db_filter = {
+            'backend.name': backend_name,
+            'status': {'regexp': '^ERROR'}
+        }
```
jyu00, line 79: You now have a dependency that the test is only useful if there were failed jobs in the past month. Since this is only to test basic endpoints, it doesn't seem that this dependency is necessary.

test/ibmq/test_basic_server_paths.py
```diff
@@ -0,0 +1,129 @@
...
+                    defaults = backend.defaults()
+                    if backend.configuration().open_pulse:
+                        self.assertIsNotNone(defaults)
+                    else:
+                        self.assertIsNone(defaults)
+
+    def test_devices_status_and_job_limit(self):
+        """Test the status and job limit for devices."""
+        for desc, provider in self.providers.items():
+            for backend in provider.backends():
```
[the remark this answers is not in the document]
jyu00: No that's unnecessary.

test/ibmq/test_basic_server_paths.py
```diff
@@ -0,0 +1,129 @@
...
+                                    'Job {} creation date "{}" is not less than '
+                                    'or equal to the past month "{}".'
+                                    .format(job_id, job.creation_date(), past_month_str))
+
+    def test_devices_properties_and_defaults(self):
+        """Test the properties and defaults for devices."""
+        for _desc, provider in self.providers.items():
+            for backend in provider.backends():
+                provider_backend = {'provider': provider, 'backend': backend}
+                with self.subTest(provider_backend=provider_backend):
```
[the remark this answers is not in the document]
jyu00: That's unnecessary. You can just find a backend that is not a simulator and supports open pulse.

test/ibmq/test_basic_server_paths.py
```diff
@@ -0,0 +1,129 @@
...
+                # Fetch the results.
+                result = job.result()
+                self.assertTrue(result.success)
+
+                # Fetch the qobj.
+                qobj_downloaded = job.qobj()
+                self.assertEqual(qobj_downloaded, qobj.to_dict())
+
+    def test_job_status_and_properties(self):
+        """Test the status and properties of a job."""
```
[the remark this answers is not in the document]
luisg5: I'll update the docstring to be more specific, thanks!

test/ibmq/test_basic_server_paths.py
```diff
@@ -0,0 +1,106 @@
...
+    def test_device_status_and_job_limit(self):
+        """Test the status and job limit for a device."""
+        for desc, provider in self.providers.items():
+            backend = provider.backends(simulator=False)[0]
+            provider_backend = {'provider': provider, 'backend': backend}
+            with self.subTest(provider_backend=provider_backend):
+                self.assertTrue(backend.status())
+                job_limit = backend.job_limit()
+                if desc == 'public_provider':
+                    self.assertEqual(job_limit.maximum_jobs, 5)
```
luisg5, line 105: Should the `maximum_jobs` limit for the public provider be checked? 

I did not check the `maximum_jobs` limit of the premium provider, since the premium provider may change based on the credentials saved in the CI.

jyu00: Doesn't hurt to check that, since we don't actually have a separate test case to verify a real job limit.

jyu00: Doesn't hurt to check that, since we don't have a separate test case to verify a real job limit.

test/ibmq/test_basic_server_paths.py
```diff
@@ -0,0 +1,117 @@
...
+            with self.subTest(provider=provider):
+                backend = provider.get_backend(backend_name)
+                qobj = bell_in_qobj(backend)
+
+                job = backend.run(qobj, validate_qobj=True)
+                job_id = job.job_id()
+
+                # TODO No need to wait for job to run once api is fixed
+                while job.status() not in JOB_FINAL_STATES + (JobStatus.RUNNING,):
+                    time.sleep(0.5)
```
jyu00, line 86: This is only an issue if you're querying for a specific job status (which you're not)

test/ibmq/test_basic_server_paths.py
```diff
@@ -0,0 +1,117 @@
...
+                qobj = bell_in_qobj(backend)
+                job = backend.run(qobj, validate_qobj=True)
+
+                # Fetch the results.
+                result = job.result()
+                self.assertTrue(result.success)
+
+                # Fetch the qobj.
+                qobj_downloaded = job.qobj()
+                self.assertEqual(qobj_downloaded, qobj.to_dict())
```
jyu00, line 55: How does this work since you're comparing a `Qobj` instance (as returned by `job.qobj()`) to a `dict` type?

test/ibmq/test_basic_server_paths.py
```diff
@@ -0,0 +1,117 @@
...
+        cls.providers = providers  # Dict[str, AccountProvider]
+
+    @slow_test
+    def test_job_submission(self):
+        """Test running a job against a device."""
+        for _desc, provider in self.providers.items():
+            backend = least_busy(provider.backends(
+                simulator=False, filters=lambda b: b.configuration().n_qubits >= 5))
+            provider_backend = {'provider': provider, 'backend': backend}
+            with self.subTest(provider_backend=provider_backend):
```
jyu00, line 45: Probably not that big a deal, but I think this will expose the premium h/g/p in the Travis log? Why not use `_desc` instead of `provider` in `provider_backend`?

luisg5: Yeah, that's a good point! I'll go ahead and change that.

test/ibmq/test_basic_server_paths.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @slow_test
+    def test_job_submission(self):
+        """Test running a job against a device."""
+        for desc, provider in self.providers.items():
+            backend = least_busy(provider.backends(
+                simulator=False, open_pulse=False,
+                filters=lambda b: b.configuration().n_qubits >= 5))
+            provider_backend = {'provider': desc, 'backend': backend}
+            with self.subTest(provider_backend=provider_backend):
```
jyu00, line 43: nit - instead of putting `desc` and `backend` in a dictionary, it formats better to just do `self.subTest(desc=desc, backend=backend)`

Issue #585, opened 2020-03-11
Convert timestamps to local time


jyu00:
### What is the expected behavior?
There are many timestamps returned today (e.g. job creation date, estimated start time, etc). These are all in UTC time, as that's returned by the server. It'd be nice to convert them to local time so the users don't have to. We need to check this doesn't break the jupyter tools though.

nonhermitian:
This tool already exists in the provider

PR #622, opened 2020-04-10
Convert UTC date times to local time

luisg5:
### Summary

Addresses #585 

### Details and comments

This PR does not store the local date times within a class. Rather, it does the UTC to local conversion with each method call that is supposed to return a date time(s). This allows date time attributes to be stored in UTC, which is useful since conversions are done using UTC, such as calculating the estimated time a job will start or complete.

Note: Update changelog.


qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -428,12 +429,13 @@ def queue_info(self) -> Optional[QueueInfo]:
...
 
     def creation_date(self) -> str:
-        """Return job creation date.
+        """Return job creation date, in local time.
 
         Returns:
             Job creation date.
         """
-        return self._creation_date.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
+        creation_date_local_dt = utc_to_local(self._creation_date)
```
jyu00, line 437: You need to issue a warning to inform the user of time zone change

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -476,7 +479,14 @@ def time_per_step(self) -> Optional[Dict]:
         """
         if not self._time_per_step or self._status not in JOB_FINAL_STATES:
             self.refresh()
-        return self._time_per_step
+        time_per_step_local = {}
+        if self._time_per_step:
+            for step_name, time_data in self._time_per_step.items():
+                time_data_utc_dt = str_to_datetime(time_data)
```
jyu00, line 485: Doesn't `utc_to_local()` take a str type?

qiskit/providers/ibmq/job/queueinfo.py
```diff
@@ -110,17 +113,17 @@ def format(self) -> str:
         """
         status = api_status_to_job_status(ApiJobStatus(self._status)).value \
             if self._status else self._get_value(self._status)
-        estimated_start_time = arrow.get(self.estimated_start_time).humanize() \
+        est_start_time = start_duration(self.estimated_start_time) \
```
jyu00, line 116: Why did you change the arrow humanize format to start_duration format?

qiskit/providers/ibmq/utils/converters.py
```diff
@@ -42,6 +44,30 @@ def utc_to_local(utc_dt: Union[datetime.datetime, str]) -> datetime.datetime:
...
+def str_to_datetime(date_time: str) -> datetime.datetime:
+    """Convert a formatted string representing a date time to a datetime object.
+
+    Args:
+        date_time: Input string to convert into a datetime object.
+
+    Returns:
+        The datetime representing the string.
+    """
+    return datetime.datetime.strptime(date_time, DATETIME_TO_STR_FORMATTER)
```
jyu00, line 68: Why not use `dateutil.parser`? It gives you a lot more flexibility than parsing just a specific format.

test/ibmq/test_ibmq_job.py
```diff
@@ -399,57 +400,67 @@ def test_retrieve_jobs_queued(self, provider):
     def test_retrieve_jobs_start_datetime(self, provider):
         """Test retrieving jobs created after a specified datetime."""
         backend = provider.get_backend('ibmq_qasm_simulator')
-        past_month = datetime.now() - timedelta(days=30)
-        past_month_str = past_month.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
+        past_month = datetime.utcnow() - timedelta(days=30)
```
jyu00, line 403: If you're returning job creation date in local timezone, you should also allow `.jobs()` filtering with local timezone.

test/ibmq/test_ibmq_job_attributes.py
```diff
@@ -239,21 +271,21 @@ def _mocked__api_job_submit(*args, **kwargs):
...
         queue_info = None
         for _ in range(10):
             queue_info = job.queue_info()
-            # Even if job status is QUEUED, queue information may not be
-            # immediately available.
-            if (job._status is JobStatus.QUEUED and job.queue_position() is not None) or \
-                    job._status in leave_states:
+            # Even if job status is queued, its queue info may not be immediately available.
+            if (job._status in queue_states and job.queue_position() is not None) or \
+                    job._status in JOB_FINAL_STATES:
```
jyu00, line 284: I don't understand the logic here. The old code says "if job is QUEUED and has queue position, or if the job progressed beyond QUEUED state, then break". The new one says "if job is QUEUED and RUNNING and has queue position"? But the job should not be in RUNNING if there's a queue position?

test/ibmq/test_ibmq_job_attributes.py
```diff
@@ -263,8 +295,21 @@ def test_queue_info(self, provider):
...
+            self.assertTrue(queue_info_repr)
+
+            # Convert the utc start and end times to local time, represented as a string.
+            est_start_time_local_str = datetime_to_str(
+                utc_to_local(queue_info.estimated_start_time))
+            est_complete_time_local_str = datetime_to_str(
+                utc_to_local(queue_info.estimated_complete_time))
+
+            # Ensure the local start and end date times are in the `repr` format.
+            self.assertIn(est_start_time_local_str, queue_info_repr)
```
jyu00, line 311: What's the point of the doing the assert if you're just calling the same exact functions as the mainline code?

qiskit/providers/ibmq/job/queueinfo.py
luisg5: Arrow humanize does the same thing that `start_duration()` does. They both find out how long it will be until the job is run, calculating the difference. On top of that, I replaced it since it removes the `arrow` dependency.

One thing I realize, however. If `start_duration()` is to be used, the name might need to be changed, since it's used on line `118` to find the approximate complete time of the job. Maybe a name such as `duration_difference`?

test/ibmq/test_ibmq_job_attributes.py
luisg5: I mistook it for the case where the JobStatus may display as RUNNING, but the job is actually QUEUED. I'll change it back to the original logic.

qiskit/providers/ibmq/job/ibmqjob.py
luisg5: Yeah, it does. I was using `str_to_datetime` since the string was in a specific format. But, as mentioned `dateutil.parser` might be better, which `utc_to_local` uses. So, I'll use `utc_to_local` 👍

test/ibmq/test_ibmq_job.py
```diff
@@ -400,16 +401,19 @@ def test_retrieve_jobs_start_datetime(self, provider):
...
                                           limit=5, skip=0, start_datetime=past_month)
         self.assertTrue(job_list)
-        for i, job in enumerate(job_list):
-            self.assertTrue(job.creation_date() >= past_month_str,
-                            '{}) job creation_date {} is not '
-                            'greater than or equal to past month: {}'
-                            .format(i, job.creation_date(), past_month_str))
+        for job in job_list:
+            job_id = job.job_id()
+            with self.subTest(job_id=job_id):
```
jyu00, line 411: Do you really need subTest here? You're just verifying the results, not testing with new parameters.

qiskit/providers/ibmq/job/queueinfo.py
jyu00: The output format is slightly different though. The arrow output is more humanized and matches what IQX displays. But I suppose it's not a big deal.

qiskit/providers/ibmq/ibmqbackendservice.py
```diff
@@ -198,14 +202,23 @@ def jobs(
...
-            api_filter['creationDate'] = {'gte': start_datetime.isoformat()}
-        elif end_datetime:
-            api_filter['creationDate'] = {'lte': end_datetime.isoformat()}
+        # TODO: Remove when decided the warning is no longer needed.
+        if start_datetime or end_datetime:
+            warnings.warn('The parameters `start_datetime` and `end_datetime` are expected '
+                          'to be in local time now, rather than UTC.', stacklevel=2)
+
+            # Attempt to convert the input, local datetime objects to UTC.
+            start_datetime_utc = local_to_utc(start_datetime) if start_datetime else None
```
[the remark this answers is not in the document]
luisg5: Should an exception be raised if the times do not come with a timezone info?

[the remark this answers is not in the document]
luisg5: Would it be best to raise an exception if the datetime does not come with timezone info, or should it be assumed that datetime objects without timezone info are in local time?

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -429,12 +430,16 @@ def queue_info(self) -> Optional[QueueInfo]:
...
 
         Returns:
             Job creation date.
         """
-        return self._creation_date.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
+        creation_date_local_dt = utc_to_local(self._creation_date)
+        # TODO: Remove when decided the warning is no longer needed.
+        warnings.warn('The creation date is returned in local time now, '
+                      'rather than UTC.', stacklevel=2)
+        return datetime_to_str(creation_date_local_dt)
```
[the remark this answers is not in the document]
luisg5: Terra makes use of creation date, so a patch might need to be added there. I'll link the PR in this one when I create it 👍

qiskit/providers/ibmq/ibmqbackendservice.py
```diff
@@ -198,14 +204,27 @@ def jobs(
...
-            }
-        elif start_datetime:
-            api_filter['creationDate'] = {'gte': start_datetime.isoformat()}
-        elif end_datetime:
-            api_filter['creationDate'] = {'lte': end_datetime.isoformat()}
+        # TODO: Remove when decided the warning is no longer needed.
+        if start_datetime or end_datetime:
+            warnings.warn('The parameters `start_datetime` and `end_datetime` are expected '
+                          'to be in local time now, rather than UTC. If the datetime is in '
+                          'UTC, it will be converted to local time.', stacklevel=2)
```
jyu00, line 211: Probably want to mention that the conversion only happens if timezone information is present.

qiskit/providers/ibmq/ibmqbackendservice.py
```diff
@@ -198,14 +204,27 @@ def jobs(
...
+
+            # If the datetime timezone info is not UTC, then convert the datetime into UTC.
+            # Note: datetime objects without a timezone are considered to be in local time.
+            if start_datetime and (start_datetime.tzinfo not in (tz.UTC, timezone.utc)):
```
jyu00, line 215: There are other implementations of `tzinfo` that people might use (e.g. `pytz`). Rather than checking if it's one in the set, it's probably more accurate to check for `utcoffset()`.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -429,20 +434,29 @@ def time_per_step(self) -> Optional[Dict]:
...
             self.refresh()
-        return self._time_per_step
+        time_per_step_local = {}
+        if self._time_per_step:
+            warnings.warn('The time per step date and time information is returned in '
+                          'local time now, rather than UTC.', stacklevel=2)
+            for step_name, time_data_utc in self._time_per_step.items():
+                time_per_step_local[step_name] = utc_to_local(time_data_utc)
+
+        return time_per_step_local
```
jyu00, line 459: should return `None` if information is not available

qiskit/providers/ibmq/utils/converters.py
```diff
@@ -36,12 +39,63 @@ def utc_to_local(utc_dt: Union[datetime.datetime, str]) -> datetime.datetime:
...
+
+    Args:
+        date_time: Input `datetime` object.
+
+    Returns:
+        A string representation of the `datetime`.
+    """
+    if date_time.tzinfo in (datetime.timezone.utc, tz.UTC):
+        return date_time.strftime(UTC_DATETIME_TO_STR_FORMATTER)
+    return date_time.strftime(LOCAL_DATETIME_TO_STR_FORMATTER)
```
jyu00, line 84: Any reason you use your own formatting rather than `isoformat()` ?

luisg5: At first, I used a formatter since there was a specific format the time data was returned in. So, this function was meant to do the reverse. But, we use `dateutil.parser` to handle conversions, which allows date time conversion to be more flexible. So, it would be better to use `isoformat()`, as was used before 👍

test/ibmq/test_ibmq_job.py
```diff
@@ -400,16 +401,16 @@ def test_retrieve_jobs_start_datetime(self, provider):
         """Test retrieving jobs created after a specified datetime."""
         backend = provider.get_backend('ibmq_qasm_simulator')
         past_month = datetime.now() - timedelta(days=30)
-        past_month_str = past_month.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
+        past_month = past_month.replace(tzinfo=tz.tzlocal())
```
jyu00, line 404: Why did you add `tzinfo` to all these test cases? The users are more likely to pass in a datetime without `tzinfo`.

luisg5: I added `tzinfo` since it was needed for comparison against the date returned by `creation_date` (which is tz aware).

Since users are most likely to use non timezone aware date times, I'll search for the jobs with a non timezone aware datetime and compare using the datetime with `tzinfo`.

Issue #625, opened 2020-04-13
New test doesn't run on staging


jyu00:
### Information

- **Qiskit IBMQ Provider version**:
- **API URL or version**:

- **Python version**:
- **Operating system**:

### What is the current behavior?
PR #611 introduced `test_load_account_saved_provider_invalid_format`. The new test does a `save_account()` which doesn't work on staging.

PR #627, opened 2020-04-13
fix test for staging

luisg5:
### Summary

Addresses #625


test/ibmq/test_ibmq_factory.py
```diff
@@ -269,7 +268,7 @@ def test_load_account_saved_provider_invalid_format(self, qe_token, qe_url):
                 with no_file('Qconfig.py'), custom_qiskitrc() as temp_qiskitrc, \
                         no_envs(CREDENTIAL_ENV_VARS):
                     # Save the account.
-                    self.factory.save_account(token=qe_token, url=qe_url)
+                    self.factory.save_account(token=self.token, url=AUTH_URL)
```
luisg5, line 271: The test does not do validation on the API, so it doesn't require the actual API token.

Issue #553, opened 2020-02-19
Remove marshmallow 


jyu00:
### What is the expected behavior?
`qiskit-terra` is moving away from marshmallow due to performance. The provider will need to stop using terra's marshmallow schemas as well as removing its own, since it relies on terra's methods and for consistency. 

This needs to be in sync with https://github.com/Qiskit/qiskit-terra/pull/3383 which removes schemas like `QobjSchema` used by the provider.

Tasks:
- [x] Remove marshmallow from ibmq-provider
- [x] Custom encoder to convert objects to JSON serializable
- [x] Custom decoder to convert
    - [x] list back to complex
    - [x] string back to datetime 
- [x] Test cases to verify en/decoder

mtreinish:
I started this in: https://github.com/Qiskit/qiskit-ibmq-provider/pull/555

jyu00:
One thing to note is that for complex numbers, marshmallow used to convert them from list to complex when a `from_dict()` is done. This now needs to be done in the provider when it deserializes the data from the server.

PR #632, opened 2020-04-17
Remove marshmallow

jyu00:
### Summary
Addresses #553 . Remove usage of marshmallow from the provider.


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -281,7 +273,7 @@ def status(self) -> BackendStatus:
 
         try:
             return BackendStatus.from_dict(api_status)
```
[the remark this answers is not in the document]
jyu00: Yes `ModelValidationError` is a class in terra that's being removed by Qiskit/qiskit-terra#4030. It's preemptively removed here because the odds of someone using terra master with `ModelValidationError` removed and ibmq-provider that still uses it, is much greater than a `ModelValidationError` actually being raised.

qiskit/providers/ibmq/job/queueinfo.py
```diff
@@ -108,7 +104,7 @@ def format(self) -> str:
         Returns:
              The job queue information report.
         """
-        status = api_status_to_job_status(ApiJobStatus(self._status)).value \
+        status = api_status_to_job_status(self._status).value \
```
luisg5, line 107: Nit: Only for consistency. The `.value` here could be set to `.name` or the `.name` in `line 81` could be set to `.value`.

jyu00: This is actually done on purpose. A `__repr__` is the representation of the instance, hence something like `status=RUNNING` seems more suited than `status=job is actively running`. The `format()` method, on the other hand, gives a user friendly formatted output, so it uses `status: job is actively running`.

Issue #573, opened 2020-03-02
Fix all Sphinx warnings


jyu00:
### What is the expected behavior?
Sphinx does not issue any warning when generating docs.

PR #634, opened 2020-04-21
Fix sphinx warnings

luisg5:
### Summary

Addresses #573


docs/apidocs/ibmq-provider.rst
```diff
@@ -1,4 +1,4 @@
-.. _qiskit-providers-ibmq:
+.. module:: qiskit
```
luisg5, line 1: The following error stemmed from `docs/apidocs/ibmq-provider.rst` sharing the same label as `docs/apidocs/ibmq_provider.rst`:

```
.../qiskit-ibmq-provider/qiskit/providers/ibmq/__init__.py:docstring of qiskit.providers.ibmq:1: WARNING: duplicate label qiskit-providers-ibmq, other instance in .../docs/apidocs/ibmq-provider.rst
```

As a result, I switched the label to `qiskit`, which is the label used for qiskit-terra and qiskit-aqua.

tox.ini
```diff
@@ -36,15 +36,17 @@ envdir = .tox/docs
 deps =
   -r requirements-dev.txt
   sphinx-intl
-commands = sphinx-build -b html {posargs} {toxinidir}/docs {toxinidir}/docs/_build/html
+  qiskit-aer
+commands = sphinx-build -W -b html {posargs} {toxinidir}/docs {toxinidir}/docs/_build/html
```
luisg5, line 40: The `W` flag treats warnings as failures when building the docs. Since the sphinx warnings are fixed with this PR, then setting the flag is likely ideal.

tox.ini
```diff
@@ -36,15 +36,17 @@ envdir = .tox/docs
 deps =
   -r requirements-dev.txt
   sphinx-intl
-commands = sphinx-build -b html {posargs} {toxinidir}/docs {toxinidir}/docs/_build/html
+  qiskit-aer
```
luisg5, line 39: An alternative to setting `qiskit-aer` as a requirement when building the docs would be to set the environment variable `QISKIT_SUPPRESS_PACKAGING_WARNINGS=Y` under `[testenv:docs]` (i.e.):

```
setenv =
  QISKIT_SUPPRESS_PACKAGING_WARNINGS=Y
```

Setting the environment variable would suppress the warnings `Could not import Aer provider...` and not require installing `qiskit-aer` when building the docs.

jyu00: How long does the doc build take if `qiskit-aer` is included?

luisg5: When running locally, I don't notice much of a difference. The average time it takes to build, whether aqua is included or not, is about 78 seconds. When aqua is included, it takes about an extra second to build lol.

PR #687, opened 2020-06-11
ManagedResult.combine_results() that returns combined Result

### LINKED ISSUE #681: Allow Job Manager to return a Result object
A use case described here: https://quantumcomputing.stackexchange.com/questions/12277/how-to-obtain-the-density-matrix-using-tomography-in-the-real-device

In a nutshell, the user wanted to feed the results from managed jobs into `StateTomographyFitter`, which only takes a `Result`. `ManagedResults` perhaps can have a `to_result()` method that copies data and returns a `Result` with all the combined results.

jyu00:
### Summary
Closes #681. 
This PR introduces a `ManagedResult.combine_results()` method that combines all job results and returns a single `Result` object. This allows users to use Job Manager for large experiments and still use ignis fitter methods that expect a `Result`:

```
job_manager = IBMQJobManager()
job_set = job_manager.run(lots_circuits, backend=backend)
combined_result = job_set.results().combine_results()
state_fit = StateTomographyFitter(combined_result, lots_circuits)
```

### DIFF
--- a/qiskit/providers/ibmq/managed/managedjob.py
+++ b/qiskit/providers/ibmq/managed/managedjob.py
@@ -162,13 +162,16 @@ class ManagedJob:
     def result(
             self,
             timeout: Optional[float] = None,
-            partial: bool = False
+            partial: bool = False,
+            refresh: bool = False
     ) -> Optional[Result]:
         """Return the result of the job.
 
         Args:
            timeout: Number of seconds to wait for job.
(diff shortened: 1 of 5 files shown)


jyu00:
> looks good, thanks for the quick fix. there's a doc build error but I don't see how it's related to this PR

Yeah it's due to the new Sphinx version. #689 pins the version temporarily until I can fix the warnings.

PR #727, opened 2020-07-29
[WIP] Move backend modules to backend subpackage

jyu00:
### Summary
This PR moves backend modules, including `ibmqbackend`, `ibmqbackendservice`, and `backendjoblimit` to its own `backend` subpackage. Backend-related exception classes are also moved to `backend/exceptions`. 

The motivation is to make the code better organized and pave the way for more backend related modules to be added.


jyu00:
Need to move `IBMQBackend` back so it doesn't break terra & aqua

jyu00:
Close due to potentially breaking compatibility. Will reintroduce when we do v2 provider.

PR #763, opened 2020-09-17
Add error_string to job.error_message

nonhermitian:
### Summary

Creates `IBMQErrorString` class that is a simple wrapper of `str` that returns a button in Jupyter that links out to the solution for errors with 4-digit error codes associated with them.  If not on Jupyter, or no 4-digit code exists, the functionality is the exact same a standard string.

Screen Shot 2020-09-17 at 13 44 13

Screen Shot 2020-09-17 at 13 44 48


nonhermitian:
I am not satisfied with this approach

PR #765, opened 2020-09-24
Experiment service

jyu00:
### Summary
This PR introduces the beta version of the IBMQ experiment service.


qiskit/providers/ibmq/api/clients/experiment.py
```diff
@@ -0,0 +1,257 @@
...
+                                     **credentials.connection_parameters())
+        self.base_api = Api(self._session)
+
+    def experiments(
+            self,
+            backend_name: Optional[str],
+            experiment_type: Optional[str] = None,
+            start_time: Optional[List] = None,
+            device_components: Optional[List[str]] = None,
+            tags: Optional[str] = None
```
blakejohnson, line 53: I can only specify a single tag? Or is the data type `Optional[List[str]]`?

qiskit/providers/ibmq/api/clients/experiment.py
```diff
@@ -0,0 +1,257 @@
...
+            self,
+            experiment_id: str,
+            plot: Union[bytes, str],
+            plot_name: str
+    ) -> Dict:
+        """Update an experiment plot.
+
+        Args:
+            experiment_id: Experiment UUID.
+            plot: Plot file name or data to upload.
```
blakejohnson, line 143: Does this actually work with a path in addition to `bytes`?

qiskit/providers/ibmq/api/clients/experiment.py
```diff
@@ -0,0 +1,257 @@
...
+
+        Args:
+            result_id: Analysis result ID.
+
+        Returns:
+            Analysis result data.
+        """
+        return self.base_api.analysis_result(result_id).get()
+
+    def device_components(self, backend_name: Optional[str]) -> List[Dict]:
```
blakejohnson, line 248: Seems like `backend_name` should not be optional.

qiskit/providers/ibmq/api/clients/experiment.py
jyu00: My bad, fixed.

qiskit/providers/ibmq/api/clients/experiment.py
jyu00: You don't want people to be able to query for all device components?

blakejohnson: I don't know what I device component means in the absence of a backend, but maybe someone will come up with a use case for wanting the set of all device components across all backends.

PR #787, opened 2020-10-21
[WIP] Simple backend filtering

nonhermitian:
### Summary
Simplifies backend filtering for most commonly used comparators.  For example, all 5Q devices is:

```python
provider.backends().num_qubits == 5
```

or all real devices is:

```python
provider.backends().real
```

Things can be combined.  For example, all open_pulse capable devices with more than 5Q is:

```python
provider.backends().open_pulse.num_qubits > 5
```

or 

```python
(provider.backends().open_pulse) & (provider.backends().num_qubits > 5)
```
### Details and comments

Still needs to be documented.

### DIFF
--- a/qiskit/providers/ibmq/collections/backend.py
+++ b/qiskit/providers/ibmq/collections/backend.py
@@ -1,0 +1,326 @@
+# -*- coding: utf-8 -*-
+
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2020.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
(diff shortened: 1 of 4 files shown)


nonhermitian:
Ehh, I am not dealing with the Types junk here.  will implement elsewhere

jyu00:
I don't see how this would be simpler than `provider.backends(open_pulse=True, n_qubits=5)` that we have today. However, I do agree we should add a `min_qubits` as another filtering keyword, since people are now forced to `filters=lambda b: b.configuration().n_qubits >= 5` to skip armonk.

PR #788, opened 2020-10-21
Handle new list API response body format

### LINKED ISSUE #782: Handle list API response body changes
## Information

- **Qiskit IBMQ Provider version**: 0.10.0
- **Python version**: n/a
- **Operating system**: n/a

The newly added support for experiments with the resultsdb endpoint in IQX are going to need to handle a response payload change for the following APIs:

- GET /devices
- GET /device_components
- GET /experiments
- GET /analysis_results

The response body on those today is an array but they are going to change to an object with a field that contains the array, e.g.:

```json
[]
```

becomes

```json
{
    "experiments": []
}
```

The field in the response body object is the same as the path resource, so `devices`, `device_components`, `experiments` and `analysis_results`.

### Suggested solutions

From a client perspective there are two ways to handle this really:

1. Proactive: add some compatibility code now to handle the two different forms of responses and then remove the compatibility code at some later date. This could be backported to the 0.10.x series.
2. Reactive: just change and release a new version once the server changes. Note that Travis CI will likely be blocked until this is fixed after the server changes (we could coordinate when the server side change happens if that helps).

mriedem:
### Summary

The `GET` APIs for listing resources has changed from an array
to an object. This handles those changes for devices, device_components,
experiments and analysis_results.

### Details and comments

Closes #782


mriedem:
Pushing this up to see if things pass in CI since I couldn't get things to work locally. I'm also not sure about all of the changelog stuff in the PR template because there is no CHANGELOG file (is the template old?).

mtreinish:
> Pushing this up to see if things pass in CI since I couldn't get tox to work locally. I'm also not sure about all of the changelog stuff in the PR template because there is no CHANGELOG file (is the template old?).

The template is old and needs to be updated. The changelog generation is automatic based on the `Changelog:*` labels on a PR. For release notes you can use reno. https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/CONTRIBUTING.md#changelog-generation

mriedem:
> The template is old and needs to be updated. The changelog generation is automatic based on the `Changelog:*` labels on a PR. For release notes you can use reno. https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/CONTRIBUTING.md#changelog-generation

OK so I'm assuming it's sufficient to just add the `Changelog: Bugfix` label to the PR description and not bother with a release note since this is a pretty minor change - no user facing impact, not a new feature, just a bug fix etc.

mriedem:
Actually I can't apply any labels to this PR myself so 🤷‍♂️ .

jyu00 (approved):

jyu00:
@mriedem I don't think Travis CI was triggered. Can you do an empty commit to trigger it again?

qiskit/providers/ibmq/api/clients/experiment.py
```diff
@@ -64,8 +64,9 @@ def experiments(
         Returns:
             A list of experiments.
         """
-        return self.base_api.experiments(
+        resp = self.base_api.experiments(
             backend_name, experiment_type, start_time, device_components, tags)
+        return resp['experiments']
```
mriedem, line 69: This (and below) is failing mypy but I'm not sure what I need to do here:

https://travis-ci.com/github/Qiskit/qiskit-ibmq-provider/jobs/403781076

```
qiskit/providers/ibmq/api/clients/experiment.py:69: error: No overload variant of "__getitem__" of "list" matches argument type "str"  [call-overload]

qiskit/providers/ibmq/api/clients/experiment.py:69: note: Possible overload variants:

qiskit/providers/ibmq/api/clients/experiment.py:69: note:     def __getitem__(self, int) -> Any

qiskit/providers/ibmq/api/clients/experiment.py:69: note:     def __getitem__(self, slice) -> List[Any]
```

mriedem: I think I get it. The `base_api.experiments` function says it's returning a List but it's actually returning a Dict now so I need to change that for mypy to work.

Issue #794, opened 2020-10-23
POST /experiments (resultsdb) is going to require hub_id/group_id/project_id


mriedem:
### Information

- **Qiskit IBMQ Provider version**: latest (master)
- **Python version**: n/a
- **Operating system**: n/a

### What is the current behavior?

The `POST /experiments` API is going to soon require `hub_id`, `group_id` and `project_id` fields in the request body. If they are not provided a 400 error response will be returned.

Technically that API doesn't validate against additional parameters sent in the request body so this is a change that could be made in the client _before_ the server side API change happens.

### Steps to reproduce the problem

Calling `POST /experiments` without `hub_id`, `group_id`, and `project_id` is going to result in a 400 error response soon.

### What is the expected behavior?

Creating experiments should continue to just work and the client handles the new required parameters.

### Suggested solutions

I'm not super familiar with the qiskit code but I'm assuming the logged in IQX provider information is available to the API client code somewhere and we can just pass these values along from that provider object.

mriedem:
I think this can probably be closed now, I think #796 is handling it.

PR #796, opened 2020-10-26
Pass hub/group/project to POST /experiments

mriedem:
### Summary

In an upcoming change in the `POST /experiments` API, experiments
are going to require storing the hub_id/group_id/project_id. This
change provides those values when creating an experiment.

### Details and comments

Partial #794


mriedem:
@jyu00 I could use some direction on how you'd like to see this handled based on the questions inline.

jyu00:
@mriedem If the idea is to return h/g/p data when retrieving an experiment, I'd store them as part of the experiment - i.e. extract them from provider when an experiment is created and use whatever is stored in the experiment object when uploading.

qiskit/providers/ibmq/experiment/experiment.py
```diff
@@ -95,6 +98,9 @@ def update_from_remote_data(self, remote_data: Dict) -> None:
         self._updated_datetime = str_to_utc(remote_data.get('updated_at', None))
         self._uuid = remote_data['uuid']
         self._plot_names = remote_data.get('plot_names', [])
+        self._hub = remote_data.get('hub_id')
+        self._group = remote_data.get('group_id')
+        self._project = remote_data.get('project_id')
```
jyu00, line 103: Similar to this, `from_remote_data()` should populate the experiment object with remote data instead of using the credentials in the provider.

jyu00 (changes requested):

qiskit/providers/ibmq/experiment/experiment.py
mriedem: Done.

test/ibmq/test_experiment.py
```diff
@@ -199,6 +199,8 @@ def test_retrieve_experiment(self):
         exp = self.experiments[0]
         rexp = self.provider.experiment.retrieve_experiment(exp.uuid)
         self.assertEqual(exp.uuid, rexp.uuid)
+        for attr in ['hub', 'group', 'project']:
+            self.assertIsNotNone(getattr(rexp, attr), "{} does not have a {}}".format(rexp, attr))
```
mriedem, line 203: `test/ibmq/test_experiment.py:203:54: W1302: Invalid format string (bad-format-string)`
```suggestion
            self.assertIsNotNone(getattr(rexp, attr), "{} does not have a {}".format(rexp, attr))
```

jyu00 (approved):

PR #805, opened 2020-10-30
Avoid infinite loop when passing IBMQRandomService to child

### LINKED ISSUE #804: Infinitely loop in IBMQRandomService when using multiprocessing in Python 3.8
### Information

- **Python version**: 3.8
- **Operating system**:

Passing an `IBMQRandomService` instance, or an object that contains it (e.g. `AccountProvider`), to a child process will result in an infinite loop. 

The loop is caused by the following:

- `ProcessPoolExecutor` tries to pickle `IBMQRandomService` by calling its `__getstate__`, and `__getattr__` is called as a result. 
- `__getattr__` looks for `self._initialized`, which it assumes to be there since it's set at `__init__`. But that's not the case here. 
-  As a result `__getattr__`  is called to find `self._initialized`. 

```
def foo2(provider):
    pass

def foo():
    with futures.ProcessPoolExecutor() as executor:
        provider = IBMQ.load_account()
        r = executor.submit(foo2, provider)
    r.result()

if __name__ == "__main__":
    foo()
```

No infinite loop even if it's not pickleable

### Suggested solutions
Explicitly check that `_initialized` exists.

jyu00:
### Summary
Fixes #804 . 
This PR explicitly checks for `_initialized` instead of assuming it's available.

### DIFF
--- a/qiskit/providers/ibmq/random/ibmqrandomservice.py
+++ b/qiskit/providers/ibmq/random/ibmqrandomservice.py
@@ -110,7 +110,8 @@ class IBMQRandomService:
         return self.__dict__
 
     def __getattr__(self, item: str) -> Any:
-        self._discover_services()
+        if not self.__getattribute__('_initialized'):
+            self._discover_services()
         try:
             return self._services[item]
         except KeyError:
--- a/releasenotes/notes/random-loop-85cc24a9ac024a8f.yaml
+++ b/releasenotes/notes/random-loop-85cc24a9ac024a8f.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixes the infinite loop raised when passing an ``IBMQRandomService`` instance
+    to a child process.


qiskit/providers/ibmq/random/ibmqrandomservice.py
```diff
@@ -110,7 +110,8 @@ def __dir__(self) -> Dict:
         return self.__dict__
 
     def __getattr__(self, item: str) -> Any:
-        self._discover_services()
+        if not self.__getattribute__('_initialized'):
+            self._discover_services()
```
mriedem, line 114: This still checks `if not self._initialized:` which will call this code and put you in an infinite loop won't it? Why not just set self._initialized = False in the __init__ method?

jyu00: `self._initialized` **is** set in `__init__()`. The problem is `_initialized` is not present yet when `__getattr__()` is called to find `__getstate__` to pickle the instance to be passed to another process. 

If `_initialized` is a known attribute (i.e. `self.__getattribute__('_initialized')` would return True), then the call `self._initialized` will be directed to `__getattribute__()`, not `__getattr__()`.

PR #807, opened 2020-11-02
Add paging support for experiments and analysis results

### LINKED ISSUE #806: (resultsdb) add paging support for experiments and analysis results

The `GET /experiments` and `GET /analysis_results` APIs now support paging and limit responses to 1000 records by default. This is new in version 0.16.0 of the resultsdb API.

Both requests now accept a `limit` parameter which must be between 1 and 1000 (inclusive) and a `marker`. If a request is limited the response body will contain a `marker` field that is the marker for the next "page" of records and that marker can be provided on a subsequent request.

Also note that experiments and analysis_results are now sorted by `start_time` and `created_at`, respectively, in _descending_ order (the previous default sort order was ascending).

jyu00:
### Summary
Closes #806 .


test/ibmq/test_experiment.py
```diff
@@ -526,6 +526,33 @@ def test_upload_update_plot_data(self):
...
+                self.assertEqual(len(experiments), limit)
+                self.assertEqual(len({expr.uuid for expr in experiments}), len(experiments))
+
+        with self.assertRaises(ValueError) as context_manager:
+            self.provider.experiment.experiments(limit=-1)
+        self.assertIn("limit", str(context_manager.exception))
+
+    def test_analysis_results_limit(self):
+        """Test getting analysis results with a limit."""
+        limits = [10, 1000, 1001]
```
mriedem, line 545: 1001 is going to be capped at 1000 on the server side, per the docs:

https://github.ibm.com/IBM-Q-Software/sw-ibmq-results/blob/staging/docs/api-ref.md#pagination

>limit: Restrict the result set to this many records. Must be at least 1 and at most the configured maximum for the server (default: 1000). If the limit is reached a marker field is returned in the response body which can be be provided using the marker query parameter in subsequent requests.

So how is the assertion on L549 not failing?

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -110,6 +111,7 @@ def experiments(
         """Retrieve all experiments, with optional filtering.
 
         Args:
+            limit: Number of experiments to retrieve. ``None`` indicates no limit.
```
mriedem, line 114: >``None`` indicates no limit.

This is half true, there is a limit enforced on the server side regardless. This is really saying there is no limit imposed from the client side.

test/ibmq/test_experiment.py
mriedem: Oh I get it, it's using the marker to page on the client side, so it's really making two requests, the first is limited to 1000 results and then it uses the marker to get the next page which is going to limit to 1 result from the last marker. Nice.

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -274,20 +287,32 @@ def analysis_results(
...
                 qual_str = qual if op == 'eq' else "{}:{}".format(op, qual)
                 qualit_list.append(qual_str)
-        response = self._api_client.analysis_results(
-            backend_name=backend_name, device_components=device_components,
-            experiment_uuid=experiment_id, result_type=result_type, quality=qualit_list)
         results = []
-        for result in response:
-            results.append(AnalysisResult.from_remote_data(result))
+        marker = None
+        while limit and limit > 0:
```
mriedem, line 306: Same as above, what if limit is None? Do you get anything back? Do you get *everything* back?

mriedem (commented):
Seems there is some missing testing for when limit is explicitly set to None and not the default of 10.

qiskit/providers/ibmq/experiment/experimentservice.py
jyu00: Yes. If the user sets `limit=None`, then qiskit will continue querying until no marker is returned.

qiskit/providers/ibmq/experiment/experimentservice.py
jyu00: Yeah this PR was not as done as I thought. It should be fixed now.

Issue #809, opened 2020-11-05
Migrate to v2 provider interface


jyu00:

jyu00:
Closed by #817.

Issue #800, opened 2020-10-27
Deprecate Qobj validation


jyu00:
### What is the expected enhancement?

`backend.run()` today accepts a `validate_qobj` parameter indicating whether to validate the Qobj. It uses `validate_qobj_against_schema` from Terra, which is being deprecated. Since the default for `validate_qobj` is False, and  Qobj is validated in IQX regardless, we can just deprecate it now and remove later.

PR #817, opened 2020-11-16
Use v2 provider interface

### LINKED ISSUE #808: backend.run() should take a list of circuits
Allow `backend.run()` to take a list of circuits instead of Qobj

jyu00:
### Summary
Closes #808 and #809. 

This PR migrates to using the terra v2 provider base classes. As part of the migration, the `backend.run()` method now accepts circuit(s) and schedule(s) instead of Qobj. 

Note that:
1. Unlike `execute()`, `backend.run()` doesn't transpile the circuits
2. Some of the parameters in `assemble()` today are not user-configurable backend options (e.g. `parametric_pulses`) and therefore are not included in `IBMQBackend._default_options()`. However, `backend.run()` now supports arbitrary kwargs, so the users can continue to specify these parameters.


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -589,22 +712,27 @@ def properties(
...
+            noise_model: Any = None,
+            **kwargs: Dict
     ) -> IBMQJob:
         """Run a Qobj asynchronously.
 
         Args:
-            qobj: The Qobj to be executed.
-            backend_options: Backend options.
-            noise_model: Noise model.
+            qobj: An individual or a
```
nonhermitian, line 730: We did a similar thing in the AQT provider, where it was requested that we make the change qobj -> circuit with a deprecation warning for the kwarg change.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -117,18 +129,54 @@ def __init__(
...
+                       meas_level=MeasLevel.CLASSIFIED,
+                       meas_return=MeasReturnType.AVERAGE,
+                       memory_slot_size=100,
+                       rep_time=None, rep_delay=None,
+                       init_qubits=True)
+
     def run(
             self,
-            qobj: Union[QasmQobj, PulseQobj],
+            qobj: Union[QasmQobj, PulseQobj, QuantumCircuit, Schedule,
```
nonhermitian, line 146: same should be circuit

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -589,22 +712,27 @@ def properties(
         """Return ``None``, simulators do not have backend properties."""
         return None
 
-    def run(
+    def run(    # type: ignore[override]
             self,
-            qobj: Union[QasmQobj, PulseQobj],
+            qobj: Union[QasmQobj, PulseQobj, QuantumCircuit, Schedule,
```
nonhermitian, line 717: should be circuit

nonhermitian (commented):
I think this looks good.  Just a few minor comments.  Also, I thought that changes to the jobs were also in providers v2?  e.g. `job.get_counts()` or `job.circuits()`

mtreinish:
> I think this looks good. Just a few minor comments. Also, I thought that changes to the jobs were also in providers v2? e.g. `job.get_counts()` or `job.circuits()`

The first version of the job object was basically unchanged from the old interface. This was done to simplify the migration from a backwards compat PoV and the thinking was once we have versioned abstract classes we can iterate on the interface more easily. The only real change over the old interface is that it adds explicit support for a sync job, instead of assuming its always async. I'm hoping to have a chance to iterate on the jobs abstract class to the next version (`JobsV2`) that starts to make some of those changes for the next terra release.

qiskit/providers/ibmq/ibmqbackend.py
jyu00: I kept `qobj` because I didn't want to break those doing `backend.run(qobj=qobj)`. Apparently terra has a nice little decorator to handle just this, so it's now `circuits` (plural as it takes multiple).

jyu00:
I added `job.circuits()`, `job.backend_options()`, and `job.header()`, since they're hard to get now without a Qobj. `job.get_counts()` and the likes will go to a separate PR since it's not strictly related to the `JobV1` interface .

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -141,13 +190,48 @@ def run(
                     * none: The job is not shared at any level.
 
                 If the job share level is not specified, the job is not shared at any level.
-            job_tags: Tags to be assigned to the jobs. The tags can subsequently be used
+            job_tags: Tags to be assigned to the job. The tags can subsequently be used
                 as a filter in the :meth:`jobs()` function call.
             validate_qobj: If ``True``, run JSON schema validation against the
-                submitted payload
+                submitted payload. Only applicable if a Qobj is passed in.
```
mtreinish, line 196: We might want to deprecate this. The schema's in terra are deprecated now so if we want to continue to offer this as an option we'll have to port over the schema compilation and validation and use the ibmq schemas repo.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -115,18 +128,54 @@ def __init__(
...
+
+        If a keyword specified here is also present in the ``options`` attribute/object,
+        the value specified here will be used for this run.
 
         Args:
-            qobj: The Qobj to be executed.
+            circuits: An individual or a
+                list of :class:`~qiskit.circuits.QuantumCircuit` or
+                :class:`~qiskit.pulse.Schedule` objects to run on the backend.
+                A :class:`~qiskit.qobj.QasmQobj` or a
```
mtreinish, line 176: Do we want to mention that Qobj support is deprecated here?

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -170,10 +254,45 @@ def run(
             api_job_share_level = ApiJobShareLevel.NONE
 
         validate_job_tags(job_tags, IBMQBackendValueError)
+
+        if isinstance(circuits, (QasmQobj, PulseQobj)):
+            warnings.warn("Passing a Qobj to Backend.run is deprecated and will "
+                          "be removed in a future release. Please pass in circuits "
+                          "or pulse schedules instead. The `qobj` parameter has been "
+                          "renamed to `circuits`.", DeprecationWarning,
```
mtreinish, line 262: You don't need the last sentence here, the deprecate_arguments decorator will raise a separate warning if `qobj` is used as a named parameter.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -683,15 +833,12 @@ def reservations(
...
 
-    def run(
+    def run(    # type: ignore[override]
             self,
-            qobj: Union[QasmQobj, PulseQobj],
-            job_name: Optional[str] = None,
-            job_share_level: Optional[str] = None,
-            job_tags: Optional[List[str]] = None,
-            validate_qobj: bool = False
+            **kwargs: Dict
```
mtreinish, line 838: Won't this fail for cases where qobj was passed in before. It's probably not a big deal except that `run(qobj)` will raise a TypeError now while it would raise an `IBMQBackendError` before

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -170,10 +254,45 @@ def run(
...
             validate_qobj_against_schema(qobj)
         return self._submit_job(qobj, job_name, api_job_share_level, job_tags)
 
+    def _get_run_config(self, **kwargs: Any) -> Dict:
+        """Return the consolidated runtime configuration."""
+        run_config_dict = copy.copy(self.options.__dict__)
+        for key, val in kwargs.items():
+            if val:
+                run_config_dict[key] = val
+        return run_config_dict
```
mtreinish, line 294: I'm wondering if we want to add something here to warn if somebody is setting something that isn't a supported option. I realize you could dump anything in the qobj run_config before, but only the things specified in the options object should be used now having extra fields is non-fatal, but just don't do anything right?

mtreinish (commented):
LGTM, just a few comments and questions inline nothing major or worth blocking though. The only thing I think which is missing here is that the minimum terra version in the requirements list needs to be bumped to 0.16.0 which added the the new interface.

qiskit/providers/ibmq/ibmqbackend.py
jyu00: Yes I opened #800 to deprecate it. Just haven't gotten to it yet.

qiskit/providers/ibmq/ibmqbackend.py
jyu00: It does mention Qobj is deprecated:
```
A :class:`~qiskit.qobj.QasmQobj` or a :class:`~qiskit.qobj.PulseQobj` object is also supported but
                is deprecated.
```
The API doc should be easier to read without all the sphinx tags.

mtreinish: Ah yep, sorry I missed it in there.

mtreinish (approved):
LGTM, thanks for making the updates.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -170,10 +254,47 @@ def run(
             api_job_share_level = ApiJobShareLevel.NONE
 
         validate_job_tags(job_tags, IBMQBackendValueError)
+
+        if isinstance(circuits, (QasmQobj, PulseQobj)):
+            warnings.warn("Passing a Qobj to Backend.run is deprecated and will "
```
ishapoval, line 259: It seems that `ManagedJobSet` was not adapted to the v2 provider interface: it still assembles `qobj` for `ManagedJob`. Thus, with the latest `qiskit-ibmq-provider`, the use of `IBMQJobManager` now yields two related warnings for each and every job submission which with many jobs becomes somewhat overwhelming. Is there hope to get this cleaned up in the next release, please?

PR #823, opened 2020-11-20
Look only for credentials in qiskitrc section "ibmq"

nonhermitian:
### Summary
The config file `qiskitrc` is capable of handling credentials from multiple different providers.  However, this functionality was removed some time ago.  This allows for that functionality to exist once again by having the `qiskit_ibmq_provider` look for credentials only in the `ibmq` section of the config.

### DIFF
--- a/qiskit/providers/ibmq/credentials/configrc.py
+++ b/qiskit/providers/ibmq/credentials/configrc.py
@@ -121,17 +121,12 @@ def write_qiskit_rc(
 
         return credentials_dict
 
-    def _section_name() -> str:
-        """Return the ibmq sction name string"""
-        base_name = 'ibmq'
-        return base_name
-
     filename = filename or DEFAULT_QISKITRC_FILE
     # Create the directories and the file if not found.
     os.makedirs(os.path.dirname(filename), exist_ok=True)
 
     unrolled_credentials = {
-        _section_name():
+        'ibmq':
             _credentials_object_to_dict(credentials_object, default_provider)
         for _, credentials_object in credentials.items()
     }
--- a/qiskit/providers/ibmq/credentials/configrc.py
+++ b/qiskit/providers/ibmq/credentials/configrc.py
@@ -121,8 +121,8 @@ def write_qiskit_rc(
 
         return credentials_dict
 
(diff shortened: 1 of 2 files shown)


qiskit/providers/ibmq/credentials/configrc.py
```diff
@@ -65,8 +65,8 @@ def read_credentials_from_qiskitrc(
     # Build the credentials dictionary.
     credentials_dict = OrderedDict()  # type: ignore[var-annotated]
     default_provider_hgp = None
-    for name in config_parser.sections():
-        single_credentials = dict(config_parser.items(name))
+    if config_parser.has_section('ibmq'):
+        single_credentials = dict(config_parser.items('ibmq'))
```
jyu00, line 69: Before we switched to single token, the section name used to be `ibmq_<hub>_<group>_<project>`, which is still tolerated today. This change would break those who kept the original section name and only updated the url (highly unlikely, but I recently learned someone who hand edited and continued to use `Qconfig.py`!). To be nice, we can change this to something like

```
for name in config_parser.sections():
    if not name.startswith('ibmq'):
        continue
    if name != 'ibmq':
        issue deprecation warning
```

nonhermitian: Yes, but I thought we forced everyone to upgrade their credentials when going to this single login.  It is even still in the readme of this repo: https://github.com/Qiskit/qiskit-ibmq-provider#updating-your-ibm-quantum-experience-credentials

I think we can just look for all sections that start with `ibmq` and handle accordingly.  However, at the same time, this is really legacy code that should not even be supported anymore.

jyu00: We did, but people could have manually edited their `qiskitrc` instead of using the updater and kept the old section name of `ibmq_foo_bar`, which would continue to work just fine. We should have deprecated that when we moved to single login. Since we didn't, I'd feel more comfortable deprecating it first before removing it. And yes it really needs to be removed, along with `Qconfig.py` which is even more ancient.

nonhermitian:
Closing in favor of #826

PR #826, opened 2020-12-02
Deprecate loading from multiple ibmq accounts

nonhermitian:
### Summary
Replaces #823

Looks for `qiskitrc` sections that start with `ibmq`.  If more than one, raises a deprecation warning.

### DIFF
--- a/qiskit/providers/ibmq/credentials/configrc.py
+++ b/qiskit/providers/ibmq/credentials/configrc.py
@@ -16,6 +16,7 @@
 
 import logging
 import os
+import warnings
 from ast import literal_eval
 from collections import OrderedDict
 from configparser import ConfigParser, ParsingError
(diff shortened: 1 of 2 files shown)


qiskit/providers/ibmq/credentials/configrc.py
```diff
@@ -65,29 +66,35 @@ def read_credentials_from_qiskitrc(
...
+                del single_credentials['default_provider']
+
+            new_credentials = Credentials(**single_credentials)  # type: ignore[arg-type]
+
+            credentials_dict[new_credentials.unique_id()] = new_credentials
+
+    if credentials_count > 1:
+        warnings.warn("Loading from multiple accounts is deprecated and "
+                      "will no longer be supported in future releases.",
+                      DeprecationWarning, stacklevel=4)
```
jyu00, line 97: You don't actually need to raise a warning as the existing code already [raises an error](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/ibmqfactory.py#L177). My comment from the other PR was about people using `ibmq_foo_bar` instead of just `ibmq` for as the section name. Nonetheless, this achieves your original goal of allowing credentials from different providers, as long as they don't start with `ibmq`. 

Can you add a release note (e.g. the one from the other PR)?

nonhermitian: Now I am puzzled.  I thought you were concerned because it was not deprecated, but it looks like it has been raising for 17 months.  So while not technically deprecated, anyone who uses it gets raised.  I can add a note and remove the warning.

jyu00: My concerned was about the _name_ of the section, not the number of sections in qiskitrc. The code in your previous PR ignored any section that's not named `ibmq`. And my argument was that the section could be named `ibmq_hub_group_project`, which is the format we used to have and never deprecated.

jyu00 (approved):

PR #829, opened 2020-12-07
Expose Experiment visibility and owner attributes

### LINKED ISSUE #825: ResultsDB: Expose experiment visibility and owner

When creating an experiment in the resultsdb service one can specify a `visibility` field which can be one of:

- `private` (default)
- `project`
- `group`
- `hub`
- `public`

These determine if an experiment is visible or "shared" at a given scope, i.e. if an experiment has visibility="hub" anyone within the same hub can GET the experiment. It would be nice to be able to specify the `visibility` when creating an experiment and to also provide a property on the `Experiment` class for the `visibility` field.

Also, experiments have an "owner" field now which is the ID of the IQX user that created the experiment. This is set automatically when the experiment is created. It would be nice to provide a property on the `Experiment` class to access the `owner` field for the experiment.

mriedem:
### Summary

This adds an enum for the Experiment.visibility field
and allows it to be set when creating the experiment
and change it when updating the experiment. Since it can
be updated the Experiment object defaults visibility to
None so that the server can set the default (private).
I mostly copied how that works from AnalysisResult.quality.

The experiment owner is set by the server and cannot be
updated.

### Details and comments

While updating the Experiment.__repr__ method I noticed that
it was missing hub/group/project so I also added those in.

Closes #825

### DIFF
--- a/qiskit/providers/ibmq/experiment/experiment.py
+++ b/qiskit/providers/ibmq/experiment/experiment.py
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-
 # This code is part of Qiskit.
 #
 # (C) Copyright IBM 2020.
--- a/qiskit/providers/ibmq/experiment/experiment.py
+++ b/qiskit/providers/ibmq/experiment/experiment.py
@@ -157,7 +157,7 @@ class Experiment:
         """Return the experiment visibility."""
         return self._visibility
 
-    @quality.setter
+    @visibility.setter
     def visibility(self, visibility: Union[ExperimentVisibility, str]) -> None:
         """Update the experiment visibility.
 
@@ -268,10 +268,11 @@ class Experiment:
         return experiment
 
     def __repr__(self) -> str:
-        attr_str = ('uuid="{}", backend_name="{}", type="{}", hub="{}", group="{}", '
-                    'project="{}", visibility="{}", owner="{}"').format(
-            self.uuid, self.backend_name, self.type, self.hub, self.group, self.project,
-            self.visibility, self.owner)
+        attr_str = (
+            'uuid="{}", backend_name="{}", type="{}", hub="{}", group="{}", '
+            'project="{}", visibility="{}", owner="{}"').format(
+                self.uuid, self.backend_name, self.type, self.hub, self.group, self.project,
+                self.visibility, self.owner)
         for attr in ['extra', 'tags']:
             val = getattr(self, attr)
             if val is not None:
--- a/qiskit/providers/ibmq/experiment/experiment.py
+++ b/qiskit/providers/ibmq/experiment/experiment.py
@@ -16,11 +16,12 @@
 
 import logging
 from datetime import datetime
-from typing import Dict, Optional, List
+from typing import Dict, Optional, List, Union
 
 from qiskit.providers.ibmq import accountprovider  # pylint: disable=unused-import
 
 from .analysis_result import AnalysisResult
+from .constants import ExperimentVisibility
 from .utils import requires_experiment_uuid
 from ..utils.converters import str_to_utc, convert_tz
 from ..api.exceptions import RequestsApiError
@@ -45,7 +46,9 @@ class Experiment:
             analysis_results: Optional[List[AnalysisResult]] = None,
             hub: Optional[str] = None,
             group: Optional[str] = None,
-            project: Optional[str] = None
+            project: Optional[str] = None,
+            visibility: Union[ExperimentVisibility, str] = None,
+            owner: Optional[str] = None
     ):
         """Experiment constructor.
 
@@ -68,6 +71,9 @@ class Experiment:
                 group from the provider is used.
             project: The project to which this experiment belongs. If not specified the
                 project from the provider is used.
+            visibility: The visibility of the experiment which determines who can
+                see it when listing experiments.
+            owner: The user ID for the owner of the experiment.
 
         Raises:
             ExperimentError: If the provider does not offer experiment services.
@@ -79,6 +85,8 @@ class Experiment:
         self.extra = extra or {}
         self.tags = tags or []
         self.type = experiment_type
+        self.visibility = visibility  # type: ignore[assignment]
+        self._owner = owner
         self._analysis_results = analysis_results
         self._plot_names = plot_names or []
         self._retrieved_plots = False
@@ -105,6 +113,8 @@ class Experiment:
         self._start_datetime = str_to_utc(remote_data.get('start_time', None))
         self.tags = remote_data.get('tags', [])
         self.type = remote_data['type']
+        self.visibility = ExperimentVisibility(remote_data['visibility'])
+        self._owner = remote_data['owner']
         self._updated_datetime = str_to_utc(remote_data.get('updated_at', None))
         self._uuid = remote_data['uuid']
         self._plot_names = remote_data.get('plot_names', [])
@@ -141,6 +151,27 @@ class Experiment:
     def project(self) -> str:
         """Return the experiment's project."""
         return self._project
+
+    @property
+    def visibility(self) -> ExperimentVisibility:
+        """Return the experiment visibility."""
+        return self._visibility
+
+    @quality.setter
+    def visibility(self, visibility: Union[ExperimentVisibility, str]) -> None:
+        """Update the experiment visibility.
+
+        Args:
+            visibility: Experiment visibility.
+        """
+        if isinstance(visibility, str):
+            visibility = ExperimentVisibility(visibility)
+        self._visibility = visibility
+
+    @property
+    def owner(self) -> str
+        """Return the experiment's owner."""
+        return self._owner
 
     @property
     def start_datetime(self) -> datetime:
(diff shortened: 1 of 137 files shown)


mriedem:
Needs tests:

- assert owner is set after an experiment is created and when listing/showing an experiment
- assert that if not specified the experiment visibility has a default set after it's created
- assert that an experiment can be created with a specific visibility and that visibility can be updated

mriedem:
Needs #833 to pass CI.

test/ibmq/test_experiment.py
```diff
@@ -220,6 +220,8 @@ def test_upload_experiment(self):
         self.assertEqual(credentials.project, new_exp.project)  # pylint: disable=no-member
         self.assertTrue(new_exp.uuid)
         self.assertTrue(new_exp.creation_datetime)
+        self.assertEqual(ExperimentVisibility.PRIVATE, new_exp.visibility)
+        self.assertIsNotNone(new_exp.owner, 'Owner should be set')
```
mriedem, line 224: I don't know why this is failing eslint:

https://github.com/Qiskit/qiskit-ibmq-provider/pull/829/checks?check_run_id=1513158086#step:5:10

>test/ibmq/test_experiment.py:224:29: E1101: Instance of 'Experiment' has no 'owner' member (no-member)

If this were a problem wouldn't the assertion fail the tests?

mriedem: I guess whatever causes this is also a problem for hub/group/project above, maybe a bug in pylint? Anyway I'll just disable the check for this line like those others.

qiskit/providers/ibmq/experiment/experiment.py
```diff
@@ -45,7 +46,9 @@ def __init__(
             analysis_results: Optional[List[AnalysisResult]] = None,
             hub: Optional[str] = None,
             group: Optional[str] = None,
-            project: Optional[str] = None
+            project: Optional[str] = None,
+            visibility: Union[ExperimentVisibility, str] = None,
```
jyu00, line 48: I'd prefer naming this `share_level` to match `IBMQJob`. And this should probably default to `private` to make it more obvious to the user what the default is.

qiskit/providers/ibmq/experiment/experiment.py
```diff
@@ -68,6 +71,9 @@ def __init__(
                 group from the provider is used.
             project: The project to which this experiment belongs. If not specified the
                 project from the provider is used.
+            visibility: The visibility of the experiment which determines who can
+                see it when listing experiments.
```
jyu00, line 73: The current model is that a user can instantiate an `Experiment` class, meaning the docstring needs to be more descriptive (like what the valid values are).

qiskit/providers/ibmq/experiment/experiment.py
```diff
@@ -66,6 +69,9 @@ def __init__(
...
+            owner: The user ID for the owner of the experiment.
```
jyu00, line 74: Same as above. Perhaps a few words that explains this is not expected to be set by the user when creating a new experiment

jyu00 (changes requested):
A few minor comments. Can you also add a release note?

qiskit/providers/ibmq/experiment/experiment.py
mriedem: >I'd prefer naming this share_level to match IBMQJob.

Yeah I can do that.

>And this should probably default to private to make it more obvious to the user what the default is.

I started with that based on how the analysis results quality field has a default, but then I thought about how the update function works and it looks like that would always pass the visibility value (or quality in the analysis result case) on update even if we don't want to change the value. I'm not sure if that could really happen, i.e. would someone construct an Experiment object directly with the fields set to update (uuid to identify the object and let's say end_time to update) and then if visibility defaults to `private` in the object but I didn't want to update the visibility to `private` (let's say the experiment visibility is `public`), they'd get the implicit update without thinking about it. If that's not really a scenario that should happen then I can set a default, IOW, would anyone only ever update an Experiment after first retrieving it and updating the fields from remote data?

qiskit/providers/ibmq/experiment/experiment.py
mriedem: OK, I was following the minimal description for like AnalysisResult.quality.

This ties into my question above about the default, i.e. could someone mistakenly update the visibility because of the default on a newly constructed object without intentionally trying to do that just because they didn't got through the retrieve route first?

mriedem (commented):
Couple of replies about how these objects can be updated and whether or not setting a default on visibility in the Experiment object could make someone unintentionally update the visibility.

As for the release note, I'll try (I'll need to read the docs on how the project is setup with `reno` but I've used that package before at least so it shouldn't be too hard).

qiskit/providers/ibmq/experiment/experiment.py
jyu00: The scenario you're describing is that a person would manually construct an `Experiment` object with an existing experiment uuid and default the visibility to private. They might then inadvertently overwrite the existing visibility value when saving the experiment in results db. I don't think it's a likely scenario but still possible. I suppose you can default it to `None` and describe in the docstring that visibility defaults to private for new experiments, but if they're creating an `Experiment` to represent an existing one (they really shouldn't), then the visibility field is None as it's unknown.

A better solution would probably be removing `experiment_uuid` from `Experiment` constructor so users cannot manually construct an existing experiment (i.e. they can only retrieve it). But that's more outside the scope of this PR.

qiskit/providers/ibmq/experiment/experiment.py
jyu00: `AnalysisResult.quality` values were not listed because they were mainly for the internal experimentalists and probably won't be used by general Qiskit users (values like "human bad" just sounds silly). I'm currently working on a new class that will have all the fields specifically for these experimentalists (like `fit`).

jyu00:
About release note, you can find more details about reno here: https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/CONTRIBUTING.md#adding-a-new-release-note
But basically you just do `reno new <description>`, which will create a new yaml file. You can then update the file with the actual release note, add it to git, and push the changes.

qiskit/providers/ibmq/experiment/experiment.py
mriedem: > I suppose you can default it to `None` and describe in the docstring that visibility defaults to private for new experiments, but if they're creating an `Experiment` to represent an existing one (they really shouldn't), then the visibility field is None as it's unknown.

This was the course I was thinking of taking.

qiskit/providers/ibmq/experiment/experiment.py
```diff
@@ -43,7 +44,9 @@ def __init__(
             analysis_results: Optional[List[AnalysisResult]] = None,
             hub: Optional[str] = None,
             group: Optional[str] = None,
-            project: Optional[str] = None
+            project: Optional[str] = None,
+            share_level: Union[ExperimentShareLevel, str] = None,
```
jyu00, line 48:
```suggestion
            share_level: Optional[Union[ExperimentShareLevel, str]] = None,
```
I'm surprised mypy didn't complain - if it can be `None` the type needs to have `Optional`.

qiskit/providers/ibmq/experiment/experiment.py
```diff
@@ -66,6 +69,18 @@ def __init__(
                 group from the provider is used.
             project: The project to which this experiment belongs. If not specified the
                 project from the provider is used.
+            share_level: The level at which the experiment is shared which determines who can
+                see it when listing experiments. This defaults to "private" for new
```
jyu00, line 73:
```suggestion
            share_level: The level at which the experiment is shared. This determines who can
                access the experiment, including changing its data. This defaults to "private" for new
```

qiskit/providers/ibmq/experiment/experiment.py
```diff
@@ -140,6 +159,27 @@ def project(self) -> str:
...
+
+    @share_level.setter
+    def share_level(self, share_level: Union[ExperimentShareLevel, str]) -> None:
+        """Update the experiment share_level.
+
+        Args:
+            share_level: Experiment share_level.
+        """
+        if isinstance(share_level, str):
+            share_level = ExperimentShareLevel(share_level)
```
jyu00, line 175:
```suggestion
            share_level = ExperimentShareLevel(share_level.lower())
```

qiskit/providers/ibmq/experiment/experiment.py
```diff
@@ -140,6 +159,27 @@ def project(self) -> str:
         """Return the experiment's project."""
         return self._project
 
+    @property
+    def share_level(self) -> ExperimentShareLevel:
```
jyu00, line 163:
```suggestion
    def share_level(self) -> Optional[ExperimentShareLevel]:
```
Same as before, needs `Optional` if it can be `None`.

jyu00 (commented):
Few more minor comments.

qiskit/providers/ibmq/experiment/experiment.py
jyu00: One downside of defaulting to `None` is that if I create a new `Experiment` and do `experiment.share_level`, it'll give me `None`, and I'd just have to know that means private. There are ways around it but probably not worth the effort as not allowing constructing an `Experiment` with a uuid is still the better option.

qiskit/providers/ibmq/experiment/experiment.py
```diff
@@ -66,6 +69,18 @@ def __init__(
                 group from the provider is used.
             project: The project to which this experiment belongs. If not specified the
                 project from the provider is used.
+            share_level: The level at which the experiment is shared. This determines who can
+                access the experiment, including changing its data. This defaults to "private" for new
+                experiments. Possible values inclue:
```
jyu00, line 74:
```suggestion
            share_level: The level at which the experiment is shared. This determines who can
                access the experiment, including changing its data. This defaults to "private" 
                for new experiments. Possible values inclue:
```

jyu00 (approved):

qiskit/providers/ibmq/experiment/experiment.py
```diff
@@ -66,6 +69,18 @@ def __init__(
                 group from the provider is used.
             project: The project to which this experiment belongs. If not specified the
                 project from the provider is used.
+            share_level: The level at which the experiment is shared. This determines who can
+                access the experiment, including changing its data. This defaults to "private" 
+                for new experiments. Possible values inclue:
```
jyu00, line 74:
```suggestion
                access the experiment, including changing its data. This defaults to "private"
                for new experiments. Possible values include:
```

PR #834, opened 2020-12-12
Fix tests for python 3.9

nileshpatra:
✅ I have added the tests to cover my changes.
✅ I have updated the documentation accordingly.
✅ I have read the CONTRIBUTING document.

Tests currently seem to fail due to asyncio API changes in python 3.9, and deprecated in earlier versions, as can be seen [here](https://docs.python.org/3/whatsnew/3.7.html#asyncio)

This is an attempt to fix the same

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,7 @@ setup(
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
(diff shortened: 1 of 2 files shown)


nileshpatra:
Ah, looks like this will need qiskit with [this](https://github.com/Qiskit/qiskit-terra/pull/5189) applied before we can test here.

nileshpatra:
Looks like it needs more work, closing for now. Sorry for the noise!

Issue #854, opened 2021-01-05
Add support for new experiments API changes


jyu00:
### What is the expected enhancement?

Support the new experiments API changes:

- ~[ ] make `device_components` optional for analysis results~
- ~[ ] make `fit` optional for analysis results~
- [x] support updating experiment `extra` and `tags`
- [x] support device backend name
- [x] support excluder_owner and owner_only when filtering experiments
- [x] typo in `public_only` docstring
- [x] new `verified` boolean
- [x] new free-form `notes` in experiments

mriedem:
It's now also possible to filter experiments in `GET /experiments` by `owner` using the special `me` and `!me` values:

- `GET /experiments?owner=me`: only return experiments owned by the user making the request
- `GET /experiments?owner=!me`: only return experiments available to but not owned by the user making the request

jyu00:
@mriedem FYI since these new changes are to support Ignis, I'm going to wait until the Ignis interface (Qiskit/qiskit-terra#5499) is finalized before working on this, since there will be a bit of interface change.

mriedem:
OK, no problem. Would you want to split out the `owner=me` and `owner=!me` support stuff into a separate issue to be worked on outside of the Ignis changes, or is just everything held up on that interface change?

jyu00:
Sure. `owner=me` can be done separately.

jyu00:
Closed by #945

PR #858, opened 2021-01-07
Support filtering experiments by public share_level

### LINKED ISSUE #848: Add support for filtering experiments by hub/group/project and visibility
#### Hub/Group/Project

The `GET /experiments` API allows filtering experiments by a specific `hub_id`, `group_id` and `project_id`. The following combinations are allowed:

- `hub_id` only
- `hub_id` and `group_id`
- `hub_id` and `group_id` and `project_id`

If not specified, the service returns experiments that:

- The user owns
- That are public
- That are shared at the hub, group or project level and are in a provider that the user is in

Passing specific `hub_id`, `group_id` and `project_id` filter values narrows that query on the server side and should retrieve resources faster.

#### Visibility

Currently the `GET /experiments` API supports filtering by `visibility=public`. Since we probably don't want to duplicate the valid `visibility` filter values on the client side in Qiskit, I would suggest adding an `is_public` kwarg to the client code which defaults to False so it's not included in the `GET /experiments` call but if True would include the filter like `GET /experiments?visibility=public`.

mriedem:
### Summary

This adds `exclude_public` and `public_only` filters when listing experiments.
By default the behavior is the same as today, there is no `visibility`
filter applied. If `exclude_public=True` then only non-public experiments
are returned. If `public_only=True` then only public experiments are returned.
If both are True an error is raised.

### Details and comments

This also fixes a typo in a previous release note.

Closes #848


mriedem:
@jyu00 if this looks OK I'll proceed with adding some tests.

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -141,6 +142,10 @@ def experiments(
             group: Filter by hub and group. `hub` must also be specified if `group` is.
             project: Filter by hub, group, and project. `hub` and `group` must also be
                 specified if `project` is.
+            is_public: Filter by public visibility. If True only public experiments
+                will be returned. If False only experiments that are not public will
+                be returned. If None (the default), then a visibility filter is not
+                applied.
```
jyu00, line 148: I don't know if using a boolean keyword to represent 3 things is the best approach. Without reading the docs (because who reads docs), a user is likely to think `is_public=True` implies "all experiments visible to me, including public ones". 

Perhaps 2 keywords would make it clearer - `exclude_public` and `public_only`. These are, of course, mutually exclusive so we'll need to check for that. 

In addition, I'd avoid using the word `visibility`, since `share_level` is used previously (to match ibmq jobs), and it's unnecessary to introduce yet another terminology.

jyu00: One more thing - I was going to change the experiment `share_level` to match that of ibmqjob at some point - i.e. `none` instead of `private`, and `global` instead of `public`. I actually like `public` more than `global`, but I like consistency above all :)

mriedem: > Perhaps 2 keywords would make it clearer - `exclude_public` and `public_only`. These are, of course, mutually exclusive so we'll need to check for that.

I [wondered about this myself in the issue](https://github.com/Qiskit/qiskit-ibmq-provider/issues/848#issuecomment-756306601) and while I don't love it, it's fine as long as we validate against using both to create a conflict:

| exclude_public | public_only | result |
| ---------------- | ------------ | ------- |
| False                | False           | no visibility filter (default) |
| True                 | False           | visibility=!public |
| False                | True            | visibility=public |
| True                 | True            | raise error |

Right?

> 
> In addition, I'd avoid using the word `visibility`, since `share_level` is used previously (to match ibmq jobs), and it's unnecessary to introduce yet another terminology.

Ack.

test/ibmq/test_experiment.py
```diff
@@ -604,15 +645,17 @@ def test_analysis_results_limit(self):
...
     ) -> Experiment:
         backend_name = backend_name or self.experiments[0].backend_name
         new_exp = Experiment(
             provider=self.provider,
             backend_name=backend_name,
             experiment_type='test',
             tags=['qiskit-test'],
-            start_datetime=start_dt
+            start_datetime=start_dt,
+            share_level=share_level
```
mriedem, line 658: I can't figure out why this is failing:

https://github.com/Qiskit/qiskit-ibmq-provider/pull/858/checks?check_run_id=1664935530

test/ibmq/test_experiment.py:652:18: E1123: Unexpected keyword argument 'share_level' in constructor call (unexpected-keyword-arg)

It doesn't fail for me locally.

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -141,6 +143,14 @@ def experiments(
...
             project: Filter by hub, group, and project. `hub` and `group` must also be
                 specified if `project` is.
+            exclude_public: If False (the default), experiments with a `public`
+                `share_level` may be returned. If True, experiments with a `public`
+                `share_level` will not be returned. Cannot be True if `public_only`
+                is True.
+            public_only: If False (the default), experiments with a `public`
+                `share_level` may be returned. If True, only experiments with a
+                `public` `share_level` will be returned. Cannot be True if
+                `exclude_public` is True.
```
jyu00, line 153: some nit pick on docstring: 
1. We use double backticks for code text as they're rendered as monospaced text. Single backtick is for parameters of this method (rendered as italic).
2. `True` and `False` also need to be quoted with double backticks, if they refer to boolean `True` and `False` (instead of the English words)
3. I'd do a short explanation on what "a public share_level" means, so something like
```
            exclude_public: If ``False`` (the default), experiments with ``share_level=public``
                (that is, experiments visible to all users)  may be returned. If ``True``, public experiments are excluded. 
                Cannot be ``True`` if `public_only` is ``True``.
```

test/ibmq/test_experiment.py
```diff
@@ -235,6 +235,47 @@ def test_experiments_with_hgp_error(self):
                 for key in missing_keys:
                     self.assertIn(key, str(ex_cm.exception))
 
+    def test_experiments_with_public_filters(self):
```
jyu00, line 238: Since we're really testing 2 separate keywords (`exclude_public` and `public_only`), it's probably better to split the test case into 2, or use `subTest`.

jyu00 (approved):
A couple minor comments but otherwise looks good

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -143,14 +143,12 @@ def experiments(
...
-            public_only: If False (the default), experiments with a `public`
-                `share_level` may be returned. If True, only experiments with a
-                `public` `share_level` will be returned. Cannot be True if
-                `exclude_public` is True.
+            exclude_public: If ``True``, experiments with ``share_level=public``
+                (that is, experiments visible to all users) will not be returned.
+                Cannot be ``True`` if `public_only` is ``True``.
+            public_only: If ``True``, only experiments with ``share_level=public``
+                (that is, experiments visible to all users) will be returned.
+                Cannot be ``True`` if `public_only` is ``True``.
```
mriedem, line 151: Copy/paste typo here, should be `exclude_public`.

jyu00: Oops. Will fix as part of #854.

Issue #878, opened 2021-02-08
ParameterExpression can be complex


eggerdj:
### Information

- **Qiskit IBMQ Provider version**: 0.11.1
- **Python version**: 3.8.5
- **Operating system**: Ubuntu

### What is the current behavior?
`backend.run(assemble(sched, backend))` fails when the schedule was obtained by assigning a complex amplitude to a parameter. This should not be the case.

### Steps to reproduce the problem
Run the following self-contained code
```
from qiskit import assemble
import qiskit.pulse as pulse
from qiskit.circuit import Parameter
from qiskit.pulse import Gaussian, DriveChannel

from qiskit import IBMQ
IBMQ.load_account()

provider = IBMQ.get_provider(hub=..., group=..., project=...)
backend = provider.get_backend('ibmq_montreal')

test_p1 = Parameter('test')
test_p3 = Parameter('test')

with pulse.build() as test1:
    pulse.play(Gaussian(160, amp=test_p1, sigma=40), DriveChannel(0))
    
with pulse.build() as test2:
    pulse.play(Gaussian(160, amp=0.2+0.01j, sigma=40), DriveChannel(0))
    
with pulse.build() as test3:
    pulse.play(Gaussian(160, amp=test_p3, sigma=40), DriveChannel(0))
    
test1.assign_parameters({test_p1: 0.2+0.01j})
test3.assign_parameters({test_p3: 0.2})

job = backend.run(assemble(test2, backend))  # This will run
job = backend.run(assemble(test3, backend))  # This will run
job = backend.run(assemble(test1, backend))  # This will fail
```

### What is the expected behavior?
The code above should run.
We should be able to run schedule with parameterized complex amplitudes.

### Suggested solutions
The trace log takes us back to the following file: https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/utils/json_encoder.py Here the following code is executed (lines 33-34):
```
        if isinstance(o, ParameterExpression):
            return float(o)
```
This could be modified to accept complex as follows:
```
        if isinstance(o, ParameterExpression):
            try:
                return float(o)
            except TypeError:
                return complex(o)
```

PR #879, opened 2021-02-08
Fix issue #878

eggerdj:
### Summary
Fixes issue #878 

### Details and comments
A small amendment to the `IQXJsonEncoder` should fix the issue.


mtreinish (approved):
This LGTM, although it does make me thing we probably need to add a method to `ParameterExpression` to check the type of a bound value up front.

Issue #902, opened 2021-03-10
Human/Computer ResultQuality deprecation


mriedem:
### What is the expected behavior?

The resultsdb API has deprecated the Human/Computer result quality values defined here:

https://github.com/Qiskit/qiskit-ibmq-provider/blob/0.12.1/qiskit/providers/ibmq/experiment/constants.py#L49-L53

There are two new quality enum values: `Bad` and `Good`.

There is also now a new `verified` boolean field for analysis results which identifies a result whose quality has been verified by a human.

So the mapping table is now:

| Old Quality   | New Quality | Verified |
| ------------- | ----------- | -------- |
| Human Bad     | Bad         | true     |
| Computer Bad  | Bad         | false    |
| Computer Good | Good        | false    |
| Human Good    | Good        | true     |

The `POST /analysis_results`, `PUT /analysis_results` and `GET /analysis_results` APIs will continue to honor the old quality values but will map them to the new values, e.g. if filtering results like `GET /analysis_results?quality=Computer Bad` that would internally be mapped to `GET /analysis_results?quality=Bad`. To have finer-grained control on the Human/Computer part of that you'd need to start passing the `verified` filter as well, e.g. `GET /analysis_results?quality=Bad&verified=false`.

mriedem:
@jyu00 I was thinking about having this dealt with in two parts, similar to how I did this in the service:

1. Add support to Qiskit for the `verified` field
2. Deprecate the Human/Computer `quality` values

Since deprecating the latter depends on using the former, and `verified` can be added pretty easily on its own, I figured it's easier to approach this in two phases. I'm not totally clear how Qiskit deprecates things either so I feel like I could implement (1) quickly but I'm not so sure about the direction on how (2) happens or gets signaled. One thought is emitting a DeprecationWarning if using a deprecated quality value when creating or updating an analysis result but then mapping the deprecated value to the new `quality` value and matching `verified` value, e.g. if updating a result to `Human Good` we'd emit a deprecation warning and actually update the result in the server with `quality='Good'` and `verified=True`.

mriedem:
> We do still need to provide backward compatibility to data retrieve from the server that has the old format (unless you already converted them all). For example, `Human Good` will be mapped to `quality=Good, verify=True`.

Yup this is already done in the database.

PR #909, opened 2021-03-31
Adds support for the verified field on analysis results.

markSbrandt:
Part of #902
 
### Summary
Adds support for the `verified` field on analysis results that indicate if a result's quality has been verified by a human.

### DIFF
--- a/qiskit/providers/ibmq/api/rest/root.py
+++ b/qiskit/providers/ibmq/api/rest/root.py
@@ -247,7 +247,8 @@ class Api(RestAdapterBase):
             device_components: Optional[List[str]] = None,
             experiment_uuid: Optional[str] = None,
             result_type: Optional[str] = None,
-            quality: Optional[List[str]] = None
+            quality: Optional[List[str]] = None,
+            verified: Optional[bool] = None
     ) -> Dict:
         """Return all analysis results.
 
@@ -259,6 +260,7 @@ class Api(RestAdapterBase):
             experiment_uuid: Experiment UUID used for filtering.
             result_type: Analysis result type used for filtering.
             quality: Quality value used for filtering.
+            verified: Indicates whether this result has been verified..
 
         Returns:
             JSON response.
@@ -279,6 +281,8 @@ class Api(RestAdapterBase):
             params['limit'] = limit
         if marker:
             params['marker'] = marker
+        if verified is not None:
+            params['verified'] = verified
         return self.session.get(url, params=params).json()
 
     def analysis_result_upload(self, result: Dict) -> Dict:
--- a/qiskit/providers/ibmq/experiment/analysis_result.py
+++ b/qiskit/providers/ibmq/experiment/analysis_result.py
@@ -58,6 +58,7 @@ class AnalysisResult:
             tags: Optional[List[str]] = None,
             result_uuid: Optional[str] = None,
             backend_name: Optional[str] = None,
+            verified: Optional[bool] = False
     ):
         """AnalysisResult constructor.
 
@@ -72,6 +73,7 @@ class AnalysisResult:
             tags: Tags for this result.
             result_uuid: Unique identifier for the result.
             backend_name: Name of the backend on which the experiment was run.
+            verified: Indicates whether this result has been verified..
 
         Raises:
             IBMQInputValueError: If an input argument is invalid.
@@ -88,6 +90,7 @@ class AnalysisResult:
         self._uuid = result_uuid
         self.device_components = device_components
         self.backend_name = backend_name
+        self.verified = verified
         self._creation_datetime = None
         self._updated_datetime = None
 
@@ -105,6 +108,7 @@ class AnalysisResult:
         self.quality = ResultQuality(remote_data['quality'])
         self.tags = remote_data['tags'] or []
         self.type = remote_data['type']
+        self.verified = remote_data['verified']
         self._creation_datetime = str_to_utc(remote_data['created_at'])
         self._updated_datetime = str_to_utc(remote_data['updated_at'])
         self._uuid = remote_data['uuid']
@@ -174,7 +178,8 @@ class AnalysisResult:
                   tags=remote_data['tags'],
                   result_uuid=remote_data['uuid'],
                   device_components=remote_data['device_components'],
-                  backend_name=remote_data['device_name']
+                  backend_name=remote_data['device_name'],
+                  verified=remote_data['verified']
                   )
         obj._creation_datetime = str_to_utc(remote_data['created_at'])
         obj._updated_datetime = str_to_utc(remote_data['updated_at'])
@@ -183,7 +188,7 @@ class AnalysisResult:
     def __repr__(self) -> str:
         attr_str = 'uuid="{}"'.format(self.uuid)
         for attr in ['type', 'quality', 'experiment_uuid', 'backend_name',
-                     'chisq', 'tags', 'device_components']:
+                     'chisq', 'tags', 'device_components', 'verified']:
             val = getattr(self, attr)
             if val is not None:
                 if isinstance(val, str):
--- a/qiskit/providers/ibmq/experiment/experimentservice.py
+++ b/qiskit/providers/ibmq/experiment/experimentservice.py
@@ -312,7 +312,8 @@ class ExperimentService:
             device_components: Optional[List[str]] = None,
             experiment_id: Optional[str] = None,
             result_type: Optional[str] = None,
-            quality: Optional[List[Tuple[str, Union[str, ResultQuality]]]] = None
+            quality: Optional[List[Tuple[str, Union[str, ResultQuality]]]] = None,
+            verified: Optional[bool] = None
     ) -> List[AnalysisResult]:
         """Retrieve all analysis results, with optional filtering.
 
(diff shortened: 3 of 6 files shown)


qiskit/providers/ibmq/api/rest/root.py
```diff
@@ -279,6 +281,8 @@ def analysis_results(
             params['limit'] = limit
         if marker:
             params['marker'] = marker
+        if verified:
```
mriedem, line 284: `verified` is a boolean so if you wanted to filter on `verified=False` this condition would not work.
```suggestion
        if verified is not None:
```

qiskit/providers/ibmq/experiment/analysis_result.py
```diff
@@ -88,6 +90,7 @@ def __init__(
         self._uuid = result_uuid
         self.device_components = device_components
         self.backend_name = backend_name
+        self.verified = verified
```
mriedem, line 93: Should add this to the list of attributes in `__repr__` below.

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -383,6 +386,8 @@ def upload_analysis_result(self, result: AnalysisResult) -> None:
             data['tags'] = result.tags
         if result.uuid:
             data['uuid'] = result.uuid
+        if result.verified:
```
mriedem, line 389:
```suggestion
        if result.verified is not None:
```

Same thing as above, booleans are tricky.

releasenotes/add-verified-field-to-analysis-results-5926fa1428ef1df5.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Introduced a new property (verified) on Analysis results. This property
+    identifies if the result quality was verified by a human.
```
mriedem, line 5:
```suggestion
    :class:`qiskit.providers.ibmq.analysis_result.AnalysisResult` now has an additional
    `verified` attribute which identifies if the `quality` has been verified by a human.

```

test/ibmq/test_experiment.py
```diff
@@ -406,6 +406,7 @@ def test_get_analysis_results(self):
         results = self.provider.experiment.analysis_results()
         for res in results:
             self.assertTrue(isinstance(res, AnalysisResult))
+            self.assertEqual(res.verified, None)
```
mriedem, line 409:
```suggestion
            self.assertFalse(res.verified)
```

I would expect this to be `False` by default, maybe it's None because of some of the boolean conditional logic issues?

mriedem (changes requested):

test/ibmq/test_experiment.py
```diff
@@ -422,6 +423,7 @@ def test_upload_analysis_result(self):
                                     result_type=ref_result.type,
                                     device_components=[device_comp.type],
                                     quality='No Information',
+                                    verified=False,
```
mriedem, line 426: linter is complaining about this:

https://github.com/Qiskit/qiskit-ibmq-provider/pull/909/checks?check_run_id=2240623402#step:5:10

I'm not sure how valid it is, I've had to bypass some lint failures in the test code before, maybe try to recreate locally and see if you can make it happy (I create a virtual environment and then use the make file to run things like pylint, mypy, etc since tox isn't really used for those).

releasenotes/add-verified-field-to-analysis-results-5926fa1428ef1df5.yaml
jyu00: ^ we use reStructuredText for doc rendering, which uses 2 backticks for inline literals and single for italics (I assume you want `verified` and not _verified_).

mriedem: Oops, yeah, keep forgetting that (I used to do everything in rst for years and now I'm too used to markdown). Updated the suggestion.

test/ibmq/test_experiment.py
markSbrandt: wasn't sure how optional parameters work here, I assumed an optional when not passed would be None

qiskit/providers/ibmq/experiment/analysis_result.py
```diff
@@ -58,6 +58,7 @@ def __init__(
             tags: Optional[List[str]] = None,
             result_uuid: Optional[str] = None,
             backend_name: Optional[str] = None,
+            verified: Optional[bool] = None
```
mriedem, line 61: I don't know if we want to default this to False here or not, the server side will default it based on the `quality` value passed in, so might as well just leave it default to None in the client so it's not included in the request body.

releasenotes/add-verified-field-to-analysis-results-5926fa1428ef1df5.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    :class:``qiskit.providers.ibmq.analysis_result.AnalysisResult` now has an additional
+    ``verified`` attribute which identifies if the ``quality`` has been verified by a human.
```
mriedem, line 5:
```suggestion
    ``verified`` attribute which identifies if the ``quality`` has been verified by a human.

```

nit: I don't know if you really need the new line at the end of the file or not, but GitHub is highlighting it.

test/ibmq/test_experiment.py
mriedem: But this is listing analysis results, right? So I think we're calling this to list the results:

https://github.com/Qiskit/qiskit-ibmq-provider/blob/2d26bcbe864d2b2038d96885e82cc4cdf5d265ab/qiskit/providers/ibmq/experiment/experimentservice.py#L308

which constructs the AnalysisResult objects here:

https://github.com/Qiskit/qiskit-ibmq-provider/blob/2d26bcbe864d2b2038d96885e82cc4cdf5d265ab/qiskit/providers/ibmq/experiment/experimentservice.py#L362

Which should be passing the `verified` kwarg here:

https://github.com/Qiskit/qiskit-ibmq-provider/blob/2d26bcbe864d2b2038d96885e82cc4cdf5d265ab/qiskit/providers/ibmq/experiment/analysis_result.py#L182

which should set this:

https://github.com/Qiskit/qiskit-ibmq-provider/blob/2d26bcbe864d2b2038d96885e82cc4cdf5d265ab/qiskit/providers/ibmq/experiment/analysis_result.py#L93

And that should be coming back as `false` by default from the REST API (the default value in the DB is `false` so we should never return `null` back in the JSON).

mriedem (commented):
Something is still weird about the `self.assertEqual(res.verified, None)` on that one test, maybe I'm missing something? @jyu00 do you see something I'm missing?

test/ibmq/test_experiment.py
mriedem: I wonder if the `test/ibmq/test_experiment.py:421:21: E1123: Unexpected keyword argument 'verified' in constructor call (unexpected-keyword-arg)` issue is related to this somehow? Is the kwarg not getting set properly?

test/ibmq/test_experiment.py
```diff
@@ -406,6 +406,7 @@ def test_get_analysis_results(self):
         results = self.provider.experiment.analysis_results()
         for res in results:
             self.assertTrue(isinstance(res, AnalysisResult))
+            self.assertEqual(res.verified, False)
```
mriedem, line 409:
```suggestion
            self.assertFalse(res.verified)
```

Just a nit.

mriedem (approved):
Looks OK to me now.

What happened with the pylint failure you were getting?  Did you sort it out?

test/ibmq/test_experiment.py
markSbrandt: Nope, but that test was testing the default value so I just removed it from the call to the constructor and let it's do it's thing internally.

markSbrandt:
> Looks OK to me now.
> 
> What happened with the pylint failure you were getting? Did you sort it out?

Nope, but that test was testing the default value so I just removed it from the call to the constructor and let it's do it's thing internally.

qiskit/providers/ibmq/experiment/analysis_result.py
```diff
@@ -72,6 +73,7 @@ def __init__(
             tags: Tags for this result.
             result_uuid: Unique identifier for the result.
             backend_name: Name of the backend on which the experiment was run.
+            verified: Identifies a result whose quality has been verified by a human.
```
jyu00, line 76:
```suggestion
            verified: Indicates whether this result has been verified.
```
Nit - making this more generic since there are no rules enforcing it being verified by a human.

test/ibmq/test_experiment.py
```diff
@@ -458,6 +460,7 @@ def test_update_analysis_result(self):
                 self.assertEqual(res.type, original_type)
                 self.assertEqual(res.quality, ResultQuality.HUMAN_BAD)
                 self.assertTrue(res.updated_datetime.tzinfo)
+                self.assertTrue(res.verified)
```
jyu00, line 463: Given we want to test whether `ibmq-provider` is sending the right data to the server, not whether the server does the correct mapping, I'd suggest adding `new_result.verified = True` after line 450 instead of relying on `HUMAN_BAD` mapping.

qiskit/providers/ibmq/experiment/analysis_result.py
markSbrandt: updated the various places this description showed up

jyu00 (approved):
LGTM. Thanks for adding this!

PR #915, opened 2021-04-14
Add Good/Bad ResultQuality enum values

mriedem:
### Summary

The Computer/Human analysis result `quality` values are
being deprecated and replaced with Good/Bad `quality` values.
The `verified` field indicates if a human verified the quality,
so `Human Bad` maps to `quality='Bad'` and `verified=True` while
`Computer Good` maps to `quality='Good'` and `verified=False`.
The server handles the mapping of the old deprecated `quality`
values when an analysis result is created or updated.

Existing data has already been migrated so when getting analysis
results back they should have `quality` of `Bad`, `No Information`
or `Good` which means we need to add the new `Bad` and `Good`
values to the `ResultQuality` enum.

### Details and comments

Part of #902

### DIFF
--- a/qiskit/providers/ibmq/experiment/constants.py
+++ b/qiskit/providers/ibmq/experiment/constants.py
@@ -48,9 +48,11 @@ class ResultQuality(enum.Enum):
 
     HUMAN_BAD = 'Human Bad', 1
     COMPUTER_BAD = 'Computer Bad', 2
-    NO_INFORMATION = 'No Information', 3
-    COMPUTER_GOOD = 'Computer Good', 4
-    HUMAN_GOOD = 'Human Good', 5
+    BAD = 'Bad', 3
+    NO_INFORMATION = 'No Information', 4
+    GOOD = 'Good', 5
+    COMPUTER_GOOD = 'Computer Good', 6
+    HUMAN_GOOD = 'Human Good', 7
 
 
 class ExperimentShareLevel(enum.Enum):


mriedem:
I don't know how this is working:

https://github.com/Qiskit/qiskit-ibmq-provider/blob/b1b8c64f2e3960dd77337350ba8670f3ab7913b5/test/ibmq/test_experiment.py#L462

Because that `Human Bad` quality value that's getting set on update should be mapped to `quality='Bad'` and `verified=True`. Are those tests disabled?

mriedem:
> Are those tests disabled?

Looks like yes, I see they are in the cron-other bucket now:

https://github.com/Qiskit/qiskit-ibmq-provider/runs/2339932139?check_suite_focus=true

But even then they don't get run because `USE_STAGING_CREDENTIALS` is not set.

jyu00:
Can you get rid of the old values (`HUMAN_GOOD` etc)? Since this is still beta we don't need to maintain backward compatibility. 

Also it'd be nice if you can fix the tests too (you can run them locally), but if you don't have time that's fine too. Some of these will change again with Ignis integration.

mriedem:
> Can you get rid of the old values (`HUMAN_GOOD` etc)? Since this is still beta we don't need to maintain backward compatibility.

Yeah, I wasn't sure about doing that in this PR since the tests aren't being run and I know they are likely broken right now, but figured this would get over the immediate issue of not being able to use Qiskit at all because of the missing enum values.
 
> Also it'd be nice if you can fix the tests too (you can run them locally), but if you don't have time that's fine too. Some of these will change again with Ignis integration.

If getting the tests working isn't a blocker for this PR then that makes burning out the old deprecated `quality` values easier. I'll try to get the tests to run locally but I've never done that before so I'm not sure how long that will take me (assuming with an endpoint URL, token and provider it should be pretty straight-forward).

mriedem:
> I'll try to get the tests to run locally but I've never done that before so I'm not sure how long that will take me (assuming with an endpoint URL, token and provider it should be pretty straight-forward).

Between https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/CONTRIBUTING.md#test, https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/.github/workflows/cron-other.yml#L18 and https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/test/decorators.py#L21-L24 I was able to get the tests to run in a local virtual environment, here are the results:

https://gist.github.com/mriedem/2d7b3b9a4af230e1bc1fad8ff8a550d9

jyu00 (approved):
I plan on doing a release in the next couple of days, so let's just merge this and I can fix the tests at a later time.

mriedem:
> 
> 
> I plan on doing a release in the next couple of days, so let's just merge this and I can fix the tests at a later time.

OK, I can open a separate issue and work some fixes for the tests. It looks like the main issue after this fix is there is one analysis result in the staging DB which has an empty `fit` value (that was made optional in 0.20.1 of the API to accommodate Ignis experiments) and that's what's causing the tests to blow up.

PR #919, opened 2021-04-16
Remove deprecated ResultQuality values

### LINKED ISSUE #902: Human/Computer ResultQuality deprecation

mriedem:
### Summary

The resultsdb service has deprecated the Human/Computer
analysis result `quality` values. They are replaced with
just `Good` and `Bad` values to be used with the `verified`
flag on an analysis result.

This builds on #909 and #915 to remove the deprecated
enum quality values, gets the skipped test working again,
and adds an upgrade release note.

### Details and comments

Since the experiment service in Qiskit is still considered beta
there is no deprecation period in Qiskit itself though the service/API
continues to honor and translate the old quality values for now.

Closes #902

### DIFF
--- a/qiskit/providers/ibmq/experiment/constants.py
+++ b/qiskit/providers/ibmq/experiment/constants.py
@@ -47,7 +47,7 @@ class ResultQuality(enum.Enum):
         return NotImplemented
 
     BAD = 'Bad', 1
-    NO_INFORMATION = 'No Information', 2
+    UNKNOWN = 'No Information', 2
     GOOD = 'Good', 3
 
 
--- a/qiskit/providers/ibmq/experiment/constants.py
+++ b/qiskit/providers/ibmq/experiment/constants.py
@@ -46,13 +46,9 @@ class ResultQuality(enum.Enum):
             return self.ranking < other.ranking  # type: ignore[attr-defined]
         return NotImplemented
 
-    HUMAN_BAD = 'Human Bad', 1
-    COMPUTER_BAD = 'Computer Bad', 2
-    BAD = 'Bad', 3
-    NO_INFORMATION = 'No Information', 4
-    GOOD = 'Good', 5
-    COMPUTER_GOOD = 'Computer Good', 6
-    HUMAN_GOOD = 'Human Good', 7
+    BAD = 'Bad', 1
+    NO_INFORMATION = 'No Information', 2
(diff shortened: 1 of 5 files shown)


qiskit/providers/ibmq/experiment/constants.py
```diff
@@ -46,13 +46,9 @@ def __lt__(self, other: Any) -> bool:
...
 
-    HUMAN_BAD = 'Human Bad', 1
-    COMPUTER_BAD = 'Computer Bad', 2
-    BAD = 'Bad', 3
-    NO_INFORMATION = 'No Information', 4
-    GOOD = 'Good', 5
-    COMPUTER_GOOD = 'Computer Good', 6
-    HUMAN_GOOD = 'Human Good', 7
+    BAD = 'Bad', 1
+    NO_INFORMATION = 'No Information', 2
```
jyu00, line 50: We are changing `NO_INFORMATION` to `UNKNOWN` in Qiskit to make it a single word. This class is being moved terra so it's not super important to make the change here, but if we're going to introduce breaking changes might as well do it once.

mriedem: Ack, will update today.

jyu00 (approved):
LGTM. Thanks for making the updates!

PR #935, opened 2021-05-10
Add RunnerResult

jyu00:
### Summary
This PR adds the `RunnerResult` class.


qiskit/providers/ibmq/accountprovider.py
```diff
@@ -219,14 +219,11 @@ def run_circuits(
         Note:
             This method returns a :class:``~qiskit.provider.ibmq.runtime.RuntimeJob``.
             To get the job result, you'll need to use the
-            ``qiskit_runtime.circuit_runner.RunnerResult`` class
+            ``qiskit.providers.ibmq.RunnerResult`` class
```
nonhermitian, line 222: Why can't the decoder be automatically applied here like we do in the program submission, e.g. `result_decoder=RunnerResult`?

jyu00: I forgot you can do that at program submission lol. Fixed in 0b5c7e6

PR #950, opened 2021-06-07
Added ProviderRequestParams option for run()

Rcuz8:
### Summary
Right now, executing `provider.run` with the `inputs` parameter requires a dictionary, which isn't very user-friendly. This adds an option for `provider.run` to be run with a pre-built class (`ProviderRequestParams`) as the `inputs` to enable auto-complete.

### Details and comments

- Adds the `ProviderRequestParams` class.
- Changes the documented example to use the class, rather than a dictionary. 
- Adds test coverage


qiskit/providers/ibmq/__init__.py
```diff
@@ -95,7 +95,7 @@
 from .ibmqbackendservice import IBMQBackendService
 from .utils.utils import setup_logger
 from .runner_result import RunnerResult
-
+from .runtime import ProviderRequestParams
```
jyu00, line 98: Since this is a runtime specific class, I would just leave it in `/ibmq/runtime/__init__.py` and not here.

qiskit/providers/ibmq/runtime/__init__.py
```diff
@@ -91,6 +91,7 @@
 
     from qiskit import IBMQ, QuantumCircuit
     from qiskit.providers.ibmq import RunnerResult
+    from qiskit.providers.ibmq.runtime import ProviderRequestParams as Info
```
jyu00, line 94: Using a completely different name for `as` could be very confusing to our users, who are generally less experienced in programming.

qiskit/providers/ibmq/runtime/__init__.py
```diff
@@ -102,7 +103,7 @@
     qc.measure_all()
 
     # Execute the circuit using the "circuit-runner" program.
-    program_inputs = {'circuits': circuit, 'measurement_error_mitigation': True}
+    program_inputs = Info(circuits=circuit, measurement_error_mitigation=True)
```
jyu00, line 106: I think the doc can use some description on this new interface, specially about how one can use auto complete to see the program parameters.

qiskit/providers/ibmq/runtime/utils.py
```diff
@@ -64,3 +65,19 @@ def object_hook(self, obj: Any) -> Any:
                 decoded = base64.standard_b64decode(obj['__value__'])
                 return dill.loads(decoded)
         return obj
+
+
+class ProviderRequestParams(NamedTuple):
```
jyu00, line 70: Today one can find a runtime program by doing
```
sample_program = provider.runtime.program('sample-program')
```

The idea to this new enhancement is so that a user can also do something like 
```
inputs = sample_program.input_parameters()
```
which will return a `ProviderRequestParams` instance that has the program parameters as its attributes. This allows the user to do auto complete. Each program has its own set of input parameters so the attributes (and their docs) need to be created dynamically. 

The class also needs a more runtime-y name than `ProviderRequestParams` since this is runtime specific and the current name is too generic :).

test/ibmq/runtime/test_utils.py
```diff
@@ -0,0 +1,31 @@
...
+from qiskit import QuantumCircuit
+
+
+from ...ibmqtestcase import IBMQTestCase
+
+
+class TestProviderRequestParams(IBMQTestCase):
+    """ Tests the correctness of the ProviderRequestParams class """
+
+    def test_provider_request_params(self):
```
jyu00, line 25: This is a runtime related change so test should be added to `test_runtime`. Also test should verify that the server gets the correct input parameters.

jyu00 (changes requested):
Sorry I wasn't clear about what the enhancement is about. Please see my comments on that.

qiskit/providers/ibmq/__init__.py
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/__init__.py
Rcuz8: This is updated -- the example usages should still be revisited though. I'm not sure I understand how the pieces fit together.

qiskit/providers/ibmq/runtime/__init__.py
Rcuz8: Updated, but should still revisit these example usages.

qiskit/providers/ibmq/runtime/utils.py
Rcuz8: This change should be reflected now!

test/ibmq/runtime/test_utils.py
Rcuz8: There is a validation step now that the user can take, but will also be run when calling `IBMRuntimeService.run()`

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -255,3 +258,77 @@ class ProgramResult(NamedTuple):
...
+            try:
+                # Get the value currently stored for this parameter (user-adjusted)
+                value = self.__dict__[param_name]
+            except AttributeError:
+                if req:
+                    raise IBMQInputValueError('Param (%s) missing required value!' % param_name)
+            # Check for empty value
+            if not value or value is None:
+                if req:
+                    raise IBMQInputValueError('Param (%s) missing required value!' % param_name)
```
jyu00, line 309: You can combine the two with `getattr(self, param_name, None)`.

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -255,3 +258,77 @@ class ProgramResult(NamedTuple):
...
+
+            value_type = type(value).__name__
+            # Check program param of that name is of same type
+            if value is not None and value_type != program_param.type:
+                raise IBMQInputValueError('Parameter (%s) is incorrect type! (%s != %s)' % \
+                                          (param_name, value_type, program_param.type))
```
jyu00, line 315: Types cannot be validated because it's currently only used for documentation and there are no rules on what can/cannot be specified - e.g. one may specify `string` instead of `str`.

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -255,3 +258,77 @@ class ProgramResult(NamedTuple):
...
+
+    def __str__(self) -> str:
+        """Creates string representation of object"""
+
+        # ProgramParameter object (str)
+        def param_str(param: ProgramParameter) -> str:
+            return "------------\n<> Name        : %s\n<> Type        : %s\
+        \n<> Description : %s\n<> Required?   : %s\n" % \
+                    (param.name, param.type, param.description, param.required)
```
jyu00, line 324: Can you format the code to make it more readable? Like make an actual newline whenever you have `\n`.

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -255,3 +258,77 @@ class ProgramResult(NamedTuple):
...
+
+        # List of ProgramParameter objects (str)
+        program_params_str = ''.join([param_str(param) for
+                                      param in self.__program_params.values()])
```
jyu00, line 328: I don't think it's quite necessary to have 5 lines per parameter. How about a tabular view, something like
```
Name   |  Value |  Required  | Type   | Description  
---------------------------------------------------
foo    |  3     |  True      | int    | Some number  
bar    |  None  |  True      | string | Some string  
```

I put name/value/required first because those are things users likely care about the most

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -177,13 +180,13 @@ def description(self) -> str:
         return self._description
 
     @property
-    def parameters(self) -> List['ProgramParameter']:
+    def parameters(self) -> 'ParameterNamespace':
```
jyu00, line 183: Is this supposed to be a method or property? You have `@property` here but the example has a method. If it's a property it should be `parameter` (singular) to be consistent with the other auto-completable code in this package.

releasenotes/notes/added-runtime-params-class-a458e64bea4ec8f6.yaml
```diff
@@ -0,0 +1,4 @@
+---
+upgrade:
+  - |
+    Added option to execute ``providers.run`` with ProvidersRequestParams, rather than simply a Dict object, which for auto-complete features to be used.
```
jyu00, line 4: Any mentions of methods and classes should have documentation links. Check out the other release notes for reference.

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -255,3 +258,77 @@ class ProgramResult(NamedTuple):
...
+        params (List[ProgramParameter]): The program's input parameters
+
+    Raises
+        IBMQInputValueException on validation fail
+    """
+
+    def __init__(self, params: List[ProgramParameter]):
+        super().__init__()
+        # Allow access to the raw program parameters list
+        self.params_list = params
```
jyu00, line 277: 1) Can we give it a better name than `param_list`? I for one find `program.parameter.params_list` quite confusing. How about `program_metadata` or just `metadata`? Since that's the term we use for the "documentation" part of a program.
2) I usually try to make attributes that should not be modified by users private and readable via a property. This prevents users from accidentally overwrite it.

jyu00 (commented):
We're getting close!

qiskit/providers/ibmq/runtime/runtime_program.py
Rcuz8: @jyu00 Will programs with incorrectly-specified params just fail at runtime then? 
If there's no validation on types, this could be a good spot to catch that before it pings the API with invalid info (might reduce number of calls). 
If aliases are an issue, there could be a simple (constant) lookup table for some common aliases (integer -> int, string -> str), but the program's input types are available for the user so they should be able to use the same type letter-for-letter.

qiskit/providers/ibmq/runtime/__init__.py
```diff
@@ -102,6 +102,8 @@
     qc.measure_all()
 
     # Execute the circuit using the "circuit-runner" program.
+    program = provider.runtime.program(program_id='circuit-runner')
+    program_params = program.parameters().namespace
```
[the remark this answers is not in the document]
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
jyu00: Yes they'll just fail at runtime. A lookup table won't work because right now the types are entirely arbitrary and up to the program writers. To have a lookup table we'd first need to enforce the types can only be those in the table, and we don't have that yet.

Rcuz8: Ok, not a problem!

qiskit/providers/ibmq/runtime/runtime_program.py
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
Rcuz8: Fixed!

releasenotes/notes/added-runtime-params-class-a458e64bea4ec8f6.yaml
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/__init__.py
```diff
@@ -102,11 +102,12 @@
     qc.measure_all()
 
     # Execute the circuit using the "circuit-runner" program.
-    program_inputs = {'circuits': circuit, 'measurement_error_mitigation': True}
+    param = provider.runtime.program(program_id="circuit-runner").parameter
+    param.param1 = 123
```
jyu00, line 106: Looking at this example I feel having `.parameter` as a property is a bit awkward. When you retrieve the attribute of an object multiple times without changing the state of the object, you typically expect to get the same result each time. But here if you do 
```
program = provider.runtime.program(program_id="circuit-runner")
program.parameter.param1 = 123
program.parameter.param2 = 456
```
each `.parameter` actually returns a new `ParameterNamespace` instance. And passing `program.parameter` to `run()` will be missing `param1`. 

I do like that each call returns a new instance - makes it easier to call the same program with different parameters. So perhaps just make `parameters()` a method?

qiskit/providers/ibmq/runtime/__init__.py
```diff
@@ -102,11 +102,12 @@
...
     options = {'backend_name': backend.name()}
     job = provider.runtime.run(program_id="circuit-runner",
-                               options=options,
-                               inputs=program_inputs)
+                            options=options,
+                            inputs=params)
```
jyu00, line 110: You probably want to test the examples to make sure they still work (let me know if you don't have access). In this case `params` is not defined, and `param1` is not a valid parameter to `circuit-runner`.

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -255,3 +258,81 @@ class ProgramResult(NamedTuple):
...
+    """ An abstraction for SimpleNamespace that offers param validation.
+
+    Note: Unused/Unnecessary params does not cause error (still validates)
+
+    Args:
+        params (List[ProgramParameter]): The program's input parameters
+
+    Raises
+        IBMQInputValueException on validation fail
+    """
```
jyu00, line 273: We usually have separate docstrings for the class itself and the `__init__` method.

qiskit/providers/ibmq/runtime/__init__.py
```diff
@@ -213,6 +214,8 @@ def interim_result_callback(job_id, interim_result):
    UserMessenger
    ProgramBackend
    ResultDecoder
+   RuntimeEncoder
+   RuntimeDecoder
```
jyu00, line 218: You should add the new `ParameterNamespace` class here as well so its methods get picked up in the [API doc](https://qiskit.org/documentation/apidoc/ibmq-provider.html).

qiskit/providers/ibmq/runtime/runtime_program.py
jyu00: I don't think this is fixed, unless you forgot to push lol

jyu00 (commented):
Few more small comments

qiskit/providers/ibmq/runtime/__init__.py
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/__init__.py
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/__init__.py
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -255,3 +257,78 @@ class ProgramResult(NamedTuple):
...
+            self.__program_params[param.name] = param
+
+    @property
+    def metadata(self) -> List[ProgramParameter]:
+        """Returns the parameter metadata"""
+        return self.__metadata
+
+    def validate(self) -> None:
+        """Validates the user's usage of the program's inputs
+
```
jyu00, line 292:
```suggestion
        """Validate program input values.
        
        Note:
            This method only verifies that required parameters have values. It
            does not fail the validation if the namepsace has extraneous parameters.

```

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -255,3 +257,78 @@ class ProgramResult(NamedTuple):
...
+            value = getattr(self, param_name, None)
+            # Check there exists a program parameter of that name.
+            if value is None and program_param.required:
+                raise IBMQInputValueError('Param (%s) missing required value!' % param_name)
+
+            value_type = type(value).__name__
+            # Check program param of that name is of same type
+            if value is not None and value_type != program_param.type:
+                raise IBMQInputValueError('Parameter (%s) is incorrect type! (%s != %s)' % \
+                                          (param_name, value_type, program_param.type))
```
jyu00, line 309: I thought you removed type validation after our last conversation?

jyu00 (commented):
Some doc suggestions. Also type validation needs to be removed.

qiskit/providers/ibmq/runtime/runtime_program.py
Rcuz8: Fixed! I forgot to push this 😅

qiskit/providers/ibmq/runtime/runtime_program.py
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -255,3 +257,78 @@ class ProgramResult(NamedTuple):
...
+    """ An abstraction for SimpleNamespace that offers param validation.
+
+    Note: Unused/Unnecessary params does not cause error (still validates) """
+
+    def __init__(self, params: List[ProgramParameter]):
+        """ParameterNamespace constructor.
+
+            Args:
+                params (List[ProgramParameter]): The program's input parameters
+        """
```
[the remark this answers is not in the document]
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -255,3 +257,78 @@ class ProgramResult(NamedTuple):
     name: str
     description: str
     type: str
+
+
+class ParameterNamespace(SimpleNamespace):
+    """ An abstraction for SimpleNamespace that offers param validation.
+
+    Note: Unused/Unnecessary params does not cause error (still validates) """
```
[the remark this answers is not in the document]
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -144,11 +147,19 @@ def to_dict(self) -> Dict:
...
+            "parameters": self.parameters(),
             "return_values": self.return_values,
             "interim_results": self.interim_results
         }
 
+    def parameters(self) -> 'ParameterNamespace':
+        """Program parameter definitions.
+
+        Returns:
+            Parameter definitions for this program.
```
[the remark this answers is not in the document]
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -36,7 +39,7 @@ class RuntimeProgram:
 
         # To retrieve metadata of a single program.
         program = provider.runtime.program(program_id='circuit-runner')
-        print(f"Program {program.name} takes parameters {program.parameters}")
+        print(f"Program {program.name} takes parameters {program.parameters()}")
```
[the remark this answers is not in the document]
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/__init__.py
```diff
@@ -101,12 +103,17 @@
     qc.cx(0, 1)
     qc.measure_all()
 
-    # Execute the circuit using the "circuit-runner" program.
-    program_inputs = {'circuits': circuit, 'measurement_error_mitigation': True}
+    # Set the "circuit-runner" program parameters
+    params = program(program_id="circuit-runner").parameters()
```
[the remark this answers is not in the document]
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/__init__.py
```diff
@@ -101,12 +103,17 @@
...
+    params.circuits = qc
```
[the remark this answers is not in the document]
Rcuz8: Fixed!

releasenotes/notes/added-runtime-params-class-a458e64bea4ec8f6.yaml
```diff
@@ -0,0 +1,8 @@
+---
+upgrade:
+  - |
+    The `inputs` parameter to 
+    :meth:`qiskit.providers.ibmq.runtime.ibmq_runtime_service.run` 
+    method can now be specified as a
+    :class:`qiskit.providers.ibmq.runtime.runtime_program.ParameterNamespace`
+    to support auto-complete features.
```
jyu00, line 8:
```suggestion
features:
  - |
    The `inputs` parameter to
    :meth:`qiskit.providers.ibmq.runtime.IBMRuntimeService.run`
    method can now be specified as a
    :class:`qiskit.providers.ibmq.runtime.ParameterNamespace` instance which
    supports auto-complete features. You can use
    :meth:`qiskit.providers.ibmq.runtime.RuntimeProgram.parameters` to retrieve
    an ``ParameterNamespace`` instance.
upgrade:
  - |
    :meth:`qiskit.providers.ibmq.runtime.RuntimeProgram.parameters` is now a 
    method that returns a
    :class:`qiskit.providers.ibmq.runtime.ParameterNamespace` instance, which
    you can use to fill in runtime program parameter values and pass to
    :meth:`qiskit.providers.ibmq.runtime.IBMRuntimeService.run`.
    
```
We need both `upgrade` and `feature` because it's both an API change and new feature.

jyu00 (commented):
A couple more doc fixes otherwise looks good!

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -144,11 +147,25 @@ def to_dict(self) -> Dict:
...
+    def parameters(self) -> 'ParameterNamespace':
+        """Program parameter namespace.
+
+        You can use the returned namespace to assign parameter values and pass
+        the namespace to :meth:`qiskit.providers.ibmq.runtime.IBMRuntimeService.run`.
+        The namespace allows you to use auto-completion to find program parameters.
+
+        Note that each call to this method returns a new namespace instance and
+        does not include any modification to the previous instance.
+        Returns:
```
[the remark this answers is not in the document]
Rcuz8: Fixed!

releasenotes/notes/added-runtime-params-class-a458e64bea4ec8f6.yaml
Rcuz8: Fixed!

qiskit/providers/ibmq/runtime/runtime_program.py
```diff
@@ -144,11 +147,26 @@ def to_dict(self) -> Dict:
...
+    def parameters(self) -> 'ParameterNamespace':
+        """Program parameter namespace.
+
+        You can use the returned namespace to assign parameter values and pass
+        the namespace to :meth:`qiskit.providers.ibmq.runtime.IBMRuntimeService.run`.
+        The namespace allows you to use auto-completion to find program parameters.
+
+        Note that each call to this method returns a new namespace instance and
+        does not include any modification to the previous instance.
+
```
Rcuz8, line 164:
```suggestion

```

jyu00 (approved):
Thanks for adding this!

PR #964, opened 2021-06-15
Handle error converting complex value to float

rathishcholarajan:
### Summary
Symengine was introduced in qiskit-terra for performance improvements https://github.com/Qiskit/qiskit-terra/commit/514caef5994d8c0a4767f4ec3992eb1ef93a266d and it throws RuntimeError instead of old TypeError when we try to convert a complex ParameterExpression to float.

### Details and comments
This fixes the cron terra-master error https://github.com/Qiskit/qiskit-ibmq-provider/runs/2791018014?check_suite_focus=true


jyu00 (approved):
LGTM. Thanks for fixing this!

PR #965, opened 2021-06-16
Fix ESP bug

zachschoenfeld33:
### Summary

Fix bug when `use_measure_esp=False`.

### Details and comments

We want to set `use_measure_esp` to the value of `measure_esp_enabled`, if `use_measure_esp` is not set (ie is `None`).

Previously, this was done as 
```
use_measure_esp = use_measure_esp or measure_esp_enabled
```

This is not, correct, however as it sets `use_measure_esp` to `measure_esp_enabled` when `use_measure_esp=False`. Instead, this should be done as
```
if use_measure_esp is None:
  use_measure_esp = measure_esp_enabled
```

Tests have been added to ensure the correct behavior is used.


jyu00 (approved):
Can't believe I didn't catch this in the review :woman_facepalming:

PR #968, opened 2021-06-17
remove job share level due to low adoption

### LINKED ISSUE #914: update job share level after execution
Having a command to update the share_level property of a job in the remote database after its execution.

rathishcholarajan:
### Summary
Remove job share level due to low adoption.

### Details and comments
Fixes #914 (or rather addresses why we won't be implementing #914)

### DIFF
--- a/releasenotes/notes/remove-job-share-level-a81db898a1d3b0e8.yaml
+++ b/releasenotes/notes/remove-job-share-level-a81db898a1d3b0e8.yaml
@@ -1,0 +1,6 @@
+---
+upgrade:
+  - |
+    Job share level has been removed due to low adoption. This means you can non longer pass
+    share_level when creating a job or use :meth:`qiskit.providers.ibmq.job.IBMQJob.share_level`
+    method to get a job's share level.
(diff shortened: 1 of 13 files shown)


releasenotes/notes/remove-job-share-level-a81db898a1d3b0e8.yaml
```diff
@@ -0,0 +1,6 @@
+---
+upgrade:
+  - |
+    Job share level has been removed due to low adoption. This means you can non longer pass
+    share_level when creating a job or use :meth:`qiskit.providers.ibmq.job.IBMQJob.share_level`
```
jyu00, line 5: The `:meth:` link will fail if the method no longer exists

jyu00 (changes requested):
Qiskit normally has a deprecation period of at least 3 months (see [here](https://qiskit.org/documentation/contributing_to_qiskit.html)). But in this case we can't do that since it's no longer supported on the server side. However, we can still be nice and instead of deleting the methods / parameters outright, keep them in place but issue a warning saying they no longer work and will be removed.

rathishcholarajan:
> Qiskit normally has a deprecation period of at least 3 months (see [here](https://qiskit.org/documentation/contributing_to_qiskit.html)). But in this case we can't do that since it's no longer supported on the server side. However, we can still be nice and instead of deleting the methods / parameters outright, keep them in place but issue a warning saying they no longer work and will be removed.

Sure. I will close this and open a new PR.

PR #969, opened 2021-06-17
Lazy load providers

### LINKED ISSUE #960: `qiskit.IBMQ.load_account()` is very slow
### Information

- **Qiskit IBMQ Provider version**: 0.17.4
- **Python version**: 3.8
- **Operating system**: macOS Big Sur

`qiskit.IBMQ.load_account()` is very slow, taking 7-12 seconds typically.
First few lines of profiling output below:
```
         105441 function calls (105130 primitive calls) in 10.533 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       94    9.968    0.106    9.968    0.106 {method 'read' of '_ssl._SSLSocket' objects}
        8    0.122    0.015    0.122    0.015 {method 'do_handshake' of '_ssl._SSLSocket' objects}
        8    0.110    0.014    0.110    0.014 {method 'load_verify_locations' of '_ssl._SSLContext' objects}
        8    0.088    0.011    0.088    0.011 {method 'connect' of '_socket.socket' objects}
        9    0.051    0.006    0.051    0.006 decoder.py:343(raw_decode)
        9    0.011    0.001    0.011    0.001 {built-in method _scproxy._get_proxy_settings}
        8    0.010    0.001    0.010    0.001 {built-in method _socket.getaddrinfo}
       23    0.009    0.000    0.034    0.001 backendconfiguration.py:856(_parse_channels)
        3    0.007    0.002    0.008    0.003 connectionpool.py:479(close)
```

Run `qiskit.IBMQ.load_account()` locally.

Although the function requires Network I/O, I can't see why it should take 10 seconds.

### Suggested solutions
No suggestions unfortunately.

Rcuz8:
### Summary

Adds lazy-loading to the `IBMQFactory` to speed up `load_account()`.

Note: Resolves #960 

### Details and comments

None

### DIFF
--- a/qiskit/providers/ibmq/ibmqfactory.py
+++ b/qiskit/providers/ibmq/ibmqfactory.py
@@ -103,7 +103,6 @@ class IBMQFactory:
                 'The URL specified ({}) is not an IBM Quantum Experience authentication '
                 'URL. Valid authentication URL: {}.'.format(credentials.url, QX_AUTH_URL))
 
-
         # Initialize the providers.
         self._initialize_provider(credentials)
 
@@ -507,6 +506,6 @@ class IBMQFactory:
                 except Exception:  # pylint: disable=broad-except
                     # Catch-all for errors instantiating the provider.
                     logger.warning('Unable to instantiate provider for %s: %s',
-                                hub, traceback.format_exc())
+                                   hub, traceback.format_exc())
 
         raise IBMQProviderError('Instantiable Provider with ID (%s) does not exist.' % target_id)
--- a/qiskit/providers/ibmq/ibmqfactory.py
+++ b/qiskit/providers/ibmq/ibmqfactory.py
@@ -17,7 +17,6 @@ from typing import Dict, List, Union, Callable, Optional, Any
 from collections import OrderedDict
 import traceback
 
-from .api.exceptions import RequestsApiError
 from .accountprovider import AccountProvider
 from .api.clients import AuthClient, VersionClient
 from .credentials import Credentials, discover_credentials
@@ -108,13 +107,6 @@ class IBMQFactory:
 
         # Setup credential info
         self._credentials = credentials
-        try:
-            self._auth_client = AuthClient(credentials.token,
-                                           credentials.base_url,
-                                           **credentials.connection_parameters())  # type: ignore
-            self._auth_client_hubs: List[Dict] = self._auth_client.user_hubs()  # type: ignore
-        except RequestsApiError:
-            logger.warning('Could not establish API connection with provided credentials.')
 
         # Initialize the providers.
         self._initialize_provider()
@@ -204,13 +196,6 @@ class IBMQFactory:
 
         # Setup credential info
         self._credentials = credentials
-        try:
-            self._auth_client = AuthClient(credentials.token,
-                                           credentials.base_url,
-                                           **credentials.connection_parameters())  # type: ignore
-            self._auth_client_hubs: List[Dict] = self._auth_client.user_hubs()  # type: ignore
-        except RequestsApiError:
-            logger.warning('Could not establish API connection with provided credentials.')
 
         # Initialize the providers.
         self._initialize_provider()
@@ -448,14 +433,6 @@ class IBMQFactory:
         if providers and len(providers) == 1:
             return providers[0]
 
-        # Load provider and retry
-        self._initialize_provider(hub, group, project)
-        providers = self.providers(hub, group, project)
-
-        # Test success
-        if providers and len(providers) == 0:
-            return providers[0]
-
         # Test fail: too many matches
         if providers and len(providers) > 1:
             raise IBMQProviderError('More than one provider matches the specified criteria.'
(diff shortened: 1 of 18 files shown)


qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -46,6 +46,7 @@ def __init__(self) -> None:
         """IBMQFactory constructor."""
         self._credentials = None  # type: Optional[Credentials]
         self._providers = OrderedDict()  # type: Dict[HubGroupProject, AccountProvider]
+        self.__lazy_loaded = True  # type: bool
```
jyu00, line 49: Why do you use 2 underscores here?

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -473,7 +480,11 @@ def _initialize_providers(self, credentials: Credentials) -> None:
             try:
                 provider = AccountProvider(provider_credentials, self)
                 self._providers[provider_credentials.unique_id()] = provider
+                if lazy:
+                    self.__lazy_loaded = True
+                    return
```
jyu00, line 485: So this would return after loading the first provider. But any subsequent call to `providers()`, including `get_provider()`, will still load all providers (and duplicate the `user_hubs()` call).

`load_account()` is slow mostly for people with multiple providers because it loads all providers (by this I mean getting the backend information for each provider). If I do
```
default_provider = IBMQ.load_account()
provider = IBMQ.get_provider('foo')
```
only providers `default_provider` and  `foo` need to be loaded.

jyu00 (changes requested):
Can you also add some test cases?

qiskit/providers/ibmq/ibmqfactory.py
Rcuz8: Fixed!

qiskit/providers/ibmq/ibmqfactory.py
Rcuz8: Fixed!

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -215,6 +216,21 @@ def load_account(self) -> Optional[AccountProvider]:
 
         return default_provider
 
+    def load_providers(self,
```
jyu00, line 219: This new lazy loading should be transparent to users - i.e. they shouldn't have to do a separate `_load_providers()` call.

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -215,6 +216,21 @@ def load_account(self) -> Optional[AccountProvider]:
...
+        """Loads a provider by its id (hub/group/project)
+
+        Args:
+            hub: Name of the hub to use.
+            group: Name of the group to use.
+            project: Name of the project to use.
+
+        """
+
+        self._initialize_provider(self._credentials, hub, group, project)
```
jyu00, line 232: So every time a user gets a provider this method calls `user_hubs()` again? That seems even less efficient than before where `user_hubs()` was only called once.

qiskit/providers/ibmq/ibmqfactory.py
```diff
@@ -444,36 +460,78 @@ def _check_api_version(credentials: Credentials) -> Dict[str, Union[bool, str]]:
...
                 services=service_urls.get('services', {}),
-                **hub_info, )
+                **uhub, )
 
             # Build the provider.
             try:
-                provider = AccountProvider(provider_credentials, self)
-                self._providers[provider_credentials.unique_id()] = provider
+                provider = AccountProvider(creds, self)
+                self._providers[creds.unique_id()] = provider
```
jyu00, line 528: So this would re-initialize the provider even if it has already been initialized?

jyu00 (changes requested):
As discussed before, the "lazy load" should call `user_hubs()` once to get all the high level provider information. Then upon the first reference of a provider, its `list_backends()` can be called to finish initialization.

qiskit/providers/ibmq/ibmqfactory.py
Rcuz8: Fixed via putting user hubs called in account loading/enabling functions so it only executes once

qiskit/providers/ibmq/ibmqfactory.py
Rcuz8: Fixed!

qiskit/providers/ibmq/ibmqfactory.py
Rcuz8: I think this is fixed. Added a check for the hub credentials already being initialized.

jyu00 (changes requested):
I don't think you need to change `ibmqfactory.py` at all. But if you really want to change it, the code needs to be cleaned up. For example, `_auth_client_hubs` is no longer used, `_initialize_provider()` actually initializes all providers, etc.

jyu00 (approved):
LGTM! I look forward to using this!

PR #970, opened 2021-06-17
remove job share level logic w/o breaking API

### LINKED ISSUE #914: update job share level after execution

rathishcholarajan:
### Summary
* job share level functionality is no longer working and hence removing the code here without breaking the public interface
* job share level will be removed from the public interface in a future release

### Details and comments
Fixes #914 (or rather addresses why we won't be implementing #914)

### DIFF
--- a/qiskit/providers/ibmq/ibmqbackend.py
+++ b/qiskit/providers/ibmq/ibmqbackend.py
@@ -36,7 +36,7 @@ from qiskit.providers.models import (QasmBackendConfiguration,
 from qiskit.util import deprecate_arguments
 
 from qiskit.providers.ibmq import accountprovider  # pylint: disable=unused-import
-from .apiconstants import ApiJobShareLevel, ApiJobStatus, API_JOB_FINAL_STATES
+from .apiconstants import ApiJobStatus, API_JOB_FINAL_STATES
 from .api.clients import AccountClient
 from .api.exceptions import ApiError
 from .backendjoblimit import BackendJobLimit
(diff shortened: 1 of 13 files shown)


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -267,6 +267,9 @@ def run(
         # pylint: disable=arguments-differ
         if job_share_level:
             try:
+                warnings.warn("The `job_share_level` keyword is deprecated due to low adoption "
+                              "and will be removed in a future release.",
+                              DeprecationWarning, stacklevel=3)
```
jyu00, line 272: Sorry if I wasn't clear in my previous comment. Normally we'd deprecate something for 3 months before removing it. But in this case job share level already doesn't work, so there's really no deprecating it (deprecation implies it still works). My comment was that we should keep the parameter / method in place, so user's existing code doesn't blow up (not that anyone is using lol). If the parameter/method is referenced, we can issue a warning saying the feature is no longer supported by IBM Quantum, and the parameter/method itself will also be removed in a future release. There is also no point passing it to api_client since it already doesn't work on the server side.

rathishcholarajan: @jyu00 got it, lol, I have updated now, please review.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -267,6 +267,9 @@ def run(
         # pylint: disable=arguments-differ
         if job_share_level:
             try:
```
jyu00, line 269: nit - we probably don't need to validate share level anymore since it'll be completely ignored.

qiskit/providers/ibmq/job/ibmqjob.py
```diff
@@ -627,6 +627,9 @@ def share_level(self) -> str:
         Returns:
             The share level of the job.
         """
+        warnings.warn("`IBMQJob.share_level()` method is no longer supported "
+                      "and will be removed in a future release.",
+                      Warning, stacklevel=2)
         if not self._share_level and not self._refreshed:
             self.refresh()
         return self._share_level
```
jyu00, line 635: There is no need to call `self.refresh()` (which is a network call) anymore since share level will always be `None`.

releasenotes/notes/unsupported-job-share-level-bfb3a505393f1981.yaml
```diff
@@ -0,0 +1,6 @@
+---
+other:
+  - |
+    Job share level in no longer supported due to low adoption and will be removed in a future release.
```
jyu00, line 4:
```suggestion
upgrade:
  - |
    Job share level in no longer supported due to low adoption and the 
    corresponding interface will be removed in a future release.
```

jyu00 (commented):
Can you also remove job share level from `api_client` and such? So we don't have more code that's essentially dead.

rathishcholarajan:
@jyu00 hopefully I've understood the changes requested, this time 😄,  please review. Thanks!

jyu00 (approved):
LGTM, thanks!

PR #975, opened 2021-06-21
Added runtime to iqx dashboard

### LINKED ISSUE #951: Add runtime jobs and programs to IQX dashboard
The "IQX dashboard" Jupyter widget current shows the progress of a circuit job. 

image

It should be updated to also show runtime jobs and programs (name and description). In addition, the widget should be renamed to something like "IBM Quantum dashboard" as the name IQX (IBM Quantum Experience) is no longer used.

[This tutorial](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/docs/tutorials/2_jupyter_tools.ipynb) shows how to use the Jupyter widgets, and the code is in the `jupyter` directory.

Rcuz8:
### Summary

1. Renames IQX Dashboard to IBM Quantum Dashboard
2. Adds Qiskit Runtime UI to the dashboard (2 Tabs: Programs & Jobs)
3. Updates dashboard styling

**Note:** Resolves #951 
**Note:** Continues: #956 

**Build Note**
⚠️ When I load the Devices tab, the picture for a random device will sometimes not load. My guess is that it could be a threading/updating issue because the images are the most dense data being loaded (render might occur before fully loaded). Any suggestions are welcome.


qiskit/providers/ibmq/jupyter/dashboard/dashboard.py
```diff
@@ -154,90 +231,127 @@ def stop_dashboard(self) -> None:
...
+                self.jobs = active_jobs
+            else:
+                self.rt_jobs = active_jobs
+        self.refresh_jobs_board(job_type)
 
     def _init_subscriber(self) -> None:
         """Initializes a subscriber that listens to job start events."""
 
-        def _add_job(job: IBMQJob) -> None:
+        def _add_job(job: Job) -> None:
```
jyu00, line 354: nit - `RuntimeJob` is not of type `Job`, so this technically is `Union[IBMQJob, RuntimeJob]`.

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
```diff
@@ -12,84 +12,160 @@
...
+            job: the job
+            watcher: the dashboard
+        """
+        super().__init__(job, watcher)
+        self._prev_queue_pos = None
+        self._prev_est_time = ''
+
+    def _set_job_queued(self) -> None:
+        """Perform queued updates"""
+        status = self.job.status()
```
jyu00, line 122: If you are re-querying job status here, it may no longer be `QUEUED`, in which case you shouldn't display queue position / est. start time.

jyu00 (commented):
I really like what you did with the `JobMonitor` class. Do you think you can do the same for `IQXDashboard`? I understand there is only 1 dashboard, but runtime and circuit jobs each has its own tab, so it'd be great if these tabs can also be children of a common parent.

qiskit/providers/ibmq/jupyter/dashboard/dashboard.py
Rcuz8: In the great `git force push` crash of 2021, `RuntimeJob` being made a subclass of `Job` was lost. This is updated now!

It had almost identical methods to `Job`, so making it a subclass actually reduced the size of the code (2 functions were redundant and were removed, 1 was added but given a `NotImplemented` error if used)

qiskit/providers/ibmq/jupyter/dashboard/watcher_monitor.py
Rcuz8: Fixed! This function now checks if the status is still `queued`.

PR #977, opened 2021-06-22
Use qpy to serialize quantum circuits

jyu00:
### Summary
This PR switches to using `qpy` to serialize `QuantumCircuit` for runtime instead of pickle, since pickle was only meant to be a stopgap until `qpy` was merged.

Note that this requires the next release of Qiskit Terra and hence is now on hold.

### DIFF
--- a/qiskit/providers/ibmq/runtime/utils.py
+++ b/qiskit/providers/ibmq/runtime/utils.py
@@ -16,11 +16,14 @@
 import json
 from typing import Any
 import base64
+import io
+import zlib
 
 import dill
 import numpy as np
 
 from qiskit.result import Result
+from qiskit.circuit import QuantumCircuit, qpy_serialization
 
 
 class RuntimeEncoder(json.JSONEncoder):
@@ -31,6 +34,14 @@ class RuntimeEncoder(json.JSONEncoder):
             return {'__type__': 'array', '__value__': obj.tolist()}
         if isinstance(obj, complex):
             return {'__type__': 'complex', '__value__': [obj.real, obj.imag]}
+        if isinstance(obj, QuantumCircuit):
+            # Serialize, compress, encode.
+            qpy_buffer = io.BytesIO()
+            qpy_serialization.dump(qpy_buffer, obj)
+            qpy_buffer.seek(0)
+            encoded = base64.standard_b64encode(zlib.compress(qpy_buffer.read()))
+            qpy_buffer.close()
+            return {'__type__': 'qpy', '__value__': encoded.decode('utf-8')}
         if isinstance(obj, Result):
             return {'__type__': 'result', '__value__': obj.to_dict()}
         if hasattr(obj, 'to_json'):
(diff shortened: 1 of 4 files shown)


qiskit/providers/ibmq/runtime/utils.py
```diff
@@ -31,6 +34,14 @@ def default(self, obj: Any) -> Any:  # pylint: disable=arguments-differ
             return {'__type__': 'array', '__value__': obj.tolist()}
         if isinstance(obj, complex):
             return {'__type__': 'complex', '__value__': [obj.real, obj.imag]}
+        if isinstance(obj, QuantumCircuit):
+            # Serialize, compress, encode.
+            qpy_buffer = io.BytesIO()
+            qpy_serialization.dump(qpy_buffer, obj)
```
mtreinish, line 40: This just swapped in https://github.com/Qiskit/qiskit-terra/commit/247f44ef87b08302514e512e4ed36601e95f33cd to be consistent with pickles order (it was tripping up the users testing qpy)

jyu00: Fixed in a4a84df

qiskit/providers/ibmq/runtime/utils.py
```diff
@@ -31,6 +34,14 @@ def default(self, obj: Any) -> Any:  # pylint: disable=arguments-differ
             return {'__type__': 'array', '__value__': obj.tolist()}
```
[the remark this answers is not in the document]
jyu00: This was copied from Qobj encoder which expects list. a4a84df updates it to use `np.save()`.

test/ibmq/runtime/test_runtime.py
```diff
@@ -101,6 +104,24 @@ def test_coder(self):
         self.assertIsInstance(decoded_result, Result)
         self.assertTrue((decoded_array == orig_array).all())
 
+    def test_encoder_qc(self):
```
rathishcholarajan, line 107: nit - Since both encoder and decoder are tested here, could we rename this test accordingly? Suggestion: `test_encoder_decoder_qc`

rathishcholarajan (changes requested):

test/ibmq/runtime/test_runtime.py
jyu00: I changed it to `test_coder_qc` since we already have a `test_coder` test case.

jyu00:
This PR now also does encoding and decoding for `OperatorBase` subclasses. It probably could have been a separate PR, but since it touches the same modules, combing them makes testing easier. It has dependency on Qiskit/qiskit-terra#6650.

Cryoris:
As soon as this is merged I'll base the Optimizer encoding/decoding on this 👍🏻

jyu00:
@Cryoris I was going to wait for Qiskit/qiskit-terra#6650 before merging this. But if you this sooner I can disable the tests and merge it sooner.

Cryoris:
Let's try to push the terra PR through I think that one looks good 👍🏻

PR #979, opened 2021-06-24
add pending parameter to filter jobs

### LINKED ISSUE #971: Add "pending" filter to runtime.jobs()
Runtime job querying now supports a new `pending` filter. `ibmq-provider` should support it when it's on production.

rathishcholarajan:
### Summary
add pending parameter in runtime.jobs():
* if `pending` is not specified all jobs are returned.
* if `pending` is set to True, 'QUEUED' and 'RUNNING' jobs are returned.
* if `pending` is set to False, 'DONE', 'ERROR' and 'CANCELLED' jobs are returned.

### Details and comments
Fixes #971

### DIFF
--- a/qiskit/providers/ibmq/runtime/ibm_runtime_service.py
+++ b/qiskit/providers/ibmq/runtime/ibm_runtime_service.py
@@ -418,13 +418,14 @@ class IBMRuntimeService:
         return self._decode_job(response)
 
     def jobs(self, limit: int = 10, skip: int = 0, pending: bool = None) -> List[RuntimeJob]:
-        """Retrieve all runtime jobs.
+        """Retrieve all runtime jobs, subject to optional filtering.
 
         Args:
             limit: Number of jobs to retrieve.
             skip: Starting index for the job retrieval.
-            pending: Returns 'QUEUED' and 'RUNNING' jobs if True,
-                returns 'DONE', 'CANCELLED' and 'ERROR' jobs if False.
+            pending: Filter by job pending state. If ``True``, 'QUEUED' and 'RUNNING'
+                jobs are included. If ``False``, 'DONE', 'CANCELLED' and 'ERROR' jobs
+                are included.
 
         Returns:
             A list of runtime jobs.
--- a/qiskit/providers/ibmq/runtime/ibm_runtime_service.py
+++ b/qiskit/providers/ibmq/runtime/ibm_runtime_service.py
@@ -417,12 +417,14 @@ class IBMRuntimeService:
             raise QiskitRuntimeError(f"Failed to delete job: {ex}") from None
         return self._decode_job(response)
 
-    def jobs(self, limit: int = 10, skip: int = 0) -> List[RuntimeJob]:
+    def jobs(self, limit: int = 10, skip: int = 0, pending: bool = None) -> List[RuntimeJob]:
         """Retrieve all runtime jobs.
 
         Args:
             limit: Number of jobs to retrieve.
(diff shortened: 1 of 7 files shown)


test/ibmq/runtime/fake_runtime_client.py
```diff
@@ -72,18 +72,23 @@ class BaseFakeRuntimeJob:
 
     _executor = ThreadPoolExecutor()
 
-    def __init__(self, job_id, program_id, hub, group, project, backend_name, params):
+    def __init__(self, job_id, program_id, hub, group, project, backend_name, final_status,
```
jyu00, line 75: I personally wouldn't add `final_status` here, as it's not a real job attribute. Even though this is a fake job, it'd be nice to keep it as realistic as possible. But since it is fake, I don't feel too strongly if you do't want to change it.

test/ibmq/runtime/test_runtime.py
```diff
@@ -239,6 +239,74 @@ def test_jobs_skip_limit(self):
         rjobs = self.runtime.jobs(skip=4, limit=2)
         self.assertEqual(2, len(rjobs))
 
+    def test_jobs_pending(self):
+        """Test retrieving pending jobs (QUEUED, RUNNING)."""
+        jobs = []
+        program_id = self._upload_program()
+        jobs = self._populate_jobs_with_all_statuses(jobs=jobs, program_id=program_id)
+        rjobs = self.runtime.jobs(pending=True)
+        self.assertEqual(5, len(rjobs))
```
jyu00, line 248: It'd be nice if the number of pending jobs can be calculated dynamically (e.g. based on the status of each job in `jobs`), instead of hard coding a magic number 5.

jyu00 (dismissed):
Can you also add a test to `test_runtime_integration`? So we have at least one that covers the real path.

test/ibmq/runtime/fake_runtime_client.py
rathishcholarajan: I'd like to keep it since it makes testing easier by setting the job's final status and avoid time.sleep which makes the tests flaky and slow.

test/ibmq/runtime/test_runtime.py
rathishcholarajan: Fixed!

rathishcholarajan:
> Can you also add a test to `test_runtime_integration`? So we have at least one that covers the real path.

Done!

qiskit/providers/ibmq/runtime/ibm_runtime_service.py
```diff
@@ -417,12 +417,14 @@ def job(self, job_id: str) -> RuntimeJob:
...
 
-    def jobs(self, limit: int = 10, skip: int = 0) -> List[RuntimeJob]:
+    def jobs(self, limit: int = 10, skip: int = 0, pending: bool = None) -> List[RuntimeJob]:
         """Retrieve all runtime jobs.
 
         Args:
             limit: Number of jobs to retrieve.
             skip: Starting index for the job retrieval.
+            pending: Returns 'QUEUED' and 'RUNNING' jobs if True,
+                returns 'DONE', 'CANCELLED' and 'ERROR' jobs if False.
```
jyu00, line 427:
```suggestion
            pending: Filter by job pending state. If ``True``, 'QUEUED' and 'RUNNING'
                jobs are included. If ``False``, 'DONE', 'CANCELLED' and 'ERROR' jobs
                are included.
```

jyu00 (approved):
LGTM!

PR #980, opened 2021-06-29
Add support for viewing error message if runtime job failed

### LINKED ISSUE #972: Interpret Runtime's 'Cancelled - Ran too long' state as Cancelled.
Interpret Runtime's 'Cancelled - Ran too long' state as Cancelled.

Docs will be updated soon to reflect this new state in the API. This is a workaround until the cancelled state has a reason value (implies state becomes a tuple of status, reason).

rathishcholarajan:
### Summary
Interpret runtime's "Cancelled - Ran too long" status as Error and also add new method RuntimeJob.error_message() which returns a string representing the reason for runtime job failure.

### Details and comments
Fixes #972

### DIFF
--- a/qiskit/providers/ibmq/runtime/runtime_job.py
+++ b/qiskit/providers/ibmq/runtime/runtime_job.py
@@ -24,7 +24,7 @@ from qiskit.providers.exceptions import JobTimeoutError
 from qiskit.providers.backend import Backend
 from qiskit.providers.jobstatus import JobStatus, JOB_FINAL_STATES
 
-from .constants import API_TO_JOB_STATUS
+from .constants import API_TO_JOB_ERROR_MESSAGE, API_TO_JOB_STATUS
 from .exceptions import RuntimeJobFailureError, RuntimeInvalidStateError, QiskitRuntimeError
 from .program.result_decoder import ResultDecoder
 from ..api.clients import RuntimeClient, RuntimeWebsocketClient, WebsocketClientCloseCode
@@ -107,6 +107,7 @@ class RuntimeJob:
         self._creation_date = creation_date
         self._program_id = program_id
         self._status = JobStatus.INITIALIZING
+        self._error_message = None  # type: Optional[str]
         self._result_decoder = result_decoder
 
         # Used for streaming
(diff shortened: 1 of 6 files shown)


qiskit/providers/ibmq/runtime/runtime_job.py
```diff
@@ -177,12 +178,22 @@ def status(self) -> JobStatus:
...
+            response_job_status = response['status'].upper()
             try:
-                self._status = API_TO_JOB_STATUS[response['status'].upper()]
+                self._status = API_TO_JOB_STATUS[response_job_status]
             except KeyError:
                 raise IBMQError(f"Unknown status: {response['status']}")
+            self._set_error_message(response_job_status=response_job_status)
         return self._status
 
+    def error_message(self) -> Optional[str]:
```
jyu00, line 189: I think this method can be slightly more helpful. For example, `_error_message` is only set in `status()`, so if a user doesn't call `status()` first (e.g. they saw the status on IQX), they will only get `None. In addition, `Unable to retrieve result for job` is not super helpful. The `results` endpoint returns stderr if the job failed (see line 150), so that can be used here as well.

test/ibmq/runtime/test_runtime_integration.py
```diff
@@ -53,11 +55,12 @@ def prepare_circuits(backend):
 
 def main(backend, user_messenger, **kwargs):
     iterations = kwargs['iterations']
-    interim_results = kwargs.pop('interim_results', {})
-    final_result = kwargs.pop("final_result", {})
+    interim_results = kwargs.pop('interim_results', {{}})
+    final_result = kwargs.pop("final_result", {{}})
```
jyu00, line 59: Instead of doing the double brackets format, why not just add another `sleep_per_itr` parameter to the program?

qiskit/providers/ibmq/runtime/runtime_job.py
rathishcholarajan: Fixed!

test/ibmq/runtime/test_runtime_integration.py
rathishcholarajan: Fixed!

qiskit/providers/ibmq/runtime/runtime_job.py
```diff
@@ -143,10 +144,9 @@ def result(
         _decoder = decoder or self._result_decoder
         if not self._results or (_decoder != self._result_decoder):  # type: ignore[unreachable]
             self.wait_for_final_state(timeout=timeout, wait=wait)
-            result_raw = self._api_client.job_results(job_id=self.job_id())
             if self._status == JobStatus.ERROR:
-                raise RuntimeJobFailureError(f"Unable to retrieve result for job {self.job_id()}. "
-                                             f"Job has failed:\n{result_raw}")
+                raise RuntimeJobFailureError(self.error_message())
```
jyu00, line 148:
```suggestion
                raise RuntimeJobFailureError(f"Unable to retrieve job result. {self.error_message()}")
```

jyu00 (approved):

jyu00 (commented):
Some minor doc changes but otherwise looks good!

PR #984, opened 2021-07-05
Test en-/decoding of Qiskit Optimizers

Cryoris:
### Summary

Since Qiskit's optimizers now implement a `settings` property and #977 added JSON encoding and decoding generally for such objects, optimizer serialization is already supported in the IBMQ provider. To ensure we keep supporting it, I'm adding some tests in this PR.


jyu00 (approved):
LGTM! Thanks!

PR #998, opened 2021-07-13
Fix test_upload_update_plot_data by using synchronous upload

### LINKED ISSUE #992: test_upload_update_plot_data failing since July 9
### Information

- **Qiskit IBMQ Provider version**: master (git cloned July 9)
- **Python version**: 3.8.1
- **Operating system**:  Ubuntu Bionic

This is a CI system that runs with the following variables set:

```
$ export USE_STAGING_CREDENTIALS=True
$ export QE_STAGING_URL=https://auth-dev.quantum-computing.ibm.com/api
$ export QE_STAGING_HGP=ibm-q-internal/labgroup/mriedem
```

This test started failing:

```
======================================================================

FAIL: test_upload_update_plot_data (test.ibmq.test_experiment.TestExperiment)

Test uploading and updating plot data.

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/build/IBM-Q-Software/sw-ibmq-results/qiskit-ibmq-provider-master/test/ibmq/test_experiment.py", line 694, in test_upload_update_plot_data

    self.assertEqual(rplot, friend_bytes, "Retrieved plot not equal updated plot.")

AssertionError: b'hello world' != b'hello friend!' : Retrieved plot not equal updated plot.
```

Presumably just run the qiskit-ibmq-provider experiment tests.

Tests should pass CI.

### Suggested solutions

I'm not sure what's causing the failure, there don't appear to be any related changes specific to the experiment service on July 9.

mriedem:
### Summary

Version 0.34.9 of the experiments service REST API changed the
default behavior on the `POST /experiments/:uuid/plots` and
`PUT /experiments/:uuid/plots/:name` APIs to upload the plot file
to backend (object) storage asynchronously.

That broke the `test_upload_update_plot_data` test which creates
a plot, updates it and then immediately gets the plot back and
compares the updated plot against the original plot. With the async
upload change on the server side the test intermittently fails due
to a race between the PUT and GET requests for the plot.

This change fixes that by making the plot update PUT request
synchronous by passing the `x-sync-upload: "True"` request header.
This way the client can be sure the plot was updated by the time
the PUT response is returned and before GET'ing the updated plot.

### Details and comments

The default behavior on the Qiskit client side is changed to always
upload plots synchronously both for the PUT and also the POST
request.

Closes #992


mriedem:
@jyu00 not sure if you want a release note for this?

jyu00:
> The default behavior on the Qiskit client side remains the same,
> i.e. asynchronous plot uploads to backend storage, since in most
> cases the client likely does not need to wait for the plot to be
> uploaded on the backend.

I think the default should be synchronous upload because if the upload takes very little time, then it's minimal impact to the user. If it takes more than very little, then the plot is in a limbo state - i.e. the user doesn't know if it's been uploaded.

mriedem:
> 
> 
> > The default behavior on the Qiskit client side remains the same,
> > i.e. asynchronous plot uploads to backend storage, since in most
> > cases the client likely does not need to wait for the plot to be
> > uploaded on the backend.
> 
> I think the default should be synchronous upload because if the upload takes very little time, then it's minimal impact to the user. If it takes more than very little, then the plot is in a limbo state - i.e. the user doesn't know if it's been uploaded.

It's sort of hard to get a very good sense on how the plot upload performs but in production the average plot creation (POST) routes are the slowest endpoints on average taking around 1 second but sometimes they can take 5+ seconds - those are outliers but I'm seeing it in our metrics dashboard. We're bound to whatever the network and performance is going to be on the IBM Cloud and it's object storage service for this.

I realize there isn't a great way for the client to tell if the plot is done uploading besides just trying to download it and retrying on 404 if it's not available in object storage yet, but that is a solution available today. Otherwise we'd have to build something into another API to check the status of plots for an experiment.

I can change this PR to default to `sync_upload=True` for both the plot creation (POST) and update (PUT) but I don't love it since we'll still have API requests blocking on a backend service operation.

mriedem:
> I can change this PR to default to `sync_upload=True` for both the plot creation (POST) and update (PUT) but I don't love it since we'll still have API requests blocking on a backend service operation.

A bit more complicated but probably the best long-term approach is that the service provides an API which can be used to get a temporary pre-signed `putObject` upload URL so Qiskit can upload the plots directly to object storage rather than going through resultsdb as a proxy.

jyu00:
> I realize there isn't a great way for the client to tell if the plot is done uploading besides just trying to download it and retrying on 404 if it's not available in object storage yet, 

Yeah this is the main reason I think we should go with sync upload as the default, especially since in async mode, there is a non-zero chance the API returns but the plot never makes it to object storage (e.g. due to networking error). In that case the user may discard their plot thinking its' safely stored in the cloud when it's not. 

> A bit more complicated but probably the best long-term approach is that the service provides an API which can be used to get a temporary pre-signed `putObject` upload URL so Qiskit can upload the plots directly to object storage rather than going through resultsdb as a proxy.

This is what we do with qObjects today.

jyu00:
> @jyu00 not sure if you want a release note for this?

Yes please, since it has an impact to the user interface (i.e. the new `sync_upload` keyword)

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -478,6 +479,9 @@ def upload_plot(
             plot: Name of the plot file or plot data to upload.
             plot_name: Name of the plot. If ``None``, the plot file name, if
                 given, or a generated name is used.
+            sync_upload: By default the server will upload the plot file
+                to backend storage asynchronously. Set this to False to use
+                that behavior and not block the upload.
```
jyu00, line 484:
```suggestion
            sync_upload: If ``True``, the plot will be uploaded synchronously. 
                Otherwise the upload will be asynchronous. 
```

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -490,28 +494,34 @@ def upload_plot(
...
     ) -> Dict:
         """Update an experiment plot.
 
         Args:
             experiment: The ``Experiment`` object or the experiment UUID.
             plot: Name of the plot file or plot data to upload.
             plot_name: Name of the plot to update.
+            sync_upload: By default the server will upload the plot file
+                to backend storage asynchronously. Set this to False to use
+                that behavior and not block the upload.
```
jyu00, line 515:
```suggestion
            sync_upload: If ``True``, the plot will be uploaded synchronously. 
                Otherwise the upload will be asynchronous. 
```

releasenotes/notes/resultsdb-plot-upload-sync-930e77b08da0974e.yaml
```diff
@@ -0,0 +1,10 @@
+---
+other:
```
jyu00, line 2:
```suggestion
features:
```

jyu00 (approved):
LGTM. Thanks for fixing this!

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -478,6 +479,8 @@ def upload_plot(
             plot: Name of the plot file or plot data to upload.
             plot_name: Name of the plot. If ``None``, the plot file name, if
                 given, or a generated name is used.
+            sync_upload: If ``True``, the plot will be uploaded synchronously. 
+                Otherwise the upload will be asynchronous.
```
jyu00, line 483:
```suggestion
            sync_upload: If ``True``, the plot will be uploaded synchronously.
                Otherwise the upload will be asynchronous.
```

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -490,28 +493,33 @@ def upload_plot(
...
+            sync_upload: bool = True
     ) -> Dict:
         """Update an experiment plot.
 
         Args:
             experiment: The ``Experiment`` object or the experiment UUID.
             plot: Name of the plot file or plot data to upload.
             plot_name: Name of the plot to update.
+            sync_upload: If ``True``, the plot will be uploaded synchronously. 
+                Otherwise the upload will be asynchronous.
```
jyu00, line 513:
```suggestion
            sync_upload: If ``True``, the plot will be uploaded synchronously.
                Otherwise the upload will be asynchronous.
```

PR #945, opened 2021-05-25
Update experiment service

jyu00:
### Summary
This PR updates the experiment service to use `ExperimentData`, `AnalysisResult`, and `ExperimentService` classes in `qiskit-terra`. This allows the new `qiskit-experiments` package to create experiments without requiring access to IBM Quantum.

### Details and comments
Closes #873, #854.

PR #1002, opened 2021-07-14
All Experiment Test Queries use `qiskit-test` tag

### LINKED ISSUE #999: test_experiment.py should avoid getting all resources
### Information

- **Qiskit IBMQ Provider version**: master (ebf2191)
- **Python version**: n/a
- **Operating system**: n/a

Some of the experiment service tests in https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/test/ibmq/test_experiment.py are using `limit=None` to page through _all_ resources (experiments and analysis_results it looks like). It looks like filtering is being applied but the filtering might not be scoped effectively to filter out many resources.

The problem with this is these tests run against a staging environment which we (on the service side) plan on using for scale/stress testing where we will be creating millions of experiments and analysis results. That might cause the tests here to timeout if they aren't scoping to a narrow set.

This is theoretical at this point since the staging environment has only a few dozen thousand records in it so paging through everything 1000 records at a time doesn't take too long, but once we have millions of records in the staging environment it could be a problem for these tests.

The tests should scope the resources they care about to only experiments that the tests themselves create. Granted, without using a clean empty database in the test runs that might be tricky so I'd expect some tests would need to be adjusted to only consider resources that they care about, i.e. that the tests created with a specific tag for example.

### Suggested solutions

Change the tests to always be including a filter that will be sure to only include resources created by the tests themselves. For example, the experiments and analysis results created by the tests all have a `qiskit-test` tag, so it might be sufficient to always filter by that tag when listing experiments and analysis results.

Rcuz8:
### Summary

As the experiments database populates with more non-test experiments, using `limit=None` freely in testing `experiment()` queries can become expensive. This push narrows those queries to just those with the `qiskit-test` tag, which is the tag given to all experiments creates in the `test_experiment.py` file.

### Fixes #999 

- All `experiments` queries In the experiment tests now use the `qiskit-test` tag filter, so they won't pull in unnecessary data.
- Analysis results now support the `tag` filter
- No longer depends on user to have test data. Generates if DNE.
- Fixes and enhances limit queries, which currently expect at least 10001 entries to pass.
----

**Note:** Will not prevent `analysis_results()` from loading non-test results, as there is no `tags` filter.
-  **Update**: `tags` filter being added to `analysis_results()`


Rcuz8:
@mriedem `ExperimentService` [doesn't support ](https://qiskit.org/documentation/stubs/qiskit.providers.ibmq.experiment.ExperimentService.html#qiskit.providers.ibmq.experiment.ExperimentService.analysis_results) filtering by tag for `analysis_results()` but [does](https://qiskit.org/documentation/stubs/qiskit.providers.ibmq.experiment.ExperimentService.html#qiskit.providers.ibmq.experiment.ExperimentService.experiments) for `experiments()`

mriedem:
> 
> 
> @mriedem `ExperimentService` [doesn't support ](https://qiskit.org/documentation/stubs/qiskit.providers.ibmq.experiment.ExperimentService.html#qiskit.providers.ibmq.experiment.ExperimentService.analysis_results) filtering by tag for `analysis_results()` but [does](https://qiskit.org/documentation/stubs/qiskit.providers.ibmq.experiment.ExperimentService.html#qiskit.providers.ibmq.experiment.ExperimentService.experiments) for `experiments()`

OK, but the API support is there so Qiskit _could_ support that. If it's needed for this change it could be done in a separate PR as new feature support and then this PR could build on it, but I'm not sure how necessary it is.

Rcuz8:
> OK, but the API support is there so Qiskit could support that. 

Adding the support `tags` as a filter property was pretty quick and relevant, it could definitely fit into this PR. Now just debugging the tests that depend on data already being populated. Will update here by tomorrow.

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -335,6 +337,14 @@ def analysis_results(
                 ``analysis_results(quality=[('ge', 'Bad'), ('lt', 'Good')])``
                 will return all analysis results with a quality of ``Bad`` and
                 ``No Information``.
+            tags: Filter by tags assigned to experiments.
```
mriedem, line 340:
```suggestion
            tags: Filter by tags assigned to analysis results.
```

Looks like copy/paste issue.

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -335,6 +337,14 @@ def analysis_results(
...
+            tags_operator: Logical operator to use when filtering by job tags. Valid
```
mriedem, line 341:
```suggestion
            tags_operator: Logical operator to use when filtering by tags. Valid
```

This is probably also a copy/paste problem from the experiment tags, which probably also shouldn't say "job" tags since these aren't jobs.

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -335,6 +337,14 @@ def analysis_results(
...
+                values are "AND" and "OR":
+
+                    * If "AND" is specified, then an experiment must have all of the tags
+                      specified in `tags` to be included.
+                    * If "OR" is specified, then an experiment only needs to have any
```
mriedem, line 346: More experiment wording here but this isn't an experiment resource.

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -317,6 +317,8 @@ def analysis_results(
             experiment_id: Optional[str] = None,
             result_type: Optional[str] = None,
             quality: Optional[List[Tuple[str, Union[str, ResultQuality]]]] = None,
+            tags: Optional[List[str]] = None,
+            tags_operator: Optional[str] = "OR",
```
mriedem, line 321: There should be a `features` release note for this since it's a new feature.

test/ibmq/test_experiment.py
```diff
@@ -43,9 +43,17 @@ def setUpClass(cls):
         super().setUpClass()
         try:
             cls.provider = cls._setup_provider()    # pylint: disable=no-value-for-parameter
-            cls.experiments = cls.provider.experiment.experiments()
+            cls.experiments = cls.provider.experiment.experiments(tags=['qiskit-test'], limit=None)
```
mriedem, line 46: If every call to `cls.provider.experiment.experiments` in this test module is going to include that tags filter, at least by default, I'd suggest adding a helper function that does that, e.g.:

```python
def list_experiments(self, tags=['qiskit-test'], **kwargs):
    return cls.provider.experiment.experiments(tags=tags, **kwargs)
```

Something like that just so you don't need to copy/paste the same thing all over the place for the tags filtering.

test/ibmq/test_experiment.py
```diff
@@ -43,9 +43,17 @@ def setUpClass(cls):
...
             cls.device_components = cls.provider.experiment.device_components()
-        except Exception:
+            cls.BACKEND_NAME = 'ibmq_qasm_simulator'
```
mriedem, line 48: Does this need to be set on the class instance? Couldn't it just be a constant at module scope?

test/ibmq/test_experiment.py
```diff
@@ -400,14 +409,14 @@ def test_delete_experiment(self):
 
     def test_reference_experiment_analysis_results(self):
         """Test referencing analysis results in an experiment."""
-        ref_result = self.provider.experiment.analysis_results()[0]
+        ref_result = self.provider.experiment.analysis_results(tags=['qiskit-test'])[0]
```
mriedem, line 412: Similar comment about having a help function to list analysis results that by default includes that tag.

test/ibmq/test_experiment.py
```diff
@@ -753,3 +773,30 @@ def _create_analysis_result(self):
...
+    def _create_initial_test_data(cls):
+        """Creates an initial push of test data."""
+        # Generate 3 pieces of data.
+        for _ in range(3):
+            # Create an experiment
+            exp = Experiment(  # pylint: disable=unexpected-keyword-arg
+                provider=cls.provider,
+                backend_name=cls.BACKEND_NAME,
+                experiment_type='test',
+                tags=['qiskit-test', 'qiskit-test-1'],  # 2 tags for multi-tag checks
```
mriedem, line 787:
```suggestion
        for x in range(3):
            # Create an experiment
            exp = Experiment(  # pylint: disable=unexpected-keyword-arg
                provider=cls.provider,
                backend_name=cls.BACKEND_NAME,
                experiment_type='test',
                tags=['qiskit-test', f'qiskit-test-{x}'],  # 2 tags for multi-tag checks
```

Nit but makes the experiments a bit less homogeneous.

test/ibmq/test_experiment.py
```diff
@@ -753,3 +773,30 @@ def _create_analysis_result(self):
...
+                start_datetime=datetime.now(),
+                share_level=ExperimentShareLevel.PUBLIC,
+                notes='test notes'
+            )
+            cls.provider.experiment.upload_experiment(exp)
```
mriedem, line 792: Should be creating the experiments per the helper function (`_create_experiment`) that is tracking the experiments and deleting them on tearDown otherwise these CI-generated experiments are hanging out in the database between runs:

image


That's not really ideal for a CI/test system.

mriedem (changes requested):

qiskit/providers/ibmq/experiment/experimentservice.py
```diff
@@ -353,14 +363,25 @@ def analysis_results(
                     qual = qual.value  # type: ignore[assignment]
                 qual_str = qual if op == 'eq' else "{}:{}".format(op, qual)
                 quality_list.append(qual_str)
+
+        tags_filter = None
+        if tags:
+            if tags_operator.upper() == 'OR':
+                tags_filter = 'any:' + ','.join(tags)
+            elif tags_operator.upper() == 'AND':
+                tags_filter = 'contains:' + ','.join(tags)
```
mriedem, line 372: I don't see anything testing this behavior. The default `tags_operator` value is `OR` and that's going to be exercised in the existing tests, but if you're adding a new feature here there should be test coverage for this case as well, and probably the `ValueError` case below.

test/ibmq/test_experiment.py
Rcuz8: The target of this function was to create persistent test records, in the event the user didn't have any to start with. This would make it so the initial calls to get `qiskit-test` records in `setupClass()` would be non-empty.

test/ibmq/test_experiment.py
Rcuz8: These are designed to be persistent test records so they would need to keep the `qiskit-test` tag exactly to show up in the test queries.

qiskit/providers/ibmq/experiment/experimentservice.py
Rcuz8: Updated!

qiskit/providers/ibmq/experiment/experimentservice.py
Rcuz8: Updated!

test/ibmq/test_experiment.py
Rcuz8: [Updated! ](https://github.com/Qiskit/qiskit-ibmq-provider/pull/1002/files#diff-9275794e4268a2c8dd165487dab5c87c3d271f9692458ac71902381b701725b6R787)

test/ibmq/test_experiment.py
Rcuz8: [Updated! ](https://github.com/Qiskit/qiskit-ibmq-provider/pull/1002/files#diff-6556bc5075d1aec9db8ea32abdc92d1064c71f0b43ac2a418486847932e9f0dcR65)Now a module constant.

test/ibmq/test_experiment.py
Rcuz8: Updated!

qiskit/providers/ibmq/experiment/experimentservice.py
Rcuz8: [Updated!](https://github.com/Qiskit/qiskit-ibmq-provider/pull/1002/files#diff-9275794e4268a2c8dd165487dab5c87c3d271f9692458ac71902381b701725b6R630-R671) Also both experiment and (new) analysis result tests check `ValueError` case.

test/ibmq/test_experiment.py
mriedem: > 
> 
> The target of this function was to create persistent test records, in the event the user didn't have any to start with. This would make it so the initial calls to get `qiskit-test` records in `setupClass()` would be non-empty.

We don't really _need_ persistent test records though right? Otherwise on the service side I'm going to have to implement a cron job that goes through and deletes old `qiskit-test` experiments. It shouldn't be that big a deal as part of the test class setup to assume there aren't any records and just create some test data like this is already doing.

Rcuz8: I agree that we don't need persistent test records. From what I understand, this test suite [was already dependent on them](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/test/ibmq/test_experiment.py#L46). It depends on which you prioritize:
Pros: Save creating jobs each time the test suite runs
Cons: Jobs remain in the user's job list

I personally don't think this is a harm to the user experience (i.e wouldn't require a cron job deleting):
1. the only users to see these experiments are those running these tests (developers)
2. these experiments are run only once and should fall down a user's experiments list very quickly

@mriedem If you still are thinking persistent test records should be removed, I'll remove them.

mriedem: > 
> 
> I agree that we don't need persistent test records. From what I understand, this test suite [was already dependent on them](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/test/ibmq/test_experiment.py#L46). It depends on which you prioritize:
> Pros: Save creating jobs each time the test suite runs
> Cons: Jobs remain in the user's job list
> 
> I personally don't think this is a harm to the user experience (i.e wouldn't require a cron job deleting):
> 
>     1. the only users to see these experiments are those running these tests (developers)
> 
>     2. these experiments are run only once and should fall down a user's experiments list very quickly
> 
> 
> @mriedem If you still are thinking persistent test records should be removed, I'll remove them.

Yes the tests were originally written assuming there were experiments in the database it had access to. I wouldn't say that's the best way to write tests though. CI should assume a clean environment to start and try to clean up after itself.

Also, these aren't IQX jobs, they are resultsdb experiments - different resources. The experiments resources created in these tests aren't actually running any jobs on the backend systems so they are quick to create and delete.

>the only users to see these experiments are those running these tests (developers)

This isn't true, the experiments created here are public which means anyone in the staging environment in IQX has access to them:

https://www-dev.quantum-computing.ibm.com/experiments?date_interval=last-90-days

image

That's not a huge deal but it could make someone wonder why there are these CI-generated public experiments hanging out.

So yeah my preference is that the tests should create what they need and clean up after themselves.

Rcuz8: Updated! Appreciate the reply and the patience.
`cls.analysis_results` and `cls.experiments` both pull from the test objects and `tearDownClass` will remove them!

mriedem:
Heavy merge conflict since #945. test/ibmq/test_experiment.py is gone now. I think the original issue #999 is probably not a problem anymore though since I was able to run the new test/ibmq/experiment/test_experiment_data_integration.py tests and those passed against our staging environment.

The changes to add support for analysis results tags is still valuable if that's not supported on the master branch yet, but those could either go into a new PR just for that feature or clean out the unrelated test changes in this PR and move test winkles for analysis results tags to test/ibmq/experiment/test_experiment_data_integration.py.

jyu00:
> The changes to add support for analysis results tags is still valuable if that's not supported on the master branch yet

#945 did add support for [filtering analysis results by tags](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/experiment/ibm_experiment_service.py#L759) ), along with support for other results db features that were held off waiting for `qiskit-experiments` interface (as described in #854).

mriedem:
> 
> 
> > The changes to add support for analysis results tags is still valuable if that's not supported on the master branch yet
> 
> #945 did add support for [filtering analysis results by tags](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/experiment/ibm_experiment_service.py#L759) ), along with support for other results db features that were held off waiting for `qiskit-experiments` interface (as described in #854).

OK I guess it seems this PR can be abandoned then and the related issue can be closed. I've verified that I can run the integration tests again with #1024.

PR #1012, opened 2021-07-19
Job Cleanup in Testing

### LINKED ISSUE #1010: Too Many Concurrent Jobs in Push Workflows
### Information

- **Qiskit IBMQ Provider version**: 0.15.0 (latest)
- **Python version**:
- **Operating system**: MacOS

IBMQTestJob (and potentially other test suites) currently leaves jobs running without cleaning up before moving onto other tests. This creates problems in [fork workflows](https://github.com/Rcuz8/qiskit-ibmq-provider/runs/3089349569?check_suite_focus=true#step:5:4542) with IBMQX `Max concurrent Jobs` errors when tests are running synchronously. 

This occurs on all `push` workflows on my fork. [Here](https://github.com/Rcuz8/qiskit-ibmq-provider/runs/3080613819?check_suite_focus=true#step:5:4047) is another example (for successful Travis builds). These workflows run concurrently, so they will be competing for active jobs slots (4 concurrent tests, max of 5 active jobs on any backend). 

The tests should not hit this limitation.

### Suggested solutions

Each test suite, and individual test if possible, should have no more than 1 job running concurrently. This can be done by adding `active_job.wait_for_final_result()` whenever that job's activity is no longer needed or the tests want to start another job. The problem seems particularly relevant in [TestIBMQJob](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/test/ibmq/test_ibmq_job.py).

Rcuz8:
### Summary

Currently, there are places where jobs are not being fully cleaned up after tests are done executing.

### Details and comments

Example of the GitHub Actions Push Test working [here](https://github.com/Rcuz8/qiskit-ibmq-provider/actions/runs/1091919374).

## Fixes #1010 

- Condenses/unifies test suite job creation to `TestSuite._sim_job(..)`
- Adds synchronization wherever possible (via  `job.wait_for_final_state()`)
- For the 2 tests that check concurrent jobs, an environment variable (`LIMIT_CONCURRENT_JOBS`) was introduced into the GitHub `main.yml` for Push tests only. This variable would be used only in CI builds that are testing multiple python versions concurrently.
- ⚠️ GitHub Actions (i.e `main.yml`) only began recognizing the staging environment (set via `QE_URL`) for me when I set it to the secrets variable, so that change is also reflected here.


test/ibmq/test_basic_server_paths.py
```diff
@@ -36,6 +42,17 @@ def setUpClass(cls, providers):
...
+        # Setup list of jobs to check have been deleted.
+        cls.jobs_to_delete: List[Tuple[AccountProvider, str]] = []
+
+    def tearDown(self) -> None:
+        # Ensure all jobs are deleted.
+        for provider, job_id in self.jobs_to_delete:
+            try:
+                provider.runtime.delete_job(job_id)
+            except RuntimeJobNotFound:
+                pass
```
rathishcholarajan, line 54: Could you add a log statement here like "Unable to delete runtime job: <job_id>" instead of pass?

test/ibmq/test_basic_server_paths.py
```diff
@@ -36,6 +42,17 @@ def setUpClass(cls, providers):
...
+        return super().tearDown()
```
rathishcholarajan, line 55: return is not necessary here. Just `super().tearDown()` should be good.

test/ibmq/test_basic_server_paths.py
```diff
@@ -47,6 +64,9 @@ def test_job_submission(self):
             with self.subTest(desc=desc, backend=backend):
                 job = self._submit_job_with_retry(ReferenceCircuits.bell(), backend)
 
+                # Ensure job is deleted
+                self.jobs_to_delete.append((provider, job.job_id()))
```
rathishcholarajan, line 68: Could you move this to the _submit_job_with_retry method so we don't have to add it to every test? You can pass the provider to that method.

test/ibmq/test_basic_server_paths.py
```diff
@@ -120,3 +149,14 @@ def _submit_job_with_retry(self, circs, backend, max_retry=5):
                 time.sleep(1)
 
         self.fail("Unable to submit job after {} retries: {}".format(max_retry, limit_error))
+
+    def _cancel_job(self, job: RuntimeJob) -> None:
```
rathishcholarajan, line 153: cancel_job is already imported from utils in this file. Could you use that and remove this?

rathishcholarajan (changes requested):
@Rcuz8 I have reviewed one file, could you fix same issues in other files as well and then I'll take a look again. Thanks!

test/ibmq/test_basic_server_paths.py
Rcuz8: [Updated! ](https://github.com/Qiskit/qiskit-ibmq-provider/pull/1012/files#diff-762f01fbe80fb731096879850a9910fe15c4286292d868375a0b9fdcbc576c4eR56) The language is slightly different, I can modify in next push.

test/ibmq/test_basic_server_paths.py
Rcuz8: Updated! For this specifically, the update is [here](https://github.com/Qiskit/qiskit-ibmq-provider/pull/1012/files#diff-762f01fbe80fb731096879850a9910fe15c4286292d868375a0b9fdcbc576c4eR140). A `_sim_job()` function was [also added](https://github.com/Qiskit/qiskit-ibmq-provider/pull/1012/files#diff-11a6c0aad7f62cdf57453b7d159a661b69a91cfb6d7f3e534775f800edc6a4f8R234) to accomplish the same thing for a few other files.

test/ibmq/test_basic_server_paths.py
Rcuz8: Updated! all definitions of `_cancel_job` were removed in favor of `cancel_job` from utils.

test/ibmq/test_basic_server_paths.py
Rcuz8: Updated!

test/ibmq/test_basic_server_paths.py
```diff
@@ -36,6 +43,17 @@ def setUpClass(cls, providers):
...
         cls.last_week = datetime.now() - timedelta(days=7)
+        # Setup list of jobs to check have been deleted.
+        cls.jobs_to_delete: List[Tuple[AccountProvider, str]] = []
+
+    def tearDown(self) -> None:
+        super().tearDown()
+        # Ensure all jobs are deleted.
+        for provider, job_id in self.jobs_to_delete:
+            try:
+                provider.runtime.delete_job(job_id)
```
jyu00, line 54: `provider.runtime.delete_job()` is only used to delete runtime jobs. These tests submit circuit jobs, not runtime jobs. `IBMQTestCase` is also supposed to automatically [record](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/test/ibmqtestcase.py#L81), [cancel](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/test/ibmqtestcase.py#L99), and [delete](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/test/ibmqtestcase.py#L101) jobs.

test/ibmq/test_ibmq_job.py
```diff
@@ -457,9 +470,10 @@ def test_retrieve_jobs_order(self):
             limit=10, status=JobStatus.DONE, descending=False, start_datetime=self.last_month)
         self.assertNotIn(job.job_id(), [rjob.job_id() for rjob in oldest_jobs])
 
-    @skip("Skip until aer issue 1214 is fixed")
     def test_retrieve_failed_job_simulator_partial(self):
-        """Test retrieving partial results from a simulator backend."""
+        """Test retrieving partial results from a simulator backend.
+        NOTE: Can run now! (https://github.com/Qiskit/qiskit-aer/pull/1218)
+        """
```
Rcuz8, line 476:
```suggestion
        """Test retrieving partial results from a simulator backend."""
```

test/ibmq/test_ibmq_job.py
```diff
@@ -346,6 +355,9 @@ def test_retrieve_jobs_running(self):
                             "Status for job {} is '{}' but should be RUNNING"
                             .format(rjob.job_id(), rjob._status))
 
+        # Cancel job so it doesn't consume more resources.
+        cancel_job(job)
+
```
Rcuz8, line 360:
```suggestion
```

test/ibmq/test_basic_server_paths.py
Rcuz8: Updated! Very cool, did not know `IBMQTestCase` handled all that!

PR #1014, opened 2021-07-26
Fix decoding of the `QNSPSA` optimizer

Cryoris:
### Summary

Decoding the QNSPSA optimizer fails due to the missing non-serializable first argument.
This commits fixes decoding by adding a placeholder for that argument which must be replaced by the user after decoding.

### Details and comments

Currently the tests fail since one of the settings is a dictionary with integer-keys which gets decoded into a string-keys dictionary. The optimizer will not recognize these string keys. Is this transformation on purpose or can we adapt it such that integers remain integers?

### DIFF
--- a/qiskit/providers/ibmq/runtime/utils.py
+++ b/qiskit/providers/ibmq/runtime/utils.py
@@ -177,6 +177,8 @@ class RuntimeDecoder(json.JSONDecoder):
             if obj_type == 'Instruction':
                 return _decode_and_deserialize(
                     obj_val, qpy_serialization._read_instruction, False)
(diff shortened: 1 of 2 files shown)


jyu00:
@Cryoris Since we have a patch on the runtime server side, I think the proper fix should be in `qiskit-terra`. That is, `QNSPSA`'s `settings` should just include `fidelity`. And if it's a Callable, the serialization will just fail. 

> Currently the tests fail since one of the settings is a dictionary with integer-keys which gets decoded into a string-keys dictionary.

Unfortunately that appears to be how Python JSON serializes dictionaries by default. I opened #1016 to address this issue.

PR #1018, opened 2021-07-27
Fix experiment-integration run command

mriedem:
### Summary

The experiment integration job is using the wrong make entry,
probably due to a copy/paste error.

### Details and comments

Follow up to #945


jyu00 (approved):
Thanks for making me notice I forgot to push my changes (including this fix) before merging the other PR...

Issue #1022, opened 2021-07-28
Use QE_DEVICE_STAGING in TestExperimentServerIntegration


mriedem:
### What is the expected enhancement?

The test is hard-coding to `ibmq_qasm_simulator` here:

https://github.com/Qiskit/qiskit-ibmq-provider/blob/0937e8a6c9b9150b1f809698d7931fd7c1492c9c/test/ibmq/experiment/test_experiment_server_integration.py#L43

But that might not be available to the provider being used to run those tests.

The experiments integration tests are being run setting `QE_STAGING_DEVICE=backend1` here:

https://github.com/Qiskit/qiskit-ibmq-provider/blob/0937e8a6c9b9150b1f809698d7931fd7c1492c9c/.github/workflows/cron-staging.yml#L48

So it seems we should be able to replace that hard-coding with reading from that environment variable but default to `ibmq_qasm_simulator` if not set.

mriedem:
Apparently have the same issue here for TestExperimentDataIntegration:

https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/test/ibmq/experiment/test_experiment_data_integration.py#L57

jyu00:
Closed by #1024

PR #1024, opened 2021-07-28
Use QE_STAGING_DEVICE in experiment testing

### LINKED ISSUE #1021: Add way to automatically install qiskit-requirements as dependency

Add `qiskit-experiments` to an `extras` install for something like `pip install qiskit-ibmq-provider[experiments]` or just to requirements-dev.txt to make it simpler to get that installed for running these tests:

https://github.com/Qiskit/qiskit-ibmq-provider/blob/5c11f797cd0fcbb91e1acc38e9bda6b97218dff1/test/ibmq/experiment/test_experiment_data_integration.py#L43

Right now I have to remember to manually pip install `qiskit-experiments`.

jyu00:
### Summary
Close #1021 and #1022.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -25,3 +25,5 @@ qiskit_rng
 qiskit-aer
 websockets>=8
 scikit-quant;platform_system != 'Windows'
+# TODO: Use released qiskit-experiments when it's released
+git+https://github.com/Qiskit/qiskit-experiments.git
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -26,4 +26,4 @@ qiskit-aer
 websockets>=8
 scikit-quant;platform_system != 'Windows'
 # TODO: Use released qiskit-experiments when it's released
-git+https://github.com/Qiskit/qiskit-experiments.git
+git+https://github.com/Qiskit/qiskit-experiments.git ; python_version > '3.6'
--- a/test/ibmq/experiment/test_experiment_data_integration.py
+++ b/test/ibmq/experiment/test_experiment_data_integration.py
@@ -53,8 +53,7 @@ class TestExperimentDataIntegration(IBMQTestCase):
         if not cls.provider.has_service('experiment'):
             raise SkipTest("Not authorized to use experiment service.")
 
-        # cls.backend = cls._setup_backend()  # pylint: disable=no-value-for-parameter
-        cls.backend = cls.provider.get_backend('ibmq_qasm_simulator')
+        cls.backend = cls._setup_backend()  # pylint: disable=no-value-for-parameter
         cls.device_components = cls.provider.experiment.device_components(cls.backend.name())
         if not cls.device_components:
             raise SkipTest("No device components found.")
@@ -138,7 +137,7 @@ class TestExperimentDataIntegration(IBMQTestCase):
         exp_data.save()
         self.experiments_to_delete.append(exp_data.experiment_id)
 
-        credentials = self.provider.credentials
+        credentials = self.backend.provider().credentials
         rexp = DbExperimentData.load(exp_data.experiment_id, self.experiment)
         self._verify_experiment_data(exp_data, rexp)
         self.assertEqual(credentials.hub, rexp.hub)  # pylint: disable=no-member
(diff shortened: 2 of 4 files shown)


rathishcholarajan (approved):

mriedem (approved):
LGTM (I had to lookup https://github.com/Qiskit/qiskit-ibmq-provider/blob/06e557b6245b1baf6865aee2a7dea5db6f301317/test/decorators.py#L171).

mriedem:
I got a passing run with this:

image

test/ibmq/experiment/test_experiment_data_integration.py
```diff
@@ -53,8 +53,7 @@ def setUpClass(cls):
         if not cls.provider.has_service('experiment'):
             raise SkipTest("Not authorized to use experiment service.")
 
-        # cls.backend = cls._setup_backend()  # pylint: disable=no-value-for-parameter
-        cls.backend = cls.provider.get_backend('ibmq_qasm_simulator')
```
mriedem, line 57: What about also fixing this class?

https://github.com/Qiskit/qiskit-ibmq-provider/blob/58d1ca3a9c2ffcfc6625c4483f7b113573b3d4c0/test/ibmq/experiment/test_experiment_server_integration.py#L43

mriedem (commented):
Guess I forgot about https://github.com/Qiskit/qiskit-ibmq-provider/blob/58d1ca3a9c2ffcfc6625c4483f7b113573b3d4c0/test/ibmq/experiment/test_experiment_server_integration.py#L43 - can that be updated too?

mriedem (approved):
Verified the other test module is passing as well (and I can use `backend1` with the env var override):

image

requirements-dev.txt
```diff
@@ -25,3 +25,5 @@ qiskit_rng
 qiskit-aer
 websockets>=8
 scikit-quant;platform_system != 'Windows'
+# TODO: Use released qiskit-experiments when it's released
+git+https://github.com/Qiskit/qiskit-experiments.git ; python_version > '3.6'
```
mriedem, line 29: Note that I think when there is a release for this on pypi there should be an 'extras' install available for it so users can just do `pip install qiskit-ibmq-provider[experiments]`, similar to how we can do `pip install qiskit[all]`.

jyu00: Why can't they just do `pip install qiskit-experiments` directly? Making it `pip install qiskit-ibmq-provider[experiments]` to me implies it's needed to use the "experiment service", which is not true.

mriedem: > 
> 
> Why can't they just do `pip install qiskit-experiments` directly? Making it `pip install qiskit-ibmq-provider[experiments]` to me implies it's needed to use the "experiment service", which is not true.

OK I guess I'm just confused about how these two packages are working together. Without qiskit-experiments I can continue to just use qiskit-ibmq-provider and hit the experiment service directly, as before, right? The qiskit-experiments package doesn't require qiskit-ibmq-provider either, I guess it _can_ if you want to use the IBM Quantum experiment service as a remote backend, but you don't have to - you can use a "local DB" or some other provider that implements the same API theoretically, right? But qiskit-experments doesn't make any REST API requests to any remote services, is that right - qiskit-ibmq-provider does.

It's confusing because if qiskit-ibmq-provider is essentially a configurable backend for qiskit-experiments, it's weird to me that the integration tests are in qiskit-ibmq-provider rather than qiskit-experiments, but maybe that was just easier since it was already working.

mriedem: > It's confusing because if qiskit-ibmq-provider is essentially a configurable backend for qiskit-experiments, it's weird to me that the integration tests are in qiskit-ibmq-provider rather than qiskit-experiments

I guess I should say that it's confusing to me that the qiskit-ibmq-provider experiment integration tests require qiskit-experiments.

jyu00: > Without qiskit-experiments I can continue to just use qiskit-ibmq-provider and hit the experiment service directly, as before, right? 

Yes

>The qiskit-experiments package doesn't require qiskit-ibmq-provider either, I guess it can if you want to use the IBM Quantum experiment service as a remote backend, but you don't have to - you can use a "local DB" or some other provider that implements the same API theoretically, right? 

That was the idea, but I think they are going to require `qiskit-ibmq-provider` now. However, one can still use `qiskit-experiments` to run experiments on, for example, local Aer simulator or third party backends without using results DB.

> I guess I should say that it's confusing to me that the qiskit-ibmq-provider experiment integration tests require qiskit-experiments.

This is mostly because ibmq-provider tests were already setup to access IBMQ, whereas the other Qiskit repos only use local simulator. At some point we will setup integration tests in qiskit-experiments, then maybe these can be moved.

PR #1033, opened 2021-07-30
Fix experiment dependency

jyu00:
### Summary
`qiskit-experiments` now has a dependency on `qiskit-ibmq-provider` 0.16, which hasn't been released yet. This caused CI to fail if `qiskit-experiments` is installed before `ibmq` main branch. ~This PR moves `qiskit-experiments` main out of `requirements-dev.txt` and into `cron-staging.yml`. Once `ibmq` is released it can then be moved back to `requirements-dev.txt`.~

This PR moved the installation of `requirements-dev.txt` to after `ibmq` so we get 0.16.

### DIFF
--- a/.github/workflows/cron-staging.yml
+++ b/.github/workflows/cron-staging.yml
@@ -35,8 +35,8 @@ jobs:
       - name: Install Deps
         run: |
           python -m pip install --upgrade pip
-          pip install -U -c constraints.txt -r requirements-dev.txt
           pip install -c constraints.txt -e .
+          pip install -U -c constraints.txt -r requirements-dev.txt
       - name: Run Tests
         run: make runtime_integration
   experiment-integration:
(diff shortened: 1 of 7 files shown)


.github/workflows/cron-staging.yml
```diff
@@ -57,9 +57,11 @@ jobs:
         with:
           python-version: 3.8
       - name: Install Deps
+        # TODO qiskit-experiments can be moved to requirements-dev.txt once ibmq 0.16 is released.
         run: |
           python -m pip install --upgrade pip
           pip install -U -c constraints.txt -r requirements-dev.txt
```
mtreinish, line 63: You could move this below the next line (so you install ibmq before requirements-dev.txt) and that would resolve it to without needing to remove experiments from requirements-dev.txt

jyu00: Yeah I was being lazy and didn't want to change all 6 `.yml` files lol

jyu00: ok I made the changes so @mriedem won't have to manually install `qiskit-experiments` if we don't release on Monday :see_no_evil:

PR #1038, opened 2021-08-04
encoder/decoder handle datetime

### LINKED ISSUE #1015: Support datetime in Runtime en/decoder
`RuntimeEncoder` and `RuntimeDecoder` should support Python `datetime`, which is not JSON serializable by default.

kt474:
### Summary

fixes #1015 

Still need to check all datetime formats & test cases


rathishcholarajan:
Fixed in https://github.com/Qiskit-Partners/qiskit-ibm/pull/2

PR #1061, opened 2021-10-08
Support runtime session

jyu00:


qiskit/providers/ibmq/runtime/runtime_session.py
```diff
@@ -0,0 +1,102 @@
...
+        inputs.update(kwargs)
+        self._job = self._service.run(
+            program_id=self._program_id,
+            options=self._options,
+            inputs=inputs,
+            image=self._image
+        )
+
+    @_active_session
+    def read(self) -> Any:
```
ajavadia, line 74: can you allow the ability to reduce `wait` in between polls?

PR #1065, opened 2021-10-23
Fix issue with parameters in metadata

rafal-pracht:
### Summary
When we have a Parameter in circuit metadata then during the run of the job it crashes. 
For example, the circuit below before the fix throw the error: "TypeError: keys must be str, int, float, bool or None, not Parameter"

qr = QuantumRegister(1)
cr = ClassicalRegister(1)
my_circ_str = 'test_metadata'
my_circ = QuantumCircuit(qr, cr, name=my_circ_str, metadata={Parameter('φ'): 0.2})
job = backend.run(my_circ, shots=1024) 

This issue was reported by @nbronn.

### Details and comments
The solution: in _submit_job method after converting the Qobj to the dictionary we check if in the dictionary are only the valid JSON key, if not the key is converted to a string.


rafal-pracht:
@jyu00 Thanks for your review and sorry for the delay. I've updated the IQXJsonEncoder, and now it encodes the Parameter properly.

Regarding losing the Parameter during encoding. I've discussed this issue with @nbronn and this will not be an issue (we use the Parameter in metadata only for description purposes).

But it will not be an issue for me to write a code that doesn't lose the Parameter during encoding if you want. If yes, could you give me an example or point to the python file where we save a custom object as JSON in qiskit

rafal-pracht:
Hi @jyu00 
I've spent some time fixing the unit test issue. And because I've no idea what happened, I've rollback all my changes. So now there are no changes in my branch, but the unit tests still fail with "Exception: Unable to locate valid credentials.". Do you have any hints for me?

jyu00:
@rafal-pracht The tests require some setup to run because they need IBM Quantum credentials. You'll need to define these environment variables in Github actions:

```
        * QE_TOKEN: default token to use.
        * QE_URL: default url to use (set this to https://auth.quantum-computing.ibm.com/api)
        * QE_HGP: default hub/group/project to use.
        * QE_DEVICE: default device to use.
```

There are also some known failures that we are still working on, so don't spend too much time if you see failed tests. 

@rathishcholarajan we should add the env variables needed for testing to the contribution guide.

rafal-pracht:
Thanks @jyu00, I've pushed my changes once again. It would be great if you can look and give me feedback. 

In meanwhile I'll look at the GitHub environment setup.

jyu00:
@rafal-pracht can you add some test cases?

rafal-pracht:
@jyu00 I've added some test cases.

test/ibmq/utils/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Test for the qiskit-ibmq-provider package."""
```
jyu00, line 13: We don't usually create separate packages for tests, unless the functionality requires multiple test files (e.g. `runtime`). In addition, `test_serialization.py` already has some test cases for `IQXJsonEncoder`, so you can just reuse it.

test/ibmq/utils/test_json_encoder.py
```diff
@@ -0,0 +1,61 @@
...
+
+    @requires_provider
+    def test_exception_message(self, provider):
+        """Test executing job with Parameter in methadata."""
+        qr = QuantumRegister(1)
+        cr = ClassicalRegister(1)
+        my_circ_str = 'test_metadata'
+        my_circ = QuantumCircuit(qr, cr, name=my_circ_str, metadata={Parameter('φ'): 0.2})
+        backend = provider.get_backend('ibmq_bogota')
+        backend.run(my_circ, shots=1024)
```
jyu00, line 34: Do we really need to submit a job here instead of just doing a `json.dumps()`? We usually try not to take up machine time, especially real devices.

jyu00 (changes requested):

rafal-pracht:
@jyu00  Thanks for your feedback, I've just done the changes.

jyu00 (approved):
LGTM, thanks!

PR #1094, opened 2021-12-07
Remove extra decompose in RuntimeEncoder

jyu00:
### Summary

`RuntimeEncoder` does an extra `decompose()` if the circuit is a `BlueprintCircuit`, to address a QPY issue. The issue has since been merged and deployed, and this extra decompose is causing QAOA runtime program to fail.


rathishcholarajan:
@jyu00 do we need a release note and a patch release?

jyu00:
@rathishcholarajan Yup just added a release note. And yes we'll need a patch release. Terra is planning a patch release for Thursday so we can do it at the same time.

rathishcholarajan (approved):
Looks good, thanks @jyu00!

PR #1103, opened 2022-01-07
Test no longer supported proxy url format only when requests <= 2.26

rathishcholarajan:
### Summary
requests 2.27 dropped support for proxy url format `'http:{}:{}'`. Hence running the test separately for this format only when requests is less than 2.27.

### Details and comments
Fixes https://github.com/Qiskit-Partners/qiskit-ibm/issues/226

### DIFF
--- a/test/ibmq/test_proxies.py
+++ b/test/ibmq/test_proxies.py
@@ -15,6 +15,8 @@
 import urllib
 import subprocess
 
+from unittest import skipIf
+from requests import __version__ as requests_version
 from requests.exceptions import ProxyError
 
 from qiskit.providers.ibmq import IBMQFactory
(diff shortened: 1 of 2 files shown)


test/ibmq/test_proxies.py
```diff
@@ -150,6 +151,21 @@ def test_proxy_urls(self, qe_token, qe_url):
...
+        test_urls = [
+            'http:{}:{}'.format(ADDRESS, PORT)
+        ]
+        for proxy_url in test_urls:
+            with self.subTest(proxy_url=proxy_url):
+                credentials = Credentials(
+                    qe_token, qe_url, proxies={'urls': {'https': proxy_url}})
+                version_finder = VersionClient(credentials.base_url,
+                                               **credentials.connection_parameters())
+                version_finder.version()
```
daka1510, line 167: I don't see much value in porting this PR as-is because this will add a test case that will always be skipped (`requests` currently resolves to `2.27.1` on a fresh install).

@kt474 - Can you check if there is another way to make the test compatible with most recent request versions?

@rathishcholarajan - please jump in if you think this is not a reasonable proposal. I would have probably given similar feedback for https://github.com/Qiskit/qiskit-ibmq-provider/pull/1103.

PR #1117, opened 2022-02-02
Allow users to set a logging level for runtime jobs (#99)

jyu00:
### Summary
Backport from https://github.com/Qiskit/qiskit-ibm-runtime/pull/99

Allows users to specify a custom logging level for runtime jobs

### DIFF
--- a/qiskit/providers/ibmq/runtime/runtime_options.py
+++ b/qiskit/providers/ibmq/runtime/runtime_options.py
@@ -1,0 +1,64 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2022.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Runtime options that control the execution environment."""
+
+import re
+import logging
+from dataclasses import dataclass
+from typing import Optional
+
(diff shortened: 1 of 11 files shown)


qiskit/providers/ibmq/runtime/runtime_options.py
```diff
@@ -0,0 +1,64 @@
...
+
+import re
+import logging
+from dataclasses import dataclass
+from typing import Optional
+
+from ..exceptions import IBMQInputValueError
+
+
+@dataclass
```
rathishcholarajan, line 23: I see dataclass is a Python 3.7+ feature. We still support Python 3.6 for the upcoming provider release since terra 0.19.x still supports it and then it will be dropped after the next release. Would this work with Python 3.6?

rathishcholarajan (changes requested):
Left a comment... Could you please add a release note as well?

qiskit/providers/ibmq/runtime/runtime_options.py
jyu00: But Python 3.6 end of life was more than a month ago, and the latest qiskit-terra no longer support 3.6.

mtreinish: Just add `dataclasses>=0.8;python_version<'3.7'` to the requirements list. Being in the metapackage puts extra requirements and for the metapackage we've got more conservative support versions and we've previously advertised that terra 0.19.x (and aer 0.10.x) supports python 3.6. So we can't add anything to the metapackage that requires >3.6 until we release terra 0.20.0

rathishcholarajan (changes requested):
@jyu00 Could you please add dataclasses to setup.py as well?

rathishcholarajan (approved):
LGTM! Thanks @jyu00!

PR #1140, opened 2022-08-08
Add deprecation warning

### LINKED ISSUE #1042: Deprecate qiskit-ibmq-provider
Add general deprecation warning for qiskit-ibmq-provider package and encourage users to move to qiskit-ibm-runtime, qiskit-ibm-provider an qiskit-ibm-experiment when they are ready. We will be doing only critical security fixes for qiskit-ibmq-provider for 3 more months from deprecation after which it will no longer be supported.

kt474:
### Summary

fixes #1042


kt474:
@rathishcholarajan can you take a look

kt474:
@renier

mtreinish:
So I'm a bit confused here, what is the status of qiskit-ibm-provider if we're telling people to use the runtime client instead? They're not equivalent sets of functionality. I get you're adding an interface for doing `backend.run()` via the circuit runner, but is that all that's needed? What about the other IQX api functionality? Also what about the in progress PRs like: https://github.com/Qiskit/qiskit-ibm-provider/pull/365 ?

kt474:
> So I'm a bit confused here, what is the status of qiskit-ibm-provider if we're telling people to use the runtime client instead? They're not equivalent sets of functionality. I get you're adding an interface for doing `backend.run()` via the circuit runner, but is that all that's needed? What about the other IQX api functionality? Also what about the in progress PRs like: [Qiskit/qiskit-ibm-provider#365](https://github.com/Qiskit/qiskit-ibm-provider/pull/365) ?

@renier - also unsure about the exact plan for this repo `qiskit-ibm-provider`

1ucian0:
> I get you're adding an interface for doing `backend.run()` via the circuit runner, but is that all that's needed?

For reference https://github.com/Qiskit/qiskit-ibm-runtime/issues/459

jyu00:
Putting on hold label back on. Like Matthew said we shouldn't deprecate this until there is a replacement.

qiskit/providers/ibmq/__init__.py
```diff
@@ -112,6 +113,15 @@
 QISKIT_IBMQ_PROVIDER_LOG_FILE = 'QISKIT_IBMQ_PROVIDER_LOG_FILE'
 """The environment variable name that is used to set the file for the IBM Quantum logger."""
 
+warnings.warn(
+    f"The package {__name__} is being deprecated. There will only be critical security fixes "
+    "for the next 3 months and then this package will no longer be supported. Please begin to use "
+    "qiskit-ibm-provider (https://github.com/Qiskit/qiskit-ibm-provider) and "
+    "qiskit-ibm-runtime (https://github.com/Qiskit/qiskit-ibm-runtime) instead.",
```
1ucian0, line 120:
```suggestion
    f"The package {__name__} is being deprecated."
    "Please see "
    "https://qiskit.org/documentation/partners/qiskit_ibm_provider/tutorials/Migration_Guide_from_qiskit-ibmq-provider.html
    "to get instructions on how to migrate to "
    "qiskit-ibm-provider (https://github.com/Qiskit/qiskit-ibm-provider) and "
    "qiskit-ibm-runtime (https://github.com/Qiskit/qiskit-ibm-runtime) instead.",
```

qiskit/providers/ibmq/__init__.py
```diff
@@ -112,6 +113,15 @@
...
+    DeprecationWarning,
```
1ucian0, line 121: `FutureWarning`?

qiskit/providers/ibmq/__init__.py
```diff
@@ -97,6 +98,18 @@
 from .runner_result import RunnerResult
 from .version import __version__
 
+# pylint: disable=line-too-long
+warnings.warn(
+    f"The package {__name__} is being deprecated. "
+    "Please see "
+    "https://qiskit.org/documentation/partners/qiskit_ibm_provider/tutorials/Migration_Guide_from_qiskit-ibmq-provider.html"
```
1ucian0, line 105: I just created ibm.biz/provider_migration_guide, which is shorter and can be updated if the url changes. Maybe a better option?

```suggestion
    "https://ibm.biz/provider_migration_guide"
```

HuangJunye: Just FYI the original link was added to [the release note](https://github.com/Qiskit/qiskit-terra/pull/8080/files#diff-a7772d1c2d1e096b092f263558880069ac8c4cda7539c3e4851e1f9aef54fbe3R12) of https://github.com/Qiskit/qiskit-terra/pull/8080. If the url changes, we can consider change there as well.

HuangJunye: But probably the best way to handle the link change would be redirect the original link to the new link. This can be done in sphinx using reredirect extension. Anyway, we don't need to worry about that now.

PR #1153, opened 2023-01-31
Add an error for dynamic circuit usage

taalexander:
### Summary
This PR adds an error if a user tries to use `dynamic=True` with the old provider

### DIFF
--- a/qiskit/providers/ibmq/ibmqbackend.py
+++ b/qiskit/providers/ibmq/ibmqbackend.py
@@ -289,6 +289,14 @@ class IBMQBackend(Backend):
                 "'use_measure_esp' is unset or set to 'False'."
             )
 
+        if run_config.get("dynamic", None):
+            raise IBMQBackendValueError(
+                "The IBMQ provider is being deprecated and does not support dynamic circuits. "
+                "Please use the qiskit-ibm-provider instead. "
+                "See the documentation for more information on usage - "
+                "https://qiskit.org/documentation/partners/qiskit_ibm_provider/#qiskit-ibm-quantum-provider-documentation. "
+            )
+
         if not self.configuration().simulator:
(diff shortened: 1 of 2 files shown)


qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -289,6 +289,14 @@ def run(
                 "'use_measure_esp' is unset or set to 'False'."
             )
 
+        if run_config.get("dynamic", None):
```
jakelishman, line 292: Is your intent to deliberately permit setting `dynamic=False` without an error?  If so, then fine, if not, maybe you wanted `if run_config.get("dynamic", None) is not None:`.

taalexander: Initially this is my intent since this will pass-through unset, but on second thought I'll just prevent any setting of this value.

taalexander: Updated in 57dcc2d.

qiskit/providers/ibmq/ibmqbackend.py
```diff
@@ -289,6 +289,14 @@ def run(
...
+            raise IBMQBackendValueError(
+                "The IBMQ provider is being deprecated and does not support dynamic circuits. "
+                "Please use the qiskit-ibm-provider instead. "
+                "See the documentation for more information on usage - "
+                "https://qiskit.org/documentation/partners/qiskit_ibm_provider/#qiskit-ibm-quantum-provider-documentation. " # noqa: E501
```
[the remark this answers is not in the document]
taalexander: Fixed in 1d8e962.

jakelishman (approved):
Here's my tick, though I don't think it means anything on this repo.

kt474 (approved):

PR #1159, opened 2023-02-21
fix python requirement versions

kt474:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,10 @@
-qiskit-terra>=0.18.0
-requests>=2.19
-requests_ntlm>=1.1.0
+qiskit-terra~=0.23.0
+requests~=2.19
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,10 +1,10 @@
-qiskit-terra>=0.18.0
-requests>=2.19
-requests_ntlm>=1.1.0
+qiskit-terra~=0.23.0
```
mtreinish, line 1: This isn't valid (nor are most of the changes in this file). What this does in practice is prevents people from installing the qiskit-ibmq-provider and terra >0.23.0 together which will actively cause issues in the qiskit metapackage when terra 0.24.0 is released and will block the metapackage from being releasable. 

Capping versions like this makes solving dependencies much harder  and it should only be done if there are specific incompatibilities which are known.

